PR #2, opened 2015-12-03
Remove cruft

owennw:
This strips out references to bower, and any linting/beautifying done on js. I don't see anything else to be removed, let me know if you do though.


bjedrzejewski:
Good to go

PR #18, opened 2015-12-04
Load showcase through npm and build into openfin project

owennw:
Currently targeting version 8 of the showcase. @bjedrzejewski can you pull this change and run it locally before merging? For some reason the showcase page just gives me a 'loading' screen. Looking at the console it looks like a weird error, but I'm not sure if that's just on my machine.


bjedrzejewski:
@owennw grunt-contrib-copy seems to be missing as a dependency

owennw:
Replace by #19.

PR #19, opened 2015-12-07
Load showcase through npm and build into openfin project

owennw:
Git error, new branch.

### DIFF
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -15,6 +15,9 @@ module.exports = function(grunt) {
             ],
             css: [
                 'public/**/*.css'
+            ],
+            showcase: [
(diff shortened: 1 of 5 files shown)


bjedrzejewski:
@owennw is this not what we are building:

public/assets/css/style.css
public/assets/js/app.min.js
public/d3fc-showcase.html

Shall we gitignore that?

owennw:
I'm not sure if we should ignore them: they're a core dependency of our project. They will only change when we change the version of the showcase we use, or if we make changes to the files ourselves. Besides that they won't change so I'm happy having a one time commit in here.

bjedrzejewski:
@owennw When the less/css is done, then it may just override the file. Lets keep it there for now.

PR #26, opened 2015-12-08
Added download of zip file to the build process. Closes #15

### LINKED ISSUE #15: Package into distributable exe

bjedrzejewski:
No need to say anything. Proof is in the pudding.

### DIFF
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -152,6 +152,7 @@ module.exports = function(grunt) {
         grunt.task.run('copy');
     });
 
-    grunt.registerTask('serve', ['showcase', 'less:development', 'connect:livereload', 'openfin:serve', 'watch']);
-    grunt.registerTask('build', ['showcase', 'less:development', 'connect:livereload', 'download', 'openfin:serve', 'watch']);
+    grunt.registerTask('build', ['showcase', 'less:development', 'connect:livereload']);
+    grunt.registerTask('serve', ['build', 'openfin:serve', 'watch']);
+    grunt.registerTask('createZip', ['build', 'download']);
 };
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -102,6 +102,12 @@ module.exports = function(grunt) {
                 }
             }
         },
+        download: {
+            foo: {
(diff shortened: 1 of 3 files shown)


gruntfile.js
```diff
@@ -102,6 +102,12 @@ module.exports = function(grunt) {
                 }
             }
         },
+        download: {
+            foo: {
+                src: ['https://dl.openfin.co/services/download?fileName=OpenFinD3FC&config=http://localhost:5000/app.json'],
```
owennw: I don't think we want a 'foo'.

Do you think it's worth parameterising the port number and project name too?

gruntfile.js
```diff
@@ -145,6 +152,6 @@ module.exports = function(grunt) {
         grunt.task.run('copy');
     });
 
-    grunt.registerTask('serve', ['showcase', 'less:development', 'connect:livereload', 'openfin:serve', 'watch']);
+    grunt.registerTask('serve', ['showcase', 'less:development', 'connect:livereload', 'download', 'openfin:serve', 'watch']);
     grunt.registerTask('build', ['openfin:build', 'showcase']);
```
owennw: Does this mean that everytime we run `serve` we'll be connecting to a remote server to download the zip?

Perhaps it would be better to only do that when doing a build or release?

gruntfile.js
bjedrzejewski: Will change foo, I wouldn't worry about parametrizing ports etc

gruntfile.js
bjedrzejewski: Yes, we can have another task if you prefer. Now the zip download is every time.

gruntfile.js
```diff
@@ -146,5 +153,5 @@ module.exports = function(grunt) {
     });
 
     grunt.registerTask('serve', ['showcase', 'less:development', 'connect:livereload', 'openfin:serve', 'watch']);
-    grunt.registerTask('build', ['openfin:build', 'showcase']);
+    grunt.registerTask('build', ['showcase', 'less:development', 'connect:livereload', 'download', 'openfin:serve', 'watch']);
 };
```
owennw: One more small thing:

Perhaps the `build` task could call the `serve` task? There seems to be a lot of duplication between the two.

PR #32, opened 2015-12-08
Implement search (part 1)

owennw:
This first commit enables searching with the search bar. This can be achieved by pressing enter while the input field has focus, or by clicking the search button. The user will be taken to a 'searching' page while the application searches for the requested data. When the data is found they will be redirected to the page showing the data.

The input is not sanitised at all.

Currently we are not connecting to anything, but instead I've included a dummy 1s delay, then showing them the input search as a bullet point.

Internally this introduces ngRoute, and the scaffolding of other controllers and html pages.

One thing to note is that I've not organised the file structure at all: I don't want to commit us to a structure this early while the project is still growing.

N.B. Please don't close this branch, I'll continue to use it to connect to quandl and bring back real data.

### DIFF
--- a/public/app.js
+++ b/public/app.js
@@ -3,17 +3,29 @@
 
     angular.module('OpenFinD3FC', [
         'ngRoute',
-        'openfin.showcase'
+        'openfin.thumbnails', 
+        'openfin.search',
+        'openfin.stock'
     ])
     .config(['$routeProvider', function($routeProvider) {
         $routeProvider
             .when('/', {
                 redirectTo: 'stocks'
             })
-            .when('/stocks', {
-                templateUrl: 'stocks.html',
-                controller: 'ShowcaseCtrl',
-                controllerAs: 'showcaseCtrl'
+            .when('/stocks/', {
+                templateUrl: 'thumbnails.html',
+                controller: 'ThumbnailsCtrl',
+                controllerAs: 'thumbnailsCtrl'
+            })
+            .when('/stock/:stock', {
+                templateUrl: 'stockData.html',
+                controller: 'StockCtrl',
+                controllerAs: 'stockCtrl'
+            })
+            .when('/search/:query', {
+                templateUrl: 'search.html',
+                controller: 'SearchCtrl',
+                controllerAs: 'searchCtrl'
             })
             .otherwise({
                 redirectTo: function() {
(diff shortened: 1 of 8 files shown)


public/app.js
```diff
@@ -3,17 +3,29 @@
...
+            .when('/stock/:stock', {
+                templateUrl: 'stockData.html',
+                controller: 'StockCtrl',
+                controllerAs: 'stockCtrl'
+            })
+            .when('/search/:query', {
+                templateUrl: 'search.html',
+                controller: 'SearchCtrl',
+                controllerAs: 'searchCtrl'
             })
```
bjedrzejewski: Maybe make directory structure match your routes? This can provide some logic? This is suggestion for the future, not necessary this PR.

public/index.html
```diff
@@ -16,19 +20,23 @@ <h4 class="text-center">Altogether Smarter</h4>
         <div class="row">
             <div class="col-md-offset-3 col-md-6 col-xs-12">
                 <div class="input-group" style="width: 100%">
-                    <input class="form-control" type="text" />
-                    <br>
-                    <br>
-                    <button type="button" class="btn btn-primary center-block">Search</button>
+                    <form class="form-horizontal" ng-submit="searchCtrl.submit()">
+                        <input type="text" class="form-control" placeholder="Stock name" ng-model="searchCtrl.query"/>
+                        <br />
```
bjedrzejewski: HTML5 prefers <br> over <br />

bjedrzejewski: http://stackoverflow.com/a/1946446/1611957

public/stockData.html
```diff
@@ -0,0 +1,5 @@
+﻿<div class="container-fluid">
+    <ul>
+        <li>{{stockCtrl.stock}}</li>
+    </ul>
+</div>
```
bjedrzejewski: End with blank? Or we only care about js for that?

public/index.html
owennw: ... Didn't know that! will fix.

public/stockData.html
owennw: I think that's just js.

bjedrzejewski: ok

public/index.html
bjedrzejewski: I know its like html5 pub quiz trivia

owennw:
Let me know when it's okay and I'll squash.

PR #34, opened 2015-12-08
Implement search (part 2: quandl service)

owennw:
Now the app will connect to quandl, using the input string as a search and return a bullet pointed list of results.

### DIFF
--- a/public/quandl-service.js
+++ b/public/quandl-service.js
@@ -1,0 +1,10 @@
+(function() {
+    'use strict';
+
+    angular.module('openfin.quandl', ['ngResource'])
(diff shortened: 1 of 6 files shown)


public/quandl-service.js
```diff
@@ -0,0 +1,10 @@
+(function() {
+    'use strict';
+
+    angular.module('openfin.quandl', ['ngResource'])
+        .factory('quandlService', ['$resource', function($resource) {
+            return $resource('https://www.quandl.com/api/v3/datasets.json?query=:query&database_code=WIKI', {}, {
```
bjedrzejewski: This will run out of requests pretty quickly. Shall we put here some api key?

PR #52, opened 2015-12-09
Thumbnails with pointer cursor, closes #31

### LINKED ISSUE #31: Use hand cursor over most visited thumbnails

bjedrzejewski:


public/thumbnails.html
```diff
@@ -1,7 +1,7 @@
 ﻿<div class="container-fluid">
     <div class="row">
         <div class="col-xs-4" ng-repeat="stock in thumbnailsCtrl.stocks">
-            <a class="thumbnail bg-primary" ng-click="thumbnailsCtrl.open(stock)">
+            <a class="thumbnail bg-primary" ng-click="thumbnailsCtrl.open(stock)" style="cursor: pointer">
                 <div style="color:black">
```
owennw: Would we rather do this in less/css?

bjedrzejewski: We could target all `a .thumbnail` maybe?

owennw: Maybe. Currently the logo at the top of the page, functioning as a home link, doesn't have the hand cursor. I'm not sure (!) if it should. What do you think?

PR #59, opened 2015-12-09
Revert back to R1

owennw:
I think I've got it all. Is this the kind of thing we were thinking about?

Would appreciate it if you have a detailed comb through this and what's left just to make sure. If it's acceptable I propose we both do the merge and walk through it.


public/app.json
```diff
@@ -14,7 +14,7 @@
     "shortcut": {
         "company": "OpenFin",
         "description": "Hosts the d3fc-showcase as an OpenFin application.",
-        "icon": "http://owennw.github.io/OpenFinD3FC/favicon.ico",
+        "icon": "http://localhost:5000/favicon.ico",
         "name": "OpenFinD3FC"
     }
 }
```
bjedrzejewski: Use the gh-pages not local here please. :)

PR #65, opened 2015-12-10
Whitespace fix

owennw:
I think the whitespace for this file got borked after our messing around creating a release and a new branch.


bjedrzejewski:
Are we versioning this file in the end? I thought we bring it with grunt

owennw:
maybe you're right and we shouldn't have any of the showcase and its assets versioned. However, they all currently are (and are in release 1 too...).

bjedrzejewski:
So at the moment its just the .js and .css we are building right? Html we are versioning?

owennw:
At the moment it's arguably a bit messed up. We're building the showcase and copying it into our public folder, where it gets renamed and used. Everything from that build is versioned.. *apart* from the styles.css as we've excluded that. Doesn't make a great deal of sense. 

Perhaps we should unversion and remove all the built showcase files. The reason we versioned them originally is because we weren't sure if we were going to be making changes to them before/after building them (including hooks or buttons etc), but we haven't done that yet... YAGNI...

bjedrzejewski:
ok so i am just making another pr to cleanup a bit, as this keeps tripping my git

bjedrzejewski:
Made redundant by #66

PR #66, opened 2015-12-10
Cleanup of css and fonts, fix app.json

bjedrzejewski:


bjedrzejewski:
All the stuff here gets built anyway including app.json

gruntfile.js
```diff
@@ -1,7 +1,7 @@
 /* jshint node: true*/
 'use strict';
 module.exports = function(grunt) {
-    var target = grunt.option('target') || 'http://localhost:5000',
+    var target = grunt.option('target') || 'http://owennw.github.io/OpenFinD3FC',
         port = process.env.PORT || 5000,
```
owennw: Is this the bit that determines the settings in the app.json?

.gitignore
```diff
@@ -1,4 +1,6 @@
 node_modules
-public/assets/css/style.css*
 public/OpenFinD3FC.zip
-.idea
\ No newline at end of file
+.idea
+public/assets/fonts
+public/assets/css
+public/app.json
```
owennw: why not just ignore `public/assets/**/*`?

gruntfile.js
bjedrzejewski: Yes, so we can have another issue so that we can build either local or public

owennw:
Given we're now ignoring more of the public/assets folder it would make sense to remove those files too.

.gitignore
bjedrzejewski: That will ignore all the files. I hope to have a directory structure there once the app gets bigger?

bjedrzejewski:
So you want to move your angular stuff to src?

bjedrzejewski:
And fully build public?

.gitignore
owennw: It will only ignore the files in the assets folder, and we don't have anything of our own in there at the moment right?

bjedrzejewski: My bad, you are right

owennw:
We can do that, but I think that's orthogonal to what's in this PR. If we're adding files to the .gitignore because they get built then there's no point leaving historic versions of them in the repo.

.gitignore
bjedrzejewski: Can you commit that fix?

owennw: Sure.

bjedrzejewski: Actually I will do it

bjedrzejewski:
@owennw when it passes you can merge I hope :)

owennw:
The app.json should be deleted, and the d3fc-showcase.html should be deleted and ignored.

In general anything in the .gitignore should not be in the repo: if it is it should be removed.

owennw:
Looks good. A bit frustrating that running the grunt serve with a target changes the app.json file, but we'll just remember to revert it. Can you squash these commits then I'll merge?

bjedrzejewski:
great, will do now

bjedrzejewski:
@owennw after CI, ready to merge

PR #77, opened 2015-12-14
Fetch Quandl financial data (closes #48)

### LINKED ISSUE #48: Fetch stock data from quandl for corresponding search

owennw:
Fetches financial data. Note the UI has not been designed, so I've included a skeleton just to show that data is being received, asynchronously.

As discussed first all the search results for the query are populated, and when they are all fetched the extra data for them all is then requested and displayed.

### DIFF
--- a/public/quandl-service.js
+++ b/public/quandl-service.js
@@ -3,8 +3,63 @@
 
     angular.module('openfin.quandl', ['ngResource'])
         .factory('quandlService', ['$resource', function($resource) {
-            return $resource('https://www.quandl.com/api/v3/datasets.json?api_key=kM9Z9aEULVDD7svZ4A8B&query=:query&database_code=WIKI', {}, {
-                get: { method: 'GET', cache: true }
-            });
+            var apiKey = 'api_key=kM9Z9aEULVDD7svZ4A8B';
+
+            function stock() {
+                return $resource('https://www.quandl.com/api/v3/datasets.json?' + apiKey + '&query=:query&database_code=WIKI', {}, {
+                    get: { method: 'GET', cache: true }
+                });
+            }
+
+            function stockData() {
+                var startDate = moment().subtract(1, 'weeks').format('YYYY-MM-DD'),
+                    json;
+
+                return $resource('https://www.quandl.com/api/v3/datasets/WIKI/:code/data.json?' + apiKey + '&start_date=' + startDate, {}, {
+                    get: {
+                        method: 'GET',
+                        transformResponse: function(data, headers) {
+                            json = angular.fromJson(data);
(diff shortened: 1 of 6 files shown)


public/quandl-service.js
```diff
@@ -3,8 +3,63 @@
...
+                    json;
+
+                return $resource('https://www.quandl.com/api/v3/datasets/WIKI/:code/data.json?' + apiKey + '&start_date=' + startDate, {}, {
+                    get: {
+                        method: 'GET',
+                        transformResponse: function(data, headers) {
+                            json = angular.fromJson(data);
+                            processResponse(json);
+                            return json;
+                        },
```
bjedrzejewski: I guess we don't care about stuff being immutable in JS.

public/quandl-service.js
```diff
@@ -3,8 +3,63 @@
...
+                return {
+                    date: data[0],
+                    open: data[1],
+                    high: data[2],
+                    low: data[3],
+                    close: data[4],
+                    volume: data[5]
+                };
+            }
+
```
bjedrzejewski: Can we maybe use something other than indexes? Or are indexes the only available thing?

owennw: This is the shape of the data:

`"data": [
[
    "2015-12-11",
    ​45.44,
    ​46.59,
    ​45.44,
    ​45.59,
    ​385557.0,
    ​0.0,
    ​1.0,
    ​45.44,
    ​46.59,
    ​45.44,
    ​45.59,
    ​385557.0`

So probably not.

However, we can declare const variable for the indices and use them instead..

public/stock-controller.js
```diff
@@ -4,12 +4,39 @@
...
            });
+
+           function fetchStockData(stock) {
+               quandlService.stockData().get({ code: stock.code }, function(result) {
+                   var data = result.stockData;
+                   stock.data = data.data;
+                   stock.startDate = data.startDate;
+                   stock.endDate = data.endDate;
+               });
+           }
```
bjedrzejewski: Isn't it better to pass the code and get a stock back, rather than pass stock with .code and get the mutation again?

public/quandl-service.js
bjedrzejewski: Cons't variable is probably the way to go, as we will be using these in more than one place I imagine?

public/stock-controller.js
owennw: There's absolutely no reason to re-fetch the stock here, you're right. Bit of a copy/paste. I'll fix.

public/quandl-service.js
owennw: Shall do.

PR #79, opened 2015-12-14
SVG changed, window working. Fixes #75

### LINKED ISSUE #75: Window does not open after clicking on favourites

bjedrzejewski:
This is fix for 75, still working on 62, don't close the branch.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -17,7 +17,7 @@
         <div class="row">
             <div class="col-sm-12">
                 <a ng-href="#/" class="svg logo" ng-click="searchCtrl.clear()">
-                    <object type="image/svg+xml" data="logo.svg">Your browser does not support SVG</object>
+                    <img src="logo.svg"></img>
                 </a>
             </div>
         </div>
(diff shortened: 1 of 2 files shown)


public/index.html
```diff
@@ -17,7 +17,7 @@
         <div class="row">
             <div class="col-sm-12">
                 <a ng-href="#/" class="svg logo" ng-click="searchCtrl.clear()">
-                    <object type="image/svg+xml" data="logo.svg">Your browser does not support SVG</object>
+                    <img src="logo.svg"></img>
                 </a>
             </div>
         </div>
```
owennw: Okay since we're only using a single (known) browser, but worth keeping in mind: http://www.sitepoint.com/add-svg-to-web-page/

owennw: Specifically:

>For security reasons, browsers will disable SVG scripts, linking and other types of interactivity when they’re added to your page with an img tag. In addition, IE9, Chrome and Safari won’t apply stylesheet rules to the SVG if they’re defined in a separate CSS file.

PR #81, opened 2015-12-14
Save list of searched stocks (part 1)

owennw:
This is the first part, creating a first time list of visited stocks (yet to be decided), and incrementing how many times each has been clicked, to re-order them on reload. Also internally there's now a limit of thumbnails (set to 9).

This makes use of angular-storage.

The next part will be to add new stocks to the list, after #49.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -43,10 +43,12 @@
     <script src="https://cdnjs.cloudflare.com/ajax/libs/angular.js/1.4.8/angular-route.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/angular.js/1.4.8/angular-resource.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.10.6/moment.min.js"></script>
+    <script src="https://cdn.rawgit.com/auth0/angular-storage/master/dist/angular-storage.js"></script>
     <script src="app.js"></script>
     <script src="thumbnails-controller.js"></script>
     <script src="search-controller.js"></script>
     <script src="stock-controller.js"></script>
+    <script src="store-service.js"></script>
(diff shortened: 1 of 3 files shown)


public/index.html
```diff
@@ -43,10 +43,12 @@
     <script src="https://cdnjs.cloudflare.com/ajax/libs/angular.js/1.4.8/angular-route.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/angular.js/1.4.8/angular-resource.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.10.6/moment.min.js"></script>
+    <script src="https://cdn.rawgit.com/auth0/angular-storage/master/dist/angular-storage.js"></script>
```
bjedrzejewski: Any reason for different cdn?

owennw: Nope

public/store-service.js
```diff
@@ -0,0 +1,53 @@
...
+                }
+
+                tuples = tuples.sort(function(a, b) {
+                    return b[1] - a[1];
+                });
+
+                return tuples.map(function(x) {
+                    return x[0];
+                });
+            }
```
bjedrzejewski: Comment or a slightly more explicit naming would be helpful here. Is it simply super generic code?

owennw: It's pretty generic -- it's just a sort function for an object (pseudo associative array). Any ideas on better naming?

bjedrzejewski: Looks like a super generic code for a controller. I don't know js best practices- do we need a  library or something? Anyway if you are happy with it, we can merge.

owennw:
Urg don't merge yet, a file crept in that shouldn't be there.

PR #84, opened 2015-12-14
Open showcase from search, save searched stocks

owennw:
Allow showcase to be opened from search page. Store searched stocks locally and display as thumbnails.

### DIFF
--- a/public/stock-controller.js
+++ b/public/stock-controller.js
@@ -1,42 +1,47 @@
 (function() {
     'use strict';
 
-    angular.module('openfin.stock', ['openfin.quandl'])
-       .controller('StockCtrl', ['$routeParams', 'quandlService', function($routeParams, quandlService) {
-           var self = this;
-           self.stocks = [];
+    angular.module('openfin.stock', ['openfin.quandl', 'openfin.openfin'])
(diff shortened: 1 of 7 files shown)


public/stock-controller.js
```diff
@@ -1,12 +1,12 @@
 (function() {
     'use strict';
 
-    angular.module('openfin.stock', ['openfin.quandl'])
-       .controller('StockCtrl', ['$routeParams', 'quandlService', function($routeParams, quandlService) {
+    angular.module('openfin.stock', ['openfin.quandl', 'openfin.openfin'])
+       .controller('StockCtrl', ['$routeParams', 'quandlService', 'openfinService', function($routeParams, quandl, openfin) {
```
bjedrzejewski: Smart people say that you shouldn't rename your controllers. Source: https://github.com/mgechev/angularjs-style-guide . This is in a few more places.

Quote:

>If using array definition syntax, use the original names of the controller's dependencies. This will help you produce more readable code:

owennw: As in the name of the input parameters in `function( ... )` should be the same as the injected names before in the array?

bjedrzejewski: Like: `quandlService` rather than `quandl`

owennw: Sure, will do.

Issue #86, opened 2015-12-15
Build fails ...


ColinEberhardt:
After running `npm install` then running the build ...

```
C:\projects\OpenFinD3FC>grunt build
Running "showcase" task
Loading "Gruntfile.js" tasks...ERROR
>> Error: Cannot find module 'time-grunt'
Warning: Task "build" not found. Use --force to continue.

Aborted due to warnings.

Running "copy:main" (copy) task


Running "connect:livereload" (connect) task
Started connect web server on http://0.0.0.0:5000

Done, without errors.
```

owennw:
Did you run `npm install` in the d3fc-showcase folder?

owennw:
Which admittedly isn't in our documentation...

ColinEberhardt:
Just spotted that - the dev dependencies are needed ...

bjedrzejewski:
The issue could be- add to documentation and/or automate npm install inside d3fc-showcase inside modules

PR #98, opened 2015-12-15
Travis no longer need to npm install d3fc-showcase

ColinEberhardt:
See #86


bjedrzejewski:
We already make that change on dev, but it will be good to keep master travis build look correct as well.

PR #99, opened 2015-12-15
Build to public

owennw:
Move our files into src, clean before each build (including clean the dist folder in the showcase), and copy the files across.


bjedrzejewski:
I will build it and try it first

PR #110, opened 2015-12-21
Introduce sidebar with search and tearout

owennw:
Bit of a large one sorry, but it is split into separate commits to make it a bit easier.


owennw:
Not now. Tearing out it still unstable.

PR #122, opened 2015-12-30
TODO items while Nick is away. Closes #117

### LINKED ISSUE #117: Refresh favourites list

bjedrzejewski:


bjedrzejewski:
This so far takes care of:

1. the favourites list doesn't refresh when the list is updated on the search tab. This doesn't seem trivial to do (at least I haven't thought of an easy way yet) but there is a greater question of why should be favourites be customisable from the search tab? It doesn't feel like the right place to me so we might have to take that question back to Rui.

and

3. The search is a bit greedy, in that as I showed you if you quickly type a few letters because of the asynchronicity of the search all the results will be added at once. I don't know if there's a cool way of cancelling or ignoring the last search and just running the latest one

bjedrzejewski:
Done:
4- When a search result or favourite tile is selected it doesn't change colour, as in the design mock-ups

Now I will work on the sparkline.

bjedrzejewski:
All that is left is to do the sparkline I think. I will use the area chart to make it.

src/assets/styles/style.less
```diff
@@ -35,6 +35,10 @@ a.svg:after {
     opacity: 1;
 }
 
+.darkens:hover {
+    background: #1A1F26;
+}
+
```
owennw: I think there's a less variable for this colour defined at the top.

src/index.html
```diff
@@ -13,7 +14,7 @@
     </style>
 </head>
 <body ng-cloak ng-controller="SearchCtrl as searchCtrl">
-    <div class="container-fluid main">
+    <div class="container-fluid main"  ng-controller="FavouritesCtrl as favouritesCtrl">
         <div class="sidebars" ng-controller="SidebarCtrl as sidebarCtrl">
```
owennw: double space!

src/sidebars/favourites/favourite-preview.html
```diff
@@ -12,7 +13,10 @@ <h4 class="pull-left"><strong>{{stock.code | uppercase}}</strong></h4>
                 </div>
             </div>
             <div class="row">
-                <div class="col-xs-2 col-xs-offset-10">
+                <div class="col-xs-9" style="padding-left: 0px">
+                    <svg id="{{stock.code}}chart" width="100%" height="65px" ng-init="renderChart(stock)"></svg>
+                </div>
```
owennw: I've read in the past that one should be wary of using ng-init (https://docs.angularjs.org/api/ng/directive/ngInit) -- this may well be a good use of it though. What do you think?

src/sidebars/favourites/favourites.js
```diff
@@ -21,38 +21,53 @@
             };
 
             self.bellStyle = function(stock) {
-                return stock.notification ? bellColours.on : bellColours.off;
+                if(stock.notification){
+                    return bellColours.on;
```
owennw: small whitespace with the `if`

src/sidebars/favourites/favourites.js
```diff
@@ -21,38 +21,53 @@
...
+                } else if (stock.isHovered){
+                    return bellColours.offhover;
+                }  else
+                {
+                    return bellColours.off;
```
owennw: Would prefer the bracket after the `else` to be on the same line.

src/sidebars/favourites/favourites.js
```diff
@@ -21,38 +21,53 @@
...
-                    var data = stock.data[0],
-                        price,
-                        delta,
-                        percentage;
+            self.updateFavourites = function() {
+                self.stocks = [];
+                self.favourites = storeService.get();
+                self.favourites.map(function (favourite) {
+                    quandlService.getData(favourite, function (stock) {
+                        var data = stock.data[0],
```
owennw: bit of whitespace with the `function (...) {` here.

src/sidebars/search/search.js
```diff
@@ -10,11 +10,19 @@
...
-                    return stock.favourite ? favouriteColours.on : favouriteColours.off;
+                    if(stock.favourite){
+                        return favouriteColours.on;
+                    } else if (stock.isHovered){
+                        return favouriteColours.offhover;
+                    }  else
+                    {
+                        return favouriteColours.off;
+                    }
                 };
```
owennw: same whitespace and bracket as above

src/sidebars/search/search.js
```diff
@@ -33,8 +41,19 @@
...
+                            var i;
+
+                            //removing stocks found with old query
+                            self.stocks = self.stocks.filter(function(stock, i) {
+                                return stock.query === self.query;
+                            });
+                            //not adding old stocks
+                            if(stock.query !== self.query){
+                                return;
+                            }
```
owennw: bit more whitespace with the `if`

owennw:
Mostly whitespace! Haven't tried the code myself (I'm not even meant to be looking at it right now!) but I assume it's functional!

bjedrzejewski:
Yeah, it seems that more aggressive linting would be nice in the html given how many omissions I made here :)

src/assets/styles/style.less
bjedrzejewski: thanks, didnt see it!

src/sidebars/favourites/favourite-preview.html
bjedrzejewski: I think we are good. We are not putting extra logic in template and we need to execute something the moment this svg becomes available. The other solution would be to make a directive here, but directive inside directive may be a small overkill for now. If the logic for this grows, we could think about it.

bjedrzejewski:
Now the automatic of favs doesn't work- console errors. Once this is fixed, its just the styling left.

PR #137, opened 2016-01-05
Favourite icon

owennw:
Adds the favourite icon to the favourite tiles.

### DIFF
--- a/src/sidebars/favourites/favourites.js
+++ b/src/sidebars/favourites/favourites.js
@@ -103,7 +103,8 @@
                                         price: price,
                                         delta: delta,
                                         percentage: Math.abs(percentage),
-                                        notification: false
+                                        notification: false,
+                                        favourite: true
                                     });
                                 }
                             });
--- a/src/sidebars/favourites/favourites.js
+++ b/src/sidebars/favourites/favourites.js
@@ -63,6 +63,10 @@
                     );
                 };
 
+                self.remove = function(stock) {
+                    return confirm('Are you sure you wish to remove this stock (' + stock.code + ') from your favourites?');
+                };
+
                 self.update = function() {
                     self.favourites = storeService.get();
                     var i,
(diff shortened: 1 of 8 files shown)


bjedrzejewski:
@owennw will review this now

src/sidebars/star.js
```diff
@@ -0,0 +1,49 @@
...
+    'use strict';
+
+    angular.module('openfin.star', ['openfin.store'])
+        .controller('StarCtrl', ['$scope', 'storeService', function($scope, storeService) {
+            var self = this;
+
+            var favouriteColours = {
+                on: '#42D8BD',
+                off: '#1A1F26',
+                offhover: "#263337"
```
bjedrzejewski: Shall we make it to css classes already? Colours in JS are bad.

src/sidebars/favourites/favourites.js
```diff
@@ -2,117 +2,126 @@
...
-                            }
-                        });
-                    }
-                })
-            };
+                self.update();
 
-            self.updateFavourites();
-        }])
+                $scope.$on('favouriteChanged', function(event, data) {
```
bjedrzejewski: Do we have to use the dreaded $scope?

owennw: When the icon is clicked we need to
a) remove the local storage entry
b) remove the favourite from the search bar
c) remove the favourite from the favourites

The only ways I can think of doing all of these are to
a) call several functions from the html ng-click on different (unrelated?) controllers
b) emit an event and catch it in the relevant places.

I opted for the event as I felt it was a valid use of it, and leads to tidier code -- I hope! And to do that we need to use `$scope`.

Can you think of another way to either use an event, or not use the event at all?

src/sidebars/star.js
owennw: Yes this should be improved I think. These colours are all already defined in our .less file anyway. Thankfully this refactor has brought all of these colours into one file. Should this be done as part of this PR or as a later tidy? I've already created an issue (#139) to tidy around the less file.

src/sidebars/favourites/favourites.js
bjedrzejewski: It actually is more cross cutting than i thought. I think event is the right way to go.

PR #155, opened 2016-01-07
Make search results clickable (move selection into main-controller)

owennw:
First step of #154 . The controllers in the index file were all over the place, so I've moved them around so they're covering the correct (and obvious) scopes. 

It didn't make sense to have the sidebar controller knowing the selected stock as the showcase needs to know that, and it has nothing to do with the sidebar. Hence the introduction of a new, main, controller.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -3,6 +3,7 @@
 
     angular.module('OpenFinD3FC', [
         'ngAnimate',
+        'openfin.main',
         'openfin.showcase',
         'openfin.search',
         'openfin.favourites',
--- a/src/index.html
+++ b/src/index.html
@@ -13,10 +13,10 @@
         }
     </style>
 </head>
-<body ng-cloak ng-controller="SearchCtrl as searchCtrl">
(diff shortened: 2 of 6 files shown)


owennw:
I also removed the clearing of the search by clicking the main logo.. for some reason. In any case, that logo is leaving soon -- what do you think?

bjedrzejewski:
The logo was not really going- it was moving up right and staying clickable.

owennw:
Ah yes I see why. The search controller having scope over that logo didn't make sense either. Where does that responsibility lie?

bjedrzejewski:
This is tricky. Should that be an event? Clean selection / reload page? Should clicking on the logo simply reload page actually?

owennw:
I do feel like clicking the logo should bring the app back into a similar state to how it started:
remove search, unselect a stock, return to favourites tab.

Not sure how it should be done right now.. With regards to this PR though what do you think should be done? The current scoping won't allow the old implementtion.

bjedrzejewski:
Yeah this is fine for PR. Make issue to do that logo click thing. I think i know how to do it with routing, but that would be scope creep

PR #169, opened 2016-01-11
Added scrolling

bjedrzejewski:


src/assets/styles/scroll.less
```diff
@@ -0,0 +1,40 @@
...
+
+.searchScroll .mCSB_scrollTools {
+  left: 2px;
+  margin-top: 10px;
+  margin-bottom: 10px;
+}
+
+.favScroll {
+  top: 52px;
+  bottom: 0px;
```
owennw: Is this 52px number calculated somehow, from pre-existing figures? It is also repeated -- I assume the places it's used are coupled?

src/assets/styles/variables.less
```diff
@@ -18,4 +18,4 @@
 @expanded-width: 300px;
 
 // Heights
-@splitter-height: 2px;
\ No newline at end of file
+@splitter-height: 1px;
```
owennw: Why has this changed?

src/assets/styles/scroll.less
owennw: In fact, search scroll and favourite scroll look to be identical -- can't we just use one class?

src/assets/styles/variables.less
bjedrzejewski: to look more like the design

src/assets/styles/scroll.less
bjedrzejewski: at the moment they are identical, we need 2 classes to target them with javascript separately. could add 3rd class for styling. or have class for styling target with ids

src/index.html
```diff
@@ -57,6 +62,16 @@
...
+    <script src="http://ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js"></script>
+    <script src="http://cdn.jsdelivr.net/jquery.mcustomscrollbar/3.0.6/jquery.mCustomScrollbar.concat.min.js"></script>
+    <script>
+        (function($){
+            $(window).load(function(){
+                $('.searchScroll').mCustomScrollbar();
+                $('.favScroll').mCustomScrollbar();
+            });
+        })(jQuery);
+    </script>
```
owennw: I take it we have to use jQuery? I think I'd sooner see this script in a separate js file, feels neater. Do you agree?

src/assets/styles/favourites.less
```diff
@@ -1,5 +1,5 @@
 ﻿@favourite-height: 150px;
-@favourite-padding-left: 0px;
+@favourite-padding-left: 20px;
 @favourite-padding-top: 5px;
 @favourite-padding-right: 5px;
```
owennw: There isn't a corresponding change for search.less?

src/assets/styles/variables.less
owennw: The design did specify a 2px thick splitter (at least, that's what Rui said before)

bjedrzejewski:
Build failed randomly, will look into that

src/assets/styles/scroll.less
owennw: If jQuery needs two classes that's okay, but the css can still be tidied, e.g. use `.favScroll, .searchScroll` in the two duplicated places.

src/assets/styles/favourites.less
bjedrzejewski: Search already is padded out by 20px on details

src/assets/styles/scroll.less
bjedrzejewski: Yeah, I refactored, used 2 separate ids and one class. Originally these were different.

src/assets/styles/variables.less
bjedrzejewski: Ok, will revert then. His design definitely has 1px splitter though (on the pdfs)

src/index.html
bjedrzejewski: Yeah, we need to use jQuery. There was never any hate for jQuery when jQuery is needed. Colin (and people) just warn others not to use jQuerry for whats Angular job.

bjedrzejewski: And yeah, will move the script now

src/assets/styles/style.less
```diff
@@ -8,6 +8,7 @@
 @import "favourites.less";
 @import "star.less";
 @import "showcase-changes.less";
+//@import "custom-scroll.less";
```
owennw: I don't think we want this?

bjedrzejewski: i just removed that

bjedrzejewski: i did it when trying to fix that issue

owennw:
scroll_bug
1. I think moving the chart 20px right isn't good enough, as now the numbers if big enough go through the tile as shown. 

2. When the search bar is open, clicking on the far left of the favourites tab won't select it, as that's where the hidden scroll bar is that stops the bar being selected.

3. I can't see the design because I was kicked from the conversation.. but I'm not sure if the tile splitters should be visible to the left of the scroll bars?

bjedrzejewski:
1. Ok, will fix it, just padding on the title.
2. This is not an issue.
3. Thats what the design shows

bjedrzejewski:
@owennw thanks for thorough review, the latest version should be ok

owennw:
Just two more things:

1. Should the cursor on the scroll bar still be the hand? I would have thought the default pointer was more usual. (Perhaps a question for a UX person?)
2. On both of the tabs when the scrollbar is present use the scroll wheel with the mouse over the centre of the tab to scroll through the results. Notice that a) the highlighted result does not change and b) when the scrolling stops the highlighted row is only updated when the mouse is moved slightly.

I don't know how large an issue (2) is, so might be best spun off as a different issue?

bjedrzejewski:
1. I actually prefer the hand here, we can later raise as an ux question.
2. This is difficult to fix, so I would make it another issue if we want to fix it. Somehow we need to fire some events on mouse scroll?

owennw:
Okay can you raise those two issues?

PR #174, opened 2016-01-11
Move splitters out of tilesand add to the end

owennw:
I don't think the splitters belong in the search and favourite files, so I've split (haha) them out. Also reinstated the ones at the bottom of the lists.


bjedrzejewski:
@owennw squash commits :)

PR #175, opened 2016-01-12
Integrate d3fc-showcase as a module; remove iframe

jleft:
This is still a bit rough, but I thought it was worth sharing. Hopefully this is heading in the right direction!

It's using the [jleft/d3fc-showcase#openfin-integration/](https://github.com/jleft/d3fc-showcase/tree/openfin-integration)

The showcase can now be exported as a module, using `grunt:module` with a _very_ simple API

```js
// Create a instance of the showcase app
var chart = showcase.app();

// Renders the app into the specified element
chart.run(element);

// Given a string of a Quandl product, this will change the product which the chart is displaying
chart.changeQuandlProduct(productString);

// Thinking of removing this - it allows the callback to manipulate the model
// very flexible, but (hopefully) YAGNI
chart.updateModel(callback);
```

### DIFF
--- a/src/showcase/showcase-directive.js
+++ b/src/showcase/showcase-directive.js
@@ -1,4 +1,4 @@
-(function() {
+(function(sc) {
     'use strict';
 
     angular.module('openfin.showcase', [])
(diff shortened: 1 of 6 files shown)


src/showcase/showcase-directive.js
```diff
@@ -8,7 +8,24 @@
...
+                },
+                link: function(scope, element) {
+                    var chart = sc.app(),
+                        firstRun = false;
+
+                    scope.$watch('selection', function(newSelection, previousSelection) {
+                        if (newSelection !== '') {
+                            if (!firstRun) {
+                                firstRun = true;
+                                chart.run(element[0].children[0]);
```
owennw: Surely `firstRun` would be true.. and then set to false?

jleft: Yes, that makes more sense - updated :)

bjedrzejewski:
@jleft all looks good to me, will you make any more changes? Rename the commit and we can merge.

jleft:
Thanks @bjedrzejewski, I don't intend to make any more changes to this PR unless any issues are noticed.

I'll work on getting https://github.com/ScottLogic/d3fc-showcase/pull/536 merged in d3fc-showcase fully, then when the next showcase release lands, that can be integrated (shouldn't need to be any changes to the code here, if there are, then it should be very minor).

bjedrzejewski:
Hi @jleft  once you get it integrated, give us a shout and we will update our version as well

PR #194, opened 2016-01-12
Split directives

owennw:
Seem okay?


bjedrzejewski:
Collapse it into a single commit please

PR #197, opened 2016-01-12
Move .less files into appropriate folders

owennw:


src/assets/styles/style.less
```diff
@@ -2,13 +2,13 @@
...
 
 @import "variables.less";
-@import "logo.less";
-@import "sidebar.less";
-@import "search.less";
-@import "favourites.less";
-@import "star.less";
-@import "showcase-changes.less";
-@import "custom-scroll.less";
+@import "../../logo.less";
```
bjedrzejewski: logo should stay where it is, droping it into main dir is just bad

src/assets/styles/style.less
```diff
@@ -2,13 +2,13 @@
...
+@import "../../sidebars/sidebar.less";
+@import "../../sidebars/search/search.less";
+@import "../../sidebars/favourites/favourites.less";
+@import "../../sidebars/star/star.less";
+@import "../../showcase/showcase-changes.less";
+@import "../../custom-scroll.less";
```
bjedrzejewski: Same here, I suggest no main dir as it is messy

src/assets/styles/style.less
owennw: Yeah I wasn't sure whether it should move or not.. Is that the 'correct' pattern?

src/assets/styles/style.less
owennw: There is a customScroll controller in the main dir though..

owennw: I don't know if that then means it deserves its own folder.. ?

bjedrzejewski: I think this means that the .js should be moved rather than the other way round

bjedrzejewski: But one issue at a time

src/assets/styles/style.less
bjedrzejewski: The pattern is- keep the style in assets, unless it directly relates to a directive.

PR #201, opened 2016-01-12
New app frame with dragging as per design.

bjedrzejewski:


bjedrzejewski:
I also adjusted min window size o make it minimal that still looks ok with 3 favourites and new app-bar.

src/assets/styles/toolbar.less
```diff
@@ -0,0 +1,39 @@
...
+}
+
+.min-button {
+  position: fixed;
+  top: 15px;
+  right: 45px;
+  width: 20px;
+  height: 20px;
+  cursor: pointer;
+}
```
owennw: I'm not sure about this cursor. Every other application uses the standard default. However, without a cursor change I think there should be some :hover change, as in other applications. (The same holds for both icons).

owennw:
Just the one point about the cursor. If we can get a quick answer I'd like to see it done. Otherwise I guess we can lump it in with all the other cursor questions.. ?

src/assets/styles/toolbar.less
bjedrzejewski: We can change that later, for now there is no design for what should happen :(

bjedrzejewski: Actullay, we should have an issue- design adn implement cursors across the whole app

bjedrzejewski: Cursor / hover behaviour

bjedrzejewski:
@owennw I will push at the end of the day to gh-paes to rui can check it out

Issue #202, opened 2016-01-12
Can't return to favourites tab by clicking on/around the star


owennw:

owennw:
Could be due to the toolbar over the top.

PR #203, opened 2016-01-12
toolbar fix

bjedrzejewski:
Fix for the issue #202

### DIFF
--- a/src/assets/styles/toolbar.less
+++ b/src/assets/styles/toolbar.less
@@ -1,7 +1,8 @@
 ﻿#toolbar {
   height: 50px;
   right: 65px;
-  position: relative;
+  left: 350px;
+  position: absolute;
 }
 
 #toolbar .drag {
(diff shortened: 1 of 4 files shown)


owennw:
I think this breaks the minimise and close buttons?

bjedrzejewski:
Yes it does, I am bit blind, let me fix it properly

src/assets/styles/toolbar.less
```diff
@@ -1,7 +1,8 @@
 ﻿#toolbar {
   height: 50px;
   right: 65px;
-  position: relative;
+  left: 350px;
+  position: absolute;
```
owennw: We have `@sidebar-width` for that variable.

src/assets/styles/toolbar.less
```diff
@@ -1,7 +1,8 @@
 ﻿#toolbar {
   height: 50px;
   right: 65px;
```
owennw: Similarly there's a hard-coded height for `.sidetab-top` in sidebar.less that should be made a variable and reused here.

bjedrzejewski:
@owennw while noticing it, i fixed the splitter height as well, as Rui said, he needs it at 1px

PR #212, opened 2016-01-13
Fix font in favorites, simplify less.

bjedrzejewski:


src/sidebars/favourites/favourites.less
```diff
@@ -54,11 +54,7 @@
 
   .details {
     float: right;
-
-    * {
-      width: @details-width;
-      text-align: right;
-    }
+    text-align: right;
```
owennw: That's the last use of that variable, so it can be removed.

owennw:
Good spot!

bjedrzejewski:
@owennw when you ready.

PR #214, opened 2016-01-13
Friendly message

owennw:
This provides a basic placeholder until we have a design.

### DIFF
--- a/src/services/quandl-service.js
+++ b/src/services/quandl-service.js
@@ -55,11 +55,15 @@
                 });
             }
 
-            function search(query, cb) {
+            function search(query, cb, noResultsCb) {
                 stockSearch().get({ query: query }, function(result) {
                     result.datasets.map(function(dataset) {
                         processDataset(dataset, query, cb);
-                    })
+                    });
+
+                    if (result.datasets.length === 0) {
+                        noResultsCb();
+                    }
                 });
             }
 
--- a/src/sidebars/search/search-controller.js
+++ b/src/sidebars/search/search-controller.js
@@ -6,10 +6,12 @@
             function($scope, quandlService, storeService) {
                 var self = this;
                 self.query = '';
+                self.noResults = false;
                 self.stocks = [];
 
                 function submit() {
                     self.stocks = [];
+                    self.noResults = false;
                     var favourites = storeService.get();
                     if (self.query) {
                         var length = favourites.length;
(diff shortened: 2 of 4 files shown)


bjedrzejewski:
I will test now and merge if all is good.

bjedrzejewski:
@owennw waiting for updated design, then we merge after testing.

owennw:
This is ready now.

src/services/quandl-service.js
```diff
@@ -55,17 +41,37 @@
...
+                        method: 'GET',
+                        cache: true,
+                        transformResponse: function(data, headers) {
+                            var json = angular.fromJson(data);
+                            return filterByDate(json);
+                        }
+                    }
+                };
+
+                httpConfig.get.timeout = deferred.promise;
```
bjedrzejewski: Why is that and what is that?

bjedrzejewski: Can we write at least a short comment here? Is this related to friendly message?

bjedrzejewski:
@owennw message still visible on hiding search

src/sidebars/search/search-controller.js
```diff
@@ -39,6 +37,13 @@
                                 self.stocks.push(stock);
                             }
                         });
+
+                        var resultsCheck = $interval(function() {
+                            if (lastSearch.$resolved) {
+                                self.noResults = self.stocks.length === 0;
+                                $interval.cancel(resultsCheck);
+                            }
+                        }, 50);
```
bjedrzejewski: What is this logic doing?

bjedrzejewski:
@owennw Cancel unwanted search should be a separate PR, it looks quite important and not necessary like the best thing to do.

bjedrzejewski:
@owennw I looked on stack http://stackoverflow.com/questions/22915513/cancelling-http-request-angularjs and I found this comment which sums up what I think:

I think your approach is incorrect, you shouldn't be spamming your server and cancelling requests whenever the input changes. Instead, do a timeout and send a request when the user stops typing (e.g. whenever the input changes, set a 200ms timeout to contact the server, whenever the input changes clear the timeout and create a new one) – this is how it's usually done

owennw:
That's just, like, your opinion, man.

But really -- the 200ms timeout is a tad arbitrary, and still results in http requests being created, and the results being discarded.

In any case I've removed it from this PR and used a second callback instead.

bjedrzejewski:
It's not just an opinion. If you want to do it properly, you don't spam the server and cancel gets. That's just bad design.

PR #218, opened 2016-01-13
Refresh search favourites

owennw:


src/sidebars/search/search-controller.js
```diff
@@ -62,8 +62,12 @@
...
                 $scope.$on('updateFavourites', function(event, data) {
                     var index = self.stocks.map(function(stock) { return stock.code; }).indexOf(data.code);
-                    if (index > -1 && !self.query) {
-                        self.stocks.splice(index, 1);
+                    if (index > -1) {
+                        if (!self.query) {
+                            self.stocks.splice(index, 1);
+                        } else {
+                            self.stocks[index].favourite = false;
+                        }
```
bjedrzejewski: Can we have some code comment here? What this block does? I know there was no comment before, but it is becoming bit complicated.

owennw: Sure.

owennw:
Comments added.

bjedrzejewski:
@owennw bug- you cant add favourites

bjedrzejewski:
Double clicking on favorite makes it add twice

owennw:
Can you try retesting now? Make sure your local resource is in a valid state before starting though.

bjedrzejewski:
If you search for a stock, add to fav, select it from search bar, unfav from favbar- it is still selected in search. I know. lol

owennw:
Don't believe ;)

bjedrzejewski:
@owennw yeah it seems that you are right. I will test a bit more and merge. I hope its bullet proof now

PR #221, opened 2016-01-13
The software always starts selected.

bjedrzejewski:
This is '==' on purpose here.


owennw:
What does the `==` do that the `===` wouldn't?

bjedrzejewski:
I read up a bit on that and to make it future proof I think this will be the best: `value == null || value === ""`

PR #223, opened 2016-01-13
Fix for bug with auto opening.

bjedrzejewski:
Yeah... it actually makes a difference.


owennw:
Sorry didn't notice that.. !

PR #224, opened 2016-01-13
Update sidebar styling

owennw:
This is so the new rendered styles can be pushed to gh-pages for Rui to play with.


bjedrzejewski:
Great Job!

PR #228, opened 2016-01-14
Fix no results message

owennw:
The way I see this possibly failing now is if retrieving a set of results is quicker than retrieving no results from a previous search.

While possible I don't think it's likely. Perhaps we should investigate a slight delay in searching after typing to make this even less likely?


owennw:
The diff looks so noisy because I think some bad line endings crept in before and this is fixing them. Can you check that though please?

PR #229, opened 2016-01-14
ESLint enabled

bjedrzejewski:
Don't worry about the branch name and don't delete please.


bjedrzejewski:
search-controller.js is noisy because I think you committed some different line endings?

PR #231, opened 2016-01-14
Caption moving when scroll appears.

bjedrzejewski:


owennw:
Just the one comment.

bjedrzejewski:
It will make the code longer if we make it into a var :D

owennw:
HDD space is cheap, DRY code is more maintainable ;)

PR #233, opened 2016-01-14
No CDN usage- everything into node_modules.

bjedrzejewski:
Custom scroll bar is not available as a node_module sadly.


.eslintignore
```diff
@@ -0,0 +1 @@
+src/assets/js/jquery.mCustomScrollbar.concat.min.js
```
owennw: revert pls

gruntfile.js
```diff
@@ -122,6 +122,23 @@ module.exports = function(grunt) {
         },
 
         copy: {
+            modulescss: {
+                expand: true,
+                cwd: 'node_modules/',
+                src: ['d3fc/dist/d3fc.min.css', 'malihu-custom-scrollbar-plugin/jquery.mCustomScrollbar.css'],
+                dest: 'public/assets/css/',
+                flatten: true
```
owennw: Would we rather import the CSS via the LESS file? (Maybe we should ask someone who knows more?)

gruntfile.js
```diff
@@ -122,6 +122,23 @@ module.exports = function(grunt) {
...
+            },
+            modulesjs: {
+                expand: true,
+                cwd: 'node_modules/',
+                src: ['jquery/dist/jquery.min.js', 'angular/angular.min.js', 'angular-resource/angular-resource.min.js',
+                    'moment/min/moment.min.js', 'angular-animate/angular-animate.min.js',
+                    'angular-storage/dist/angular-storage.min.js', 'd3fc/dist/d3fc.bundle.min.js',
+                    'bootstrap/dist/js/bootstrap.min.js', 'malihu-custom-scrollbar-plugin/jquery.mCustomScrollbar.js'],
+                dest: 'public/assets/js/',
+                flatten: true
```
owennw: Can we have one per line?

src/assets/styles/custom-scroll.less
```diff
@@ -13,7 +13,7 @@
 }
 
 .sideScroll:hover .mCSB_dragger_bar {
-  opacity: 1;
+  opacity: 0.75;
 }
```
owennw: Could you put this in a separate PR? It's not really to do with CDNs.

PR #238, opened 2016-01-14
Refactor current selection into service

owennw:
Much better!


src/index.html
```diff
@@ -1,11 +1,11 @@
-﻿﻿﻿<!DOCTYPE HTML>
+﻿﻿<!DOCTYPE HTML>
```
bjedrzejewski: Are you inserting the white space again here?

src/index.html
```diff
@@ -1,11 +1,11 @@
...
 <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>OpenFin D3FC</title>
-    <link href="assets/css/style.css" rel="stylesheet" >
+    <link href="assets/css/style.css" rel="stylesheet">
     <link href="assets/css/d3fc.min.css" rel="stylesheet" />
     <link href="assets/css/jquery.mCustomScrollbar.css" rel="stylesheet" />
     <style>
```
bjedrzejewski: Since you are changing the way this <link> closes, do you want to remove other self closing links? This is the html5 practice.

bjedrzejewski:
Great change, now the code is easier to understand.

bjedrzejewski:
@owennw Maybe favourites can be refactored in similar way rather than using the events mechanic? Not necessarily now, but when we have more time?

src/index.html
owennw: I have no idea -- using my diff tool I don't see a difference.. and if the illegal character were there it wouldn't pass linting..

bjedrzejewski: HTML doesn't get linted. Anyway, I was just curious if you can see something suspicious.

PR #240, opened 2016-01-14
Replace favourite icons

owennw:
Depends on the changes for the selection service.


src/sidebars/star/star.html
```diff
@@ -1 +1 @@
-﻿<div class="glyphicon glyphicon-star {{starClasses(stock)}}" ng-click="starClick(stock); $event.stopPropagation();"></div>
\ No newline at end of file
+﻿<img class="star" ng-src="../../assets/svg/{{favouriteUrl(stock)}}.svg" ng-click="starClick(stock); $event.stopPropagation();" />
```
bjedrzejewski: I think this relative path wont work on gh-pages. Needs to be ng-src="assets/svg/{{favouriteUrl(stock)}}.svg"

bjedrzejewski:
Small comment rest is good I think

PR #242, opened 2016-01-14
Styling

owennw:
Adjusts the splitters, search text and toolbar icons.

### DIFF
--- a/src/sidebars/favourites/favourites.less
+++ b/src/sidebars/favourites/favourites.less
@@ -4,7 +4,7 @@
 @favourite-padding-right: 4px; // not sure why this isn't 5
 @favourite-padding-bottom: 10px;
 @top-height: 30px;
-@favourite-splitter-style: 1px solid @dark-background;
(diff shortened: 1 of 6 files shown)


src/sidebars/favourites/favourites.less
```diff
@@ -4,6 +4,7 @@
 @favourite-padding-right: 4px; // not sure why this isn't 5
 @favourite-padding-bottom: 10px;
 @top-height: 30px;
+@favourite-splitter-style: 1px solid @dark-background;
```
bjedrzejewski: This is the required color: #323A48

src/sidebars/search/search.less
```diff
@@ -1,4 +1,5 @@
 ﻿@search-result-height: 60px;
+@search-splitter-style: 1px solid #323A48;
```
bjedrzejewski: And this one here: #5B626B

bjedrzejewski:
Your splitter style does not include the 0.5 opacity adjustment.

PR #247, opened 2016-01-15
Remove bootstrap

owennw:
.. yup!


bjedrzejewski:
You need to remove it from grunt as well.

owennw:
Good spot! Updated now.

Issue #239, opened 2016-01-14
The top splitte ron the search tab has disappeared. And there shouldn't be one on the favourites tab.


owennw:

owennw:
Closed by #251

PR #249, opened 2016-01-15
Tidy LESS files

owennw:
Tidy for the less files.

Keep this branch open, I'll use it for the missing separator on the search tab #239

### DIFF
--- a/src/sidebars/favourites/favourites.less
+++ b/src/sidebars/favourites/favourites.less
@@ -12,88 +12,87 @@
   background-color: @middle-background;
   transition: @transition;
   cursor: pointer;
+
+  .favourite {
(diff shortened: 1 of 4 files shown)


src/sidebars/favourites/favourites.less
```diff
@@ -12,88 +12,87 @@
...
+    padding-right: @favourite-padding-right;
+    padding-left: @favourite-padding-left;
+    cursor: pointer;
+    height: @favourite-height;
+    border-bottom: @favourite-splitter-style;
+
+    .top {
+      padding-left: 10px;
+      transition: @transition;
+    }
```
bjedrzejewski: Why is all of this (and many more) needlessly nested inside .favourites?

Issue #193, opened 2016-01-12
Showcase now doesn't use quandl API key


owennw:

bjedrzejewski:
This was fixed on my branch, can be a quick fix in James on.

owennw:
This:

> `+        historicFeed.apiKey('kM9Z9aEULVDD7svZ4A8B');`?

owennw:
Can you take a look at this? You seem to have an idea of the best way to approach it.

Also we should look at which fork/version of the showcase we're using. Has James' personal branch been merged into main? Should your branch be updated?

bjedrzejewski:
Yeah, it will require me forking James branch, which is fine.

bjedrzejewski:
I will do it later today.

owennw:
Fixed by #256

PR #256, opened 2016-01-15
Update showcase integration

jleft:
This updates integration of d3fc-showcase as a module, now that it has landed in the main repo: https://github.com/ScottLogic/d3fc-showcase/pull/537

Changes in d3fc-showcase:
- Bug fixes (see [the commit history](https://github.com/ScottLogic/d3fc-showcase/commits/2a659f6ae092c8cff85cfd956d271eb4f4e894c1) for more info)
- No longer fetch Coinbase products by default
- Navigator extent gradient! :rainbow: (May need to override some of our Less variables to get the correct colours)
- Fixes the indicator close label positioning (thanks for raising the issue :smiley:)
- Added a method to the showcase API to set the Quandl API key (#193)

Proposed changes to OpenFinD3FC (in this PR):
- Rather than using my branch, the dependency is linked to commit in the showcase's `develop` branch. When the next release lands, I'd suggest that this dependency is updated to the release.
- Updated d3fc from 5.1.0 -> [5.2.0](https://github.com/ScottLogic/d3fc/releases/tag/v5.2.0) (the showcase needs this version)
- Showcase depends on Bootstrap's dropdown JS, so I've added this back in (it's used for the series and indicator dropdowns)
- Set Quandl API key for the showcase through its API

Let me know what you think.

If you think anything is missing or needs to be changed, just let me know and I'll see what we can do.

### DIFF
--- a/src/showcase/showcase-directive.js
+++ b/src/showcase/showcase-directive.js
@@ -10,7 +10,7 @@
                     selection: '&'
                 },
                 link: function(scope, element) {
-                    var chart = sc.app(),
+                    var chart = sc.app().quandlApiKey('kM9Z9aEULVDD7svZ4A8B'),
                         firstRun = true;
 
                     scope.$watch('selection()', function(newSelection, previousSelection) {
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -141,6 +141,7 @@ module.exports = function(grunt) {
                     'angular-animate/angular-animate.min.js',
                     'angular-storage/dist/angular-storage.min.js',
                     'd3fc/dist/d3fc.bundle.min.js',
+                    'd3fc-showcase/node_modules/bootstrap/js/dropdown.js',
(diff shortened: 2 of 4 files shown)


src/showcase/showcase-directive.js
```diff
@@ -10,7 +10,7 @@
                     selection: '&'
                 },
                 link: function(scope, element) {
-                    var chart = sc.app(),
+                    var chart = sc.app().quandlApiKey('kM9Z9aEULVDD7svZ4A8B'),
```
jleft: I copied the key from the Quandl service just to keep the changeset small - I expect you may want to update this to get the key from the service (or some other way, which doesn't involve duplicating the string)

owennw: Yup, we'll do that.

owennw:
Cheers! This makes our lives quite a bit easier.

PR #259, opened 2016-01-19
Reorder cards

owennw:
Re-ordering the favourite cards using the OpenFin API and tearout functionality. Note: This is purely re-ordering -- no torn out window will persist.

I'm not entirely sure how robust this is -- I think it'll need some exercise.

Also there is a fair amount of new code here -- I don't know into how much detail you'll want to go.

### DIFF
--- a/src/services/store-service.js
+++ b/src/services/store-service.js
@@ -40,6 +40,24 @@
                 return order(favouriteStocks);
             }
 
+            // Move given item in an array to directly after the to-item
+            function reorder(fromItem, toItem) {
+                if (fromItem === toItem) {
+                    return;
+                }
+
+                var oldArray = order(favouriteStocks);
+                var fromIndex = oldArray.indexOf(fromItem);
+                var toIndex = oldArray.indexOf(toItem);
+                oldArray.splice(toIndex, 0, oldArray.splice(fromIndex, 1)[0]);
+
+                for (var i = 0, max = oldArray.length; i < max; i++) {
+                    favouriteStocks[oldArray[i]] = i;
+                }
+
+                save();
+            }
+
             function add(stock) {
                 var stockName = stock.code;
                 if (!favouriteStocks[stockName]) {
@@ -67,6 +85,7 @@
 
             return {
                 get: get,
+                reorder: reorder,
                 add: add,
                 remove: remove
             };
--- a/src/services/store-service.js
+++ b/src/services/store-service.js
@@ -2,7 +2,7 @@
     'use strict';
 
     angular.module('openfin.store', ['angular-storage'])
-        .factory('storeService', ['store', function(store) {
+        .factory('storeService', ['store', '$rootScope', function(store, $rootScope) {
             var initialStocks = {
                 'AAPL': 0,
                 'MSFT': 1,
@@ -31,8 +31,9 @@
                 });
             }
 
-            function save() {
+            function save(stock) {
                 store.set('stocks', favouriteStocks);
+                $rootScope.$broadcast('updateFavourites', stock);
             }
 
             function get() {
@@ -43,7 +44,7 @@
                 var stockName = stock.code;
                 if (!favouriteStocks[stockName]) {
                     favouriteStocks[stockName] = Object.keys(favouriteStocks).length;
-                    save();
+                    save(stock);
                 }
             }
 
@@ -60,7 +61,7 @@
                             favouriteStocks[keys[i]]--;
                         }
                     }
-                    save();
+                    save(stock);
                 }
             }
 
--- a/src/sidebars/favourites/geometry-service.js
+++ b/src/sidebars/favourites/geometry-service.js
@@ -1,0 +1,63 @@
+(function() {
+    'use strict';
+
+    angular.module('openfin.geometry', [])
+        .factory('geometryService', [function() {
+
+            var Point = function(aX, aY) {
+                this.x = aX || 0;
+                this.y = aY || 0;
+            };
+            Point.constructor = Point;
+
+            var Rectangle = function(rect) {
+                this.origin = new Point(rect.left, rect.top);
+                this.extent = new Point(rect.width, rect.height);
+            };
+            Rectangle.constructor = Rectangle;
+
+            Rectangle.prototype = {
+                top: function() {
+                    return this.origin.y;
+                },
+                left: function() {
+                    return this.origin.x;
+                },
+                bottom: function() {
+                    return this.top() + this.extent.y;
+                },
+                right: function() {
+                    return this.left() + this.extent.x;
+                },
+                corner: function() {
+                    return new Point(this.right(), this.bottom());
+                },
+                intersects: function(otherRectangle) {
+                    //return true if we overlap, false otherwise
+
+                    var otherOrigin = otherRectangle.origin,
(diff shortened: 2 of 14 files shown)


owennw:
Also the styling inside the torn-out window isn't correct. I think the relative URLs for the assets aren't correct, or the current directory of the tearout window is wrong?

Or not enough is being injected? Not sure for now.

src/services/store-service.js
```diff
@@ -31,19 +31,38 @@
...
+                var oldArray = order(favouriteStocks);
+                var fromIndex = oldArray.indexOf(fromItem);
+                var toIndex = oldArray.indexOf(toItem);
+                oldArray.splice(toIndex, 0, oldArray.splice(fromIndex, 1)[0]);
+
+                for (var i = 0, max = oldArray.length; i < max; i++) {
+                    favouriteStocks[oldArray[i]] = i;
+                }
+
+                save();
```
bjedrzejewski: You always put stock as a parameter, why not here? Why save then?

src/sidebars/favourites/geometry-service.js
```diff
@@ -0,0 +1,63 @@
...
+            };
+            Point.constructor = Point;
+
+            var Rectangle = function(rect) {
+                this.origin = new Point(rect.left, rect.top);
+                this.extent = new Point(rect.width, rect.height);
+            };
+            Rectangle.constructor = Rectangle;
+
+            Rectangle.prototype = {
```
bjedrzejewski: We may want Chris or Colin to review / have a look? Unless you are 100% sure that this is legit.

src/services/store-service.js
owennw: I agree this looks odd (perhaps it's an anti-pattern). The `stock` input to the `save` function is the data that gets bound to the event emitted. This event is caught in the search controller and the bound stock data is used to find the correct search result and update its icon / remove it from the list.

In this case the stock's favourite status has not changed, only its order. Also, given at least two stocks' orders have changed it's not clear which stock should be passed in the event.

However, passing any stock (the one that was just moved feels most sensible) would probably be safe as the subscribers to the event wouldn't care.

src/sidebars/favourites/geometry-service.js
owennw: This came mostly from the OpenFin tutorial on tearouts, but I guess your point stands as that's just passing it up the chain. Don't mind really, as long as it stays working (which I'm pretty sure it is now).

src/sidebars/favourites/hover-service.js
```diff
@@ -0,0 +1,26 @@
...
+            var self = this;
+            var hoverTargets = [];
+
+            function get() {
+                return hoverTargets;
+            }
+
+            function add(hoverArea, dropTarget, stockCode) {
+                var newHoverTarget = { hoverArea: hoverArea, dropTarget: dropTarget, code: stockCode };
+                if (hoverTargets.indexOf(newHoverTarget) === -1) {
```
bjedrzejewski: JavaScript and it's `contains` method ;)

src/sidebars/favourites/tearout.html
```diff
@@ -0,0 +1,20 @@
...
+            pointer-events: all;
+        }
+
+        input {
+            -webkit-app-region: no-drag;
+            pointer-events: all;
+        }
+    </style>
+</head>
+</html>
```
bjedrzejewski: Can you fix it to look good?

src/sidebars/favourites/tearout.js
```diff
@@ -0,0 +1,218 @@
...
+    'use strict';
+
+    angular.module('openfin.tearout', ['openfin.geometry', 'openfin.hover', 'openfin.store'])
+        .directive('tearable', ['geometryService', 'hoverService', 'storeService', function(geometryService, hoverService, storeService) {
+            return {
+                restrict: 'C',
+                link: function(scope, element, attrs) {
+                    var tileWidth = 230,
+                        tileHeight = 100;
+
```
bjedrzejewski: Do we have to store this in Javascript?

bjedrzejewski: I guess yes...

src/sidebars/favourites/tearout.js
```diff
@@ -0,0 +1,218 @@
...
+                    var dragElement = element[0];
+                    var tearElement = dragElement.parentNode.parentNode;
+
+                    fin.desktop.main(function() {
+                        function initialiseTearout() {
+                            var myDropTarget = tearElement.parentNode,
+                                offset = { x: 0, y: 0 },
+                                currentlyDragging = false;
+
+                            var me = {};
```
bjedrzejewski: What is this?

src/sidebars/favourites/hover-service.js
```diff
@@ -0,0 +1,26 @@
...
+                    // This target is new
+                    hoverTargets.push(newHoverTarget);
+                }
+            }
+
+            return {
+                get: get,
+                add: add
+            };
+        }]);
```
owennw: I have a feeling that as favourites are removed they'll mistakenly persist in the service's local array. They should be removed.. somehow.

As far as the user is concerned I don't think this will have an implication as they extra targets will be ignored, but still it's messy.

src/sidebars/favourites/hover-service.js
owennw: Oh cool, I'd forgotten about that.

bjedrzejewski: It does not exist, I was just joking. Sorry :)

src/sidebars/favourites/tearout.html
owennw: What's wrong with it?

src/sidebars/favourites/tearout.js
owennw: There might be a way to get this information from the element by traversing the DOM.. not sure though I'll have a look.

src/services/store-service.js
owennw: Your thoughts?

src/sidebars/favourites/tearout.html
bjedrzejewski: I meant also to make the visuals work

src/sidebars/favourites/geometry-service.js
bjedrzejewski: Ok if its from there, then it is fine I think

src/sidebars/favourites/tearout.js
bjedrzejewski: I meant the `me` thing, but I take its from the openfin tutorial as well

owennw: Yes, but in general it's a way to structure the code to have a more fluent API to do things like `me.method().method()...` in the mouse handling methods below.

owennw:
@bjedrzejewski I've added two new commits to address the comments and improve the styling of the torn out window.

There's still one small bug I know of: when a card is first selected to be torn out the sparkline does not have the correct background colour, but it will for all subsequent tearing outs.

I am yet to divine why this is... the HTML and CSS are identical between tearing out the first and subsequent times, so maybe something is getting stuck in a pipe somewhere...

Issue #283, opened 2016-01-27
Add hover states for BitFlux buttons


owennw:

owennw:
Closed by #289

Issue #282, opened 2016-01-27
Add hover states for search/favourites bar


owennw:

owennw:
Closed by #289

PR #289, opened 2016-01-27
Update pointers

owennw:
Updates the cursors as per the design.


owennw:
Added #283

owennw:
Added #282

akhmorgan:
Looks good to me.

PR #291, opened 2016-01-27
Dropdown style

owennw:
Does most of the styling as specified. Doesn't merge the dropdown window with the icon above -- will have to investigate to see how much work that is.

### DIFF
--- a/src/showcase/showcase-changes.less
+++ b/src/showcase/showcase-changes.less
@@ -23,12 +23,12 @@
 }
 
 #selectors {
-  button:hover {
-    background: #214244;
-  }
-
   button {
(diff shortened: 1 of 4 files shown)


akhmorgan:
The pedant in me would like to see hex colour values in their own variables, but seeing as they're only used once at present it's OK. Other than that; looks good to me.

owennw:
Yes I was wondering whether it was bloat to give them names or not.

akhmorgan:
Giving names is a little bit bloating when they're only used once, but in my opinion it improves readability and future-proofs against typos etc. should the colours be used again later. Happy to merge as is for now though.

PR #294, opened 2016-01-29
Select search results with up/down arrow keys

### LINKED ISSUE #161: Add support for arrow keys for searching

WPFerg:
Resolves #161

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -33,7 +33,7 @@
                 <div class="sidetab-top">
                     <img src="assets/svg/search_tab.svg" class="top-icon" />
                     <input class="sidetab" type="text" maxlength="20" placeholder="Enter stock name or symbol"
-                           ng-show="sidebarCtrl.showSearches()" ng-model="searchCtrl.query" />
+                           ng-show="sidebarCtrl.showSearches()" ng-model="searchCtrl.query" ng-keyup="searchCtrl.onSearchKeyUp($event)" />
                 </div>
                 <div id="searchScroll" class="sideScroll">
                     <div class="sidetab" ng-show="sidebarCtrl.showSearches()">
--- a/src/index.html
+++ b/src/index.html
@@ -33,7 +33,7 @@
                 <div class="sidetab-top">
                     <img src="assets/svg/search_tab.svg" class="top-icon" />
                     <input class="sidetab" type="text" maxlength="20" placeholder="Enter stock name or symbol"
-                           ng-show="sidebarCtrl.showSearches()" ng-model="searchCtrl.query" ng-keyup="searchCtrl.onSearchKeyUp($event)" />
+                           ng-show="sidebarCtrl.showSearches()" ng-model="searchCtrl.query" ng-keydown="searchCtrl.onSearchKeyDown($event)" />
                 </div>
                 <div id="searchScroll" class="sideScroll">
                     <div class="sidetab" ng-show="sidebarCtrl.showSearches()">
--- a/src/sidebars/search/search-controller.js
+++ b/src/sidebars/search/search-controller.js
@@ -8,6 +8,7 @@
                 self.query = '';
                 self.noResults = false;
                 self.stocks = [];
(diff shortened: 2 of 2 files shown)


src/sidebars/search/search-controller.js
```diff
@@ -17,8 +18,35 @@
...
+                function changePointer(delta) {
+                    var newPointer = self.stockSelectionPointer + delta;
+
+                    self.stockSelectionPointer = Math.max(
+                        0,
+                        Math.min(
+                            newPointer,
+                            self.stocks.length - 1
+                        )
+                    );
```
owennw: Does this mean that you can't arrow up from the top search result back into the search box? Similarly can you key down from the search box into the results? (Or do you think that is out of the scope of this?)

WPFerg: It's limiting it so you can't try and set the pointer to something that doesn't exist . The search box retains focus throughout (similar to how most autocompletes work), all the arrow keys do is change the selected stock

owennw: Nice.

src/index.html
```diff
@@ -33,7 +33,7 @@
                 <div class="sidetab-top">
                     <img src="assets/svg/search_tab.svg" class="top-icon" />
                     <input class="sidetab" type="text" maxlength="20" placeholder="Enter stock name or symbol"
-                           ng-show="sidebarCtrl.showSearches()" ng-model="searchCtrl.query" />
+                           ng-show="sidebarCtrl.showSearches()" ng-model="searchCtrl.query" ng-keyup="searchCtrl.onSearchKeyUp($event)" />
                 </div>
```
owennw: Not sure how I feel about this event being on keyup.

It's a bit weird not having the stock change on key down, and also if you hold down the key the selection only changes by one, and won't move throughout the search results, as it would on a keypress I believe.

src/sidebars/search/search-controller.js
```diff
@@ -56,6 +84,12 @@
                             if (!stockAdded) {
                                 self.stocks.push(stock);
                             }
+
+                            // Change the selection pointer to be the selected stock, if it exists in the list
+                            // (otherwise, set to -1, which is acceptable as there is no selection yet)
+                            self.stockSelectionPointer = self.stocks.map(function(stockItem) {
+                                return stockItem.code;
+                            }).indexOf(self.selection());
                         },
```
owennw: Ensure you have 'AAPL' as a favourite stock, and search for 'a'. Key down to a stock a few down from AAPL (which should be at the top). Switch to the favourites tab and select AAPL. Switch back to the search tab and you'll see AAPL is selected. Give the search bar the focus and press arrow-down. You'll see the selection will jump back down to where it was initially, and not continue from AAPL at the top.

WPFerg: Thanks -- fixed (which resulted in a code tidy!)

src/index.html
WPFerg: Keypress events don't trigger for arrow keys, rather weirdly, just down/up. I chose keyup just because it felt more analogous to a click event, where that's registered on a mousedown + up, but I don't mind switching it out

owennw: That's really quite frustrating. I'd like to see functionality where a user can press the down key and the selection will initially move one down, and then start to go down the list. Maybe some sort of setInterval? But surely this is a solved problem...

In any case keyDown feels a more natural experience for me..

owennw:
Much better! Just one more thing though...

When using the arrow keys to move the selection above/below the displayed search results the results don't scroll with the selection. If it's simple I'd like to see that in here too. If it's a lot of work though it could be as a separate PR.

WPFerg:
@owennw I was thinking about that, but I'd suggest another PR. You'd need to use something like `scrollIntoView` on the underlying node/do some DOM jiggery-pokery, possibly adding a controller to the search card directive

owennw:
Alright, raise it as an issue?

PR #296, opened 2016-01-29
Directive split

akhmorgan:
Splits index.html down a bit.

### DIFF
--- a/src/sidebars/sidebar.html
+++ b/src/sidebars/sidebar.html
@@ -1,4 +1,4 @@
-<div class="sidebars" ng-controller="SidebarCtrl as sidebarCtrl">
+<div class="sidebars">
     <div class="favourites {{sidebarCtrl.favouritesClass()}}" ng-controller="FavouritesCtrl as favouritesCtrl" ng-click="sidebarCtrl.favouritesClick()">
         <div class="sidetab-top">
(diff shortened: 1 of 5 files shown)


src/sidebars/sidebar.html
```diff
@@ -0,0 +1,32 @@
+<div class="sidebars" ng-controller="SidebarCtrl as sidebarCtrl">
+    <div class="favourites {{sidebarCtrl.favouritesClass()}}" ng-controller="FavouritesCtrl as favouritesCtrl" ng-click="sidebarCtrl.favouritesClick()">
```
owennw: Given the controller is declared in the directive does it need to be (re)declared here?

src/toolbar/toolbar.html
```diff
@@ -0,0 +1,8 @@
+<div id="toolbar" ng-controller="ToolbarCtrl as toolbarCtrl">
+    <img id="logo" src="../assets/svg/scottlogic_logo.svg" />
```
owennw: Same comment as above. I'm also curious about the relative path here too.

owennw:
Couple of small comments, looks good though!

src/sidebars/sidebar.html
akhmorgan: You're quite right. I'll fix that.

PR #300, opened 2016-02-02
Restructure local storage

owennw:


src/services/store-service.js
```diff
@@ -46,41 +40,39 @@
                     return;
                 }
 
-                var oldArray = order(favouriteStocks);
-                var fromIndex = oldArray.indexOf(fromItem);
-                var toIndex = oldArray.indexOf(toItem);
-                oldArray.splice(toIndex, 0, oldArray.splice(fromIndex, 1)[0]);
+                var window = getWindowStore();
```
WPFerg: Fairly nitpicky, but I think it'd be better to rename `window` to something else (`windowStore`?) to avoid ambiguity with the native object.

src/services/store-service.js
```diff
@@ -1,43 +1,37 @@
...
 
-                return tuples.map(function(x) {
-                    return x[0];
-                });
+                return windowIndex > -1 ? storage[windowIndex] : undefined; // TODO: undefined?
             }
 
             function save(stock) {
-                store.set('stocks', favouriteStocks);
+                store.set(KEY_NAME, JSON.stringify(storage));
```
WPFerg: Ideally this would only change the object for the current window, but what happens when another window saves? Would it overwrite the other window's changes, since the `initialStocks` variable is cached on creation, and so wouldn't reflect the updated values?

owennw: That's a very good point. In this current branch that won't be a problem has new windows can't be created yet, but in the (not too far.. !) future when they are I think I'll tackle this problem by promoting the store-service to be created initially by the overall parent, and shared with all the other windows. That will mean there is only one instance of it. I'll add some sort of comment to that effect.

src/services/store-service.js
owennw: Will do.

owennw:
I haven't tackled the overwriting of the store yet, as in this branch that isn't possible.

PR #301, opened 2016-02-03
Tearout windows

owennw:
It's landing.

What this PR includes:

- The entire application has been restructured to have a hidden parent window delegating services and events to all the child windows. This approach is rather rough and I would like to see it refactored and improved in the future.
- The local storage has been re-written to cope with the concept of multiple windows. Included is the concept of whether the window is closed or open, but this is just a placeholder, it is not being used at the moment.
- Favourite cards can now be torn out to instantiate a new app where they were dragged. There seems to be a small intermittent issue with a duplicate favourite card appearing at the moment -- needs more testing.
- If in the old window the torn out favourite was also selected, the selection in the showcase will be outdated: this should be refreshed.
- No attempt is made to restore the last closed window on next startup, and at the moment the number of kept windows in local storage is near-limitless.

Sorry it's so big.


src/sidebars/favourites/tearout.js
```diff
@@ -11,37 +11,53 @@
                         var dragElement = element[0],
                             tearElement = dragElement.parentNode.parentNode,
                             tileWidth = tearElement.clientWidth,
-                            tileHeight = tearElement.clientHeight;
+                            tileHeight = tearElement.clientHeight,
+                            store;
 
-                        function createConfig() {
+                        function createConfig(minWidth, minHeight, width, height, url, resizable, maximizable, showTaskbarIcon, saveWindowState) {
```
WPFerg: Rather than pass the arguments in as one long function declaration, could you pass in a JSON object with those fields set, and add in `autoShow`/`frame` (or defaults for any undefined key)? That way, order isn't important and would make it easier to drop in/out keys.

owennw: Yeah that seems okay. A JSON object is a weaker type but with sensible defaults it should be okay.

src/parent-controller.js
```diff
@@ -0,0 +1,32 @@
...
+                'url': 'index.html'
+            };
+
+            windowCreationService.ready(function() {
+                // TODO: Restore correct window(s)
+                windowCreationService.createWindow(config, function(newWindow) {
+                    newWindow.show();
+                });
+
+                $scope.$on('updateFavourites', function(event, data) {
```
WPFerg: This and the accompanying `broadcast` seem to be the only time you're leveraging angular on the parent window. More of a debate, but do you think it's needed for it to be wrapped up in angular services, just for this case? It *might* make it seem a little simpler to not wrap up in angular services expose some bindings on the child via one variable (like `fin`), which gets wrapped up in the desktop/childWindow service.

owennw: Given this window is never shown, using angular doesn't make a lot of sense I think. Perhaps ditching it here and using bindings is the way to go, yes.

src/sidebars/favourites/favourite-controller.js
```diff
@@ -87,7 +92,7 @@
 
                 self.update();
 
-                $scope.$on('updateFavourites', function(event, data) {
+                window.addEventListener('updateFavourites', function(event) {
```
WPFerg: To minimise window accesses/the diff, could you add the `updateFavourites` event listener once, and on the callback fire a `$rootScope.$broadcast` in the store service?

src/sidebars/favourites/tearout.js
owennw: @WPFerg The thing is though it's easy to leave variables that have no sensible default out of this object. For example `minWidth`, which will be very difference for a tearout card and a main window instance. Is it better to allow the developer to (accidentally) leave this out?

WPFerg: If there's no sensible default, you wouldn't need to define one, since Openfin will apply its defaults. I'm fine with it being either way, just a long list of arguments to a function seems arbitrary without having to check the signature each time for the order.

owennw: I'm inclined to leave it as it is, given there are only 2 static properties..

src/parent-controller.js
owennw: Angular is also being used to create the services (as singletons) and inject them into the application.. is it easy to recreate that?

WPFerg: I've seen `window.storeService` etc used in places, would it not be something like `parentBindings.store.save()`, for example? It'll still be a singleton as the hidden parent will be the only one creating it, and all children would have the same bindings. Injection logic should still be the same, using the `super hack`

src/sidebars/favourites/tearout.js
WPFerg: That's fair enough. It's only called in 2 places, if it were called more frequently and in more configurations it might be worth it

akhmorgan: I'm a little late to the party, but here's my 2 cents: I'd say that a json object makes more sense, but considering that this function essentially just spits out what it takes in, but formatted into an object; is this function even necessary at all? This would be an easier decision if we were using typescript where we could enforce the format of the json object a bit better, but still, I'm inclined to say that a json object is better than tons of params. I see your point @owennw about the lack of defaults and accidentally missing things, but as it is you might accidentally miss something whilst putting the parameters in, or you might get them in the wrong order, which could lead to weird behaviour. However since it is only called in two places, it's not that bad I guess

owennw: I've refactored this method slightly based on the comments. When I've finished all the (current and upcoming) comments we can rediscuss.

owennw:
Superceded.

PR #302, opened 2016-02-05
Concat + minify

WPFerg:
Fixes #51

PR #303, opened 2016-02-05
Es6 classes

WPFerg:
*Built on #302*
Big refactor -- using es6 classes with transpilation using Babel. I'd recommend a thorough test first! Some things could have slipped through I've missed


owennw:
Cool, I think I've looked through it all now. What's the situation with the angular directives?

WPFerg:
I've pushed up some amendments. Re: directives, personally I think it's better just to let the controller be a class, since the directive function returns a JSON object. There's ways to get it working but it feels to me that it's trying to force a class in where it's perhaps not the best use for one (similar to a `factory` returning a JSON object, where you'd use a service instead because it's `new`ed)

PR #304, opened 2016-02-09
Tearout backup

owennw:
New shiny PR, replacing #301


WPFerg:
With changes, it seems okay. Having the `$rootScope`, etc passed to the new class isn't great, but it makes it a bit cleaner and easier to see how it all links together.

owennw:
Get in there.

PR #318, opened 2016-02-09
Update selection when selected favourite torn out

owennw:


WPFerg:
Looks fine to me, just one small thing (but that's a matter of preference, imo)

PR #320, opened 2016-02-10
Cache closed window

owennw:
The first commit updates the windowService to a more ES6 compliant style. The second one updates the `closed` variable for each window in the local storage, except for the final window which stays with `closed: false`.

I'm interested particularly in feedback for the `createWindow` and `createTearoutWindow` methods, as I'm not currently too happy with the responsibility and boundaries between the two.

### DIFF
--- a/src/parent-controller.js
+++ b/src/parent-controller.js
@@ -1,33 +1,50 @@
 (function() {
     'use strict';
 
-    const config = {
-        'autoShow': true,
-        'minWidth': 918,
-        'minHeight': 510,
-        'defaultWidth': 1280,
-        'defaultHeight': 720,
-        'frame': false,
-        'url': 'index.html'
-    };
-    class ParentCtrl {
-        constructor($scope, windowCreationService) {
-            windowCreationService.ready(function() {
-                // TODO: Restore correct window(s)
-                windowCreationService.createMainWindow(config);
+    function getConfig() {
+        return {
+            'autoShow': true,
+            'minWidth': 918,
+            'minHeight': 510,
+            'defaultWidth': 1280,
+            'defaultHeight': 720,
+            'frame': false,
+            'url': 'index.html'
+        };
+    }
 
-                $scope.$on('updateFavourites', function(event, data) {
+    class ParentCtrl {
+        constructor($scope, storeService, windowCreationService) {
+            windowCreationService.ready(() => {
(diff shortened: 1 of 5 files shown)


owennw:
Arghhhh line ending fracas with the store-service in the second commit.

WPFerg:
I agree, they do seem quite intertwined. What about having an additional function that retains all the logic, but invokes the `successCb` parameter, allowing the different logics to be applied for the two different functions.
ie. `createMainWindow` calls `createWindow` with the success callback applying the logic in `if (!tearout)`
and `createTearoutWindow` calls `createWindow`?

WPFerg:
I've got a weird error. From the default favourites, I dragged out `TSLA` and `TITN` each into a new window, and closed all of them. Restarting, my local storage is `[{"id":"main","stocks":["AAPL","MSFT","TSLA","SNDK"],"closed":false},{"id":"window573197","stocks":["TITN"],"closed":false}]`

Shouldn't it be something like `[{"id":"main","stocks":["AAPL","MSFT","SNDK"],"closed":false},{"id":"window573197","stocks":["TITN"],"closed":true},{"id":"windowUNIQUEID","stocks":["TSLA"],"closed":true}]`? If so, is this similar to the comment here? https://github.com/owennw/OpenFinD3FC/pull/300#discussion_r51564662

owennw:
Can you delete your cache and try again? But while making the changes keep and eye on your local storage and check it updates correctly? It may be the way you're closing the app -- if you force quit it OF may not get the change to update its cache..

WPFerg:
I've cleared everything, and am still getting `closed: false` on closed windows on the next start up. Even though `closed:false` holds, they aren't being initialised on the next start up.

I think it might have something to do with time between windows closing/race conditions. I closed a window, waited a bit before closing the next, and `true` applied itself. I closed another window then quickly closed another and they are both `false`. Moreso, on the next restart, the `closed: true`s turned to `closed: false`s.

owennw:
I'm not sure what to do about that..

WPFerg:
It's a weird one, it might just be my local storage being odd -- wouldn't be the first time. What happens when you try?

WPFerg:
The new commits seem fine, I'm getting less unexpected behaviours. There still seems to be some race conditions on fast window closing (one straight after another), though. :confused: Are you sure that the comment above doesn't apply? That's the only thing I can think of that would cause it

PR #327, opened 2016-02-11
Compact view

akhmorgan:


src/main/main-controller.js
```diff
@@ -0,0 +1,22 @@
...
+      constructor($timeout, currentWindowService) {
+          this.$timeout = $timeout;
+          this.currentWindowService = currentWindowService;
+          currentWindowService.ready(this.onReady.bind(this));
+      }
+      onReady() {
+          this.window = this.currentWindowService.getCurrentWindow();
+      }
+
+      isSummarised() {
```
owennw: I know we've been using the term "summary view" a lot but I think I'd prefer a name more like `isCompact`, and "compactView".

src/main/main.html
```diff
@@ -0,0 +1,14 @@
+<div class="main" ng-class="{summary: mainCtrl.isSummarised()}">
+    <toolbar></toolbar>
+	<side-bar></side-bar>
+	<div class="right-content">
+		<div class="main-content">
+			<div ng-controller="ShowcaseCtrl as showcaseCtrl">
```
owennw: `ng-show="showcaseCtrl.hasSelection()` seems to have been removed.

src/sidebars/sidebar.html
```diff
@@ -17,8 +17,26 @@
...
+                </div>
+                <div class="no-results" ng-show="searchCtrl.noResults">
+                    Oops!<br>
+                    Looks like no matches were found.
+                </div>
+            </div>
+        </div>
+    </div>
+    <div class="search summary-search {{sidebarCtrl.searchClass()}}" ng-controller="SearchCtrl as searchCtrl" ng-click="sidebarCtrl.searchClick()">
+         <div>
```
owennw: It feels like there's a lot of duplication between the summary-search and the main-search HTML. Can this be put into a directive and reused?

src/sidebars/sidebar.html
```diff
@@ -17,8 +17,26 @@
             </div>
         </div>
     </div>
-    <div class="search {{sidebarCtrl.searchClass()}}" ng-controller="SearchCtrl as searchCtrl" ng-click="sidebarCtrl.searchClick()">
-        <div class="sidetab-top">
```
owennw: Why has `sidetab-top` left?

src/toolbar/toolbar-controller.js
```diff
@@ -6,11 +6,21 @@
             this.$timeout = $timeout;
             this.currentWindowService = currentWindowService;
             this.maximised = false;
+            this.summarised = false;
             currentWindowService.ready(this.onReady.bind(this));
         }
 
         onReady() {
             this.window = this.currentWindowService.getCurrentWindow();
+            this.window.getBounds(function(bounds) {
```
owennw: Would prefer the `=>` syntax here.

src/toolbar/toolbar-controller.js
```diff
@@ -6,11 +6,21 @@
...
+                if (bounds.width === 230) {
+                    this.summarised = true;
+                    this.window.summarised = true;
+                }
+                else {
+                    this.window.summarised = false;
+                }
```
owennw: `this.summarised = this.window.summarised = bounds.width === 230` perhaps?

src/toolbar/toolbar-controller.js
```diff
@@ -6,11 +6,21 @@
...
             this.currentWindowService = currentWindowService;
             this.maximised = false;
+            this.summarised = false;
             currentWindowService.ready(this.onReady.bind(this));
         }
 
         onReady() {
             this.window = this.currentWindowService.getCurrentWindow();
+            this.window.getBounds(function(bounds) {
+                if (bounds.width === 230) {
```
owennw: This `230` feels a bit like a magic number. However, the JSON file I mentioned on Slack might help with that.

src/toolbar/toolbar-controller.js
```diff
@@ -22,6 +32,16 @@
...
             });
+            this.window.addEventListener('bounds-changed', (e) => {
+                this.window.getBounds(function(bounds) {
+                    if (bounds.width === 230) {
+                        this.window.summarised = true;
+                    }
+                    else {
+                        this.window.summarised = false;
+                    }
+                });
```
owennw: Same `=>` and `if else` refactor comment as above.

src/toolbar/toolbar-controller.js
```diff
@@ -34,6 +54,21 @@
...
+            this.summarised = !this.summarised;
+            this.window.summarised = this.summarised;
+            if (this.summarised) {
+                this.window.resizeTo(230, 500, 'top-right');
+            }
+            else if (this.maximised) {
+                this.window.maximize();
+            }
+            else {
+                this.window.resizeTo(1280, 720, 'top-right');
```
owennw: It might be nice to store the window size (and maybe even position?) when the window is changed into compact view. Then when the window is restored the old size could also be restored, instead of using the default?

This is more of a feature though perhaps.

src/toolbar/toolbar.html
```diff
@@ -1,8 +1,9 @@
 <div id="toolbar">
     <img id="logo" src="assets/svg/scottlogic_logo.svg" />
     <toolbar-icon name="minimise" icon-click="toolbarCtrl.minimiseClick()"></toolbar-icon>
-    <toolbar-icon name="maximise" ng-show="!toolbarCtrl.maximised" icon-click="toolbarCtrl.maximiseClick()"></toolbar-icon>
-    <toolbar-icon name="normal_size" ng-show="toolbarCtrl.maximised" icon-click="toolbarCtrl.normalSizeClick()"></toolbar-icon>
+    <toolbar-icon name="summarise" icon-click="toolbarCtrl.summariseClick()"></toolbar-icon>
+    <toolbar-icon name="maximise" ng-show="!toolbarCtrl.maximised && !toolbarCtrl.summarised" icon-click="toolbarCtrl.maximiseClick()"></toolbar-icon>
+    <toolbar-icon name="normal_size" ng-show="toolbarCtrl.maximised && !toolbarCtrl.summarised" icon-click="toolbarCtrl.normalSizeClick()"></toolbar-icon>
```
owennw: <minor comment> This might be too much logic for HTML. Though I'm not overly concerned (it's just an `&&` after all..). If @WPFerg thinks so too might be worth making bespoke properties/functions.

src/toolbar/toolbar.less
```diff
@@ -28,14 +28,32 @@
...
+.summary {
+  #summarise-button {
+    right: 45px;
+  }
+
+  #minimise-button {
+    right: 75px;
+  }
+
+  #logo, #minimise-button, #summarise-button, #close-button {
```
owennw: Is the logo shown in compacy view?

src/sidebars/search/search.less
```diff
@@ -65,3 +65,17 @@
     }
   }
 }
+
+.summary-search {
+  display: none;
+}
+
+.summary{
```
owennw: :nitpick: just some whitespace here.

src/main/main-controller.js
akhmorgan: That's fine by me. I just based the names off the assets that Rui made. If the consensus is to change to compact rather than summary, I'm happy to do that.

src/main/main.html
akhmorgan: Good spot. That's a merge error. I'll fix that.

owennw:
Overall I like the approach -- it certainly seems simpler than what I'd imagine not doing it in CSS would provide.

I'm still a bit hazy on the use of the `window` object to store the `isSummarised` information. It looks to me that both the `mainCtrl` and the `sidebarCtrl` have very similar logic for retrieving this information, which makes sense, but what would be lost by moving the `isSummarised` information into the `currentWindowService`?

Both of the controllers have access to this service, so the information can be kept there instead of on the window.

src/main/main-controller.js
owennw: @WPFerg ?

src/sidebars/sidebar.html
akhmorgan: The styling it was applying wasn't necessary and it didn't really make sense to leave it in considering the search can now appear at the bottom not at the top. So it made more sense to just remove the class than overriding the styles.

src/sidebars/sidebar.html
akhmorgan: Yes. That's definitely the way I'd like to go with this. This implementation was just a quick fix to the animation issue.

src/toolbar/toolbar-controller.js
akhmorgan: I agree. The default window sizes should definitely be extracted into config somewhere.

src/sidebars/sidebar.html
owennw: In that case given it's now specific to the favourites section it could probably do with a rename.

src/toolbar/toolbar-controller.js
akhmorgan: Yeah, I'll change that. This originally did more conditional stuff, but the if/else is unnecessary now.

src/toolbar/toolbar-controller.js
akhmorgan: I agree. I'm planning on looking into that shortly.

src/toolbar/toolbar.html
akhmorgan: Yeah, I can just tale this out into the less and set display:none on these in the compact view.

src/toolbar/toolbar.less
akhmorgan: It is not. I'll remove that.

src/main/main-controller.js
WPFerg: I prefer compact, as that has more connotation with the window being smaller. Summary, for example, could imply the window's the same size, just showing a 'summary' of all favourites

src/toolbar/toolbar-controller.js
WPFerg: I'd be inclined to say it could be another PR, since this is already fairly large

src/toolbar/toolbar.html
WPFerg: I think it's probably fine just with one `&&`, if there's more then it's probably worth doing

src/toolbar/toolbar-controller.js
akhmorgan: @WPFerg absolutely. I'm not planning on adding that onto this PR

owennw:
Comments now I've used it:

1. There doesn't seem to be a way to get from searching back to favourites.
2. The scrollbars have disappeared entirely (probably due to a CSS change).
3. When searching using the compact view the search results bleed over the search area.
4. The searches in the main view and compact view are disconnected (i.e. search for something in compact, restore the app and note that search query isn't there).
5. While the hover and selected CSS states are removed, clicking the stock will still change the selection in the main view.
6. It's possible to get the whole window to have the hover state 'green' colour, by placing the mouse in certain areas.

src/toolbar/toolbar-controller.js
owennw: Agreed.

owennw:
According to the spec when the search bar is clicked there should be a transition where the tab grows up from the bottom, to near the top, similarly to how it grows from the left to the right in the main view.

This would provide a natural way of re-selecting the favourites tab in compact view and prevent any bleed of the search results into the search area (as the search box will be *above* the list of search results)

akhmorgan:
Re: isSummarised - I'm going to remove that logic from sidebar as it is no longer necessary, the logic in main is sufficient. I'm not sure that moving the flag onto the currentWindowService will work. It would work if that service is instantiated once per window, but I'm not sure that it is. Also it seems like something that the window itself should be aware of. It has a flag for maximised etc. so this seems like a logical progression within the context of this app to me.

Re: the rest -  All on the todo list. This PR is just to get the initial foundation in place so that this branch doesn't keep drifting further and further from dev

owennw:
As far as I'm aware each window is a new angular app, so the service will be instantiated per app/window, and will be unique. (This was part of the motivation for creating the parent window, so some services are *only* ever created once.)

akhmorgan:
Well in that case, I can certainly give it a try. I'm still of the opinion that putting it on the window object itself makes more sense than having it on the service though.

WPFerg:
I think it should be on the service, it feels more like it's in the scope of the state of the current window

Found another bug: when resizing (horizontally or diagonally), the compact view snaps to a width greater than that of the compact view, which looks a little odd.

owennw:
I haven't tried it, but the compact view should only have vertical resize anyway.

src/sidebars/sidebar.html
WPFerg: I agree, looking through it's quite a big introduction with a lot of reuse.

akhmorgan:
Yep the re-size is a known issue too. I'll start putting these issues on GH so we can all keep track

src/toolbar/toolbar-controller.js
WPFerg: I'll lump this in here since it's about restoring states. Maybe it's worth adding a `compact: true` or something to the StoreService so it can remember if a window is in compact view between sessions?

owennw: Definitely worth doing that. Definitely a different issue/PR.

PR #328, opened 2016-02-11
Fix first window name

owennw:


WPFerg:
Looks fine, aside from the line endings in `src/window-service.js`. Maybe it's worth adding `.gitattributes` in another PR

PR #330, opened 2016-02-12
Fix first window name

owennw:


WPFerg:
Looks good with those changes.

PR #334, opened 2016-02-12
Fix custom scroll bars not working

WPFerg:
Moves the custom scroller into its own angular directive, so it doesn't check on page load.


WPFerg:
I'm resubmitting this, since CI is being silly.

PR #341, opened 2016-02-15
Window pool

owennw:
Implements a cache of ready OpenFin windows to increase usability and performance for tearing out favourite cards into new windows. Only main application windows are pooled -- windows for tearout cards are still handled as they were before.


akhmorgan:
looks good to me

PR #353, opened 2016-02-15
Default sizes for tearout windows

owennw:


src/sidebars/favourites/tearout.js
```diff
@@ -21,8 +21,8 @@
                             };
 
                             if (tearout) {
-                                config.minWidth = tileWidth;
-                                config.minHeight = tileHeight;
+                                config.maxWidth = tileWidth;
```
bjedrzejewski: Why did you change maxWindths?

owennw: That's the size of the card being torn out, and using the values as a max made more sense and fixed an odd bug too. Perhaps it should be both min and max though.

akhmorgan:
looks good to me

PR #354, opened 2016-02-15
Tear in

owennw:
Refactor the window service so that the tearout windows are no longer tracked, so update events won't be pushed to them. Also moves some code into the geometry service, which is now shared with the parent app.

### DIFF
--- a/src/window-service.js
+++ b/src/window-service.js
@@ -69,8 +69,9 @@
     }
 
     class WindowCreationService {
-        constructor(storeService, $q) {
+        constructor(storeService, geometryService, $q) {
             this.storeService = storeService;
+            this.geometryService = geometryService;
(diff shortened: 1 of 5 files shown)


src/window-service.js
```diff
@@ -46,74 +46,92 @@
...
+            }
+        }
+
+        dispose(_window, closedCb) {
+            var parent = this.openWindows[_window.name];
+            if (parent) {
+                // Close all the OpenFin tearout windows associated with the closing parent.
+                for (var i = 0, max = parent.length; i < max; i++) {
+                    parent[i].close();
+                }
```
WPFerg: Could this be replaced with `parent.forEach((child) => child.close());`?

owennw: Yup!

PR #367, opened 2016-02-16
Prevent tearing in to minimised windows

owennw:


src/window-service.js
```diff
@@ -94,26 +94,34 @@
...
+                        this.otherInstance = mainWindow;
+                        result = true;
+                    }
 
-            return false;
+                    deferred.resolve();
+                });
+            });
+
+            this.$q.all(promises).then(() => cb(result));
```
WPFerg: The native `Promise.race` might be a good thing to try, I don't think it'll make much difference aside from running time.

owennw: I'm a bit wary of using both `$q` and native promises (and then one day using `Q`). I think I'd rather stick to just the one. Also I don't see getting a window state to be a long wait... (famous last words).

WPFerg: That's fair enough, it makes sense for what little it would save.

WPFerg:
Closing this, and using #370.

PR #368, opened 2016-02-16
Added the closed_tabs button - work in progress

bjedrzejewski:
If you want to have the look already there so everyone can comment / suggest / fix


owennw:
Should the button still be visible when the search tab is open?

bjedrzejewski:
@owennw that was going to be my question. What do you think? Maybe I will make it fade?

owennw:
The icon should remain, but clicking it should not reopen the favourites tab. Also FYI there is a design for the icon's placement in compact view (it's moved to the toolbar, left of the minimise icon) but you may not want to do that in this PR / you should check for merge issues first.

owennw:
Having looked at the spec again we (I) may have misinterpreted the two icons. It looks like the blue one is for when a window has recently closed, and grey for when the closed windows have been seen, I think. I don't know if the blue one was also intended as a hover-state or not.

src/app.js
```diff
@@ -15,7 +15,9 @@
         'openfin.star',
         'openfin.tearout',
         'openfin.minichart',
-        'openfin.scroll'
+        'openfin.scroll',
+        'openfin.closedwindows',
+        'openfin.closedcard'
```
owennw: I feel these should be camel-cased.

src/assets/styles/style.less
```diff
@@ -3,6 +3,7 @@
 
 @import "variables.less";
 @import "../../main/main.less";
+@import "../../closedwindows/closedcard.less";
```
owennw: Similar camel-casing here

src/closedwindows/closedcard-controller.js
```diff
@@ -0,0 +1,13 @@
+(function() {
+    'use strict';
+
+    class ClosedcardCtrl {
+        constructor() {
+
+        }
```
owennw: The constructor can be deleted.

src/app.js
bjedrzejewski: minichart is not, I followed the convention

src/closedwindows/closedcard-controller.js
```diff
@@ -0,0 +1,13 @@
...
+    }
+    ClosedcardCtrl.$inject = [];
+
+    angular.module('openfin.closedcard', [])
```
owennw: The injected parameters shouldn't be declared again here (they're declared in the app.js).

src/closedwindows/closedcard-controller.js
bjedrzejewski: I will be adding it right back in the net work.

src/closedwindows/closedcard.html
```diff
@@ -0,0 +1,7 @@
+<div class="closedcard">
+    <div class="closedtime">{{cwindow.closed | date:'yyyy-MM-dd HH:mm:ss'}}
```
owennw: The spec says the date format should be 'dd MMM yyyy' I think.

src/closedwindows/closedcard.html
```diff
@@ -0,0 +1,7 @@
...
+    </div>
+    <span class="closedstock" ng-repeat="stock in cwindow.stocks">
+        {{stock}} {{$last ? '' : ', '}}
```
owennw: I imagine there's an unwanted space between the stock name and the comma here?

src/closedwindows/closedcard-controller.js
bjedrzejewski: I don't understand. I thought this follows the pattern of every other controller.

src/closedwindows/closedcard.less
```diff
@@ -0,0 +1,22 @@
+.closedtime {
+  font-size: 12px;
+  font-weight: 100;
```
owennw: We have variables for font weights (this might be `@font-light`? not sure which though.)

src/closedwindows/closedcard.less
```diff
@@ -0,0 +1,22 @@
...
+  font-size: 13px;
+}
+
+.closedcard {
+  height: 60px;
+  border-bottom: 1px solid @separator-colour;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  padding: 5px;
+  cursor: default;
```
owennw: Is the cursor line necessary?

src/closedwindows/closedcard.html
bjedrzejewski: True, just changing.

src/closedwindows/closedwindows-controller.js
```diff
@@ -0,0 +1,19 @@
+(function() {
+    'use strict';
+
+    class ClosedWindowsCtrl {
+        constructor(storeService) {
+            this.storeService = storeService;
+            this.numbers = [1, 2, 3];
```
owennw: Not entirely sure what this variable is for.

src/closedwindows/closedwindows-controller.js
```diff
@@ -0,0 +1,19 @@
...
+            this.refreshClosedWindows();
+        }
+        refreshClosedWindows() {
+            this.storeService.refreshStore();
+            this.closedWindows = this.storeService.getPreviousClosedWindows().sort((a, b) => b.closed - a.closed);
+        }
+    }
+    ClosedWindowsCtrl.$inject = ['storeService'];
+
+    angular.module('openfin.closedwindows', [])
```
owennw: Same comment about the injection here.

src/closedwindows/closedcard.html
bjedrzejewski: Well spotted.

src/closedwindows/closedcard.less
bjedrzejewski: Changed

src/sidebars/sidebar.html
```diff
@@ -16,6 +16,20 @@
                 </div>
             </div>
         </div>
+        <div ng-controller="ClosedWindowsCtrl as closedWindowsCtrl">
+            <div class="favClosed">
+                <icon name="closed_tabs" ng-show="true" icon-click="closedTabsShow = true"
```
owennw: The `ng-show` is redundant here.

src/closedwindows/closedcard.less
bjedrzejewski: Yes, otherwise it keeps changing as you mouse over and looks blah

src/sidebars/sidebar.html
```diff
@@ -16,6 +16,20 @@
...
+                      ng-click="$event.stopPropagation(); closedWindowsCtrl.refreshClosedWindows()"></icon>
```
owennw: I notice both `icon-click` and `ng-click` are being used here. Is there a reason for using both, or can they be combined?

src/closedwindows/closedwindows-controller.js
bjedrzejewski: Leftovers, removed.

src/closedwindows/closedwindows-controller.js
```diff
@@ -0,0 +1,19 @@
...
+
+    class ClosedWindowsCtrl {
+        constructor(storeService) {
+            this.storeService = storeService;
+            this.numbers = [1, 2, 3];
+            this.refreshClosedWindows();
+        }
+        refreshClosedWindows() {
+            this.storeService.refreshStore();
+            this.closedWindows = this.storeService.getPreviousClosedWindows().sort((a, b) => b.closed - a.closed);
```
owennw: I wonder whether the sorting should be performed in the HTML with angular. I think in general controllers are used for bindings, and the UI is responsible for knowing its own order. I think the native filter `orderBy` could be used to do it? I wonder what other people think though: @WPFerg ?

src/sidebars/sidebar.html
```diff
@@ -16,6 +16,20 @@
...
+            </div>
+            <div ng-show="closedTabsShow" class="favClosedCover" ng-click="closedTabsShow = false">
+                <div class="closedSelection" ng-click="$event.stopPropagation()">
+                    <div ng-repeat="closedwindow in closedWindowsCtrl.closedWindows">
```
owennw: This is where that ordering filter mentioned above would be done.

src/sidebars/sidebar.less
```diff
@@ -35,6 +35,33 @@
...
+}
+
+.closedSelection {
+  background: @light-background;
+  position: fixed;
+  bottom: 40px;
+  left: 240px;
+  width: 180px;
+  height: 300px;
+  box-shadow: 2px 2px 2px 2px @dark-background;
```
owennw: Is this what's doing the drop-shadow? That could be useful for our other windows...

src/sidebars/sidebar.html
bjedrzejewski: This is WIP, will not show if no windows were closed.

src/closedwindows/closedcard-directive.js
```diff
@@ -0,0 +1,14 @@
+(function() {
+    'use strict';
+
+    angular.module('openfin.closedcard')
+        .directive('closedcard', [function() {
```
owennw: Since you started this I changed all the `function()` over to `() =>`, so would be good to change this one too!

src/sidebars/sidebar.html
bjedrzejewski: good point

src/closedwindows/closedwindows-controller.js
bjedrzejewski: I was wondering the same to be honest. Can refactor to get the orderBy

src/closedwindows/closedcard-controller.js
```diff
@@ -0,0 +1,13 @@
...
+        .controller('ClosedcardCtrl', ClosedcardCtrl);
+}());
```
owennw: In fact.. is this controller used/necessary?

src/closedwindows/closedcard.less
```diff
@@ -0,0 +1,22 @@
+.closedtime {
+  font-size: 12px;
+  font-weight: 100;
+  line-height: 1.4;
+}
+
+.closedstock {
```
owennw: More camel-casing here too.

src/store-service.js
```diff
@@ -104,6 +107,16 @@
                 .map((store) => store.id);
         }
 
+        getPreviousClosedWindowNames() {
```
owennw: I think this function is only called in here? If so it can be performed inline and taken off the interface.

src/closedwindows/closedcard.less
bjedrzejewski: Shall we standardised on `closed-time` etc. - using `-` instead?

src/store-service.js
bjedrzejewski: I think it helps readability.

src/closedwindows/closedcard-directive.js
bjedrzejewski: Why did you change all functions to () => ?

src/closedwindows/closedcard-controller.js
bjedrzejewski: Yes, it is WIP.

src/app.js
owennw: `minichart` is a bit of an oddball, perhaps it should be too. In any case, `currentWindow` and I think everything else should be.

src/closedwindows/closedcard-controller.js
owennw: Talked about IRL.

src/closedwindows/closedcard-directive.js
owennw: `function()` and `() =>` behave slightly different in the way they handle `this`; we'd had some bugs using the old pattern before so to remain consistent they were all changed. Hopefully it'll minimise the change of unexpected bugs around this in the future.

src/closedwindows/closedcard.less
owennw: Yes probably, though for all the others perhaps an issue should be raised.

bjedrzejewski:
@owennw ok, changes made. Can you merge when you have time? I will finish some of the 'empty controller' things with the real functionality, no point committing just to revert now.

src/store-service.js
```diff
@@ -104,6 +107,16 @@
...
+            return (this.storage || [])
+                .filter((store) => store.closed > 0)
+                .map((store) => store.id);
+        }
+
+        getPreviousClosedWindows() {
+            return this.getPreviousClosedWindowNames().map((name) => this.open(name).store);
```
owennw: This feels like a roundabout way to get the object you actually want. We know this object exists in the store ready, so I don't think we need to go to the trouble of creating a `StoreWrapper` by injecting the store object, just to fetch the object back off of it.

bjedrzejewski: I don't understand the comment, can we talk IRL?

owennw:
I haven't finished reviewing it yet...

bjedrzejewski:
Pull request #391 we can close this.

PR #371, opened 2016-02-16
Hover colour

owennw:
Changes the colour of the empty space left behind by a tearout card.


src/sidebars/favourites/favourites.less
```diff
@@ -20,6 +20,7 @@
 
 .drop-target {
   height: @favourite-height;
+  background-color: #2E5356;
```
WPFerg: Since this hexcode's only used the once I think it's fine to merge in.

Issue #322, opened 2016-02-10
Can overwrite local storage between windows


WPFerg:
Windows can write to local storage, but they overwrite any changes since the creation of that window.

owennw:
@WPFerg Is this definitely true? If so, do you have reproducible steps?

owennw:
I can't reproduce this, if you continue to have problems feel free to reopen the issue.

PR #380, opened 2016-02-18
Persist compact view across runs

owennw:
Restore all windows to their previous state, now including compact view. Note when the app is requesting the services off the `window` object they may not be there yet. Angular will notice when the services are present and update accordingly. (E.g., on the `mainCtrl` when checking whether the window should be compact the service may not be present. However, when it does arrive this will be noticed. On my machine I don't see any weird visual behaviour when starting the app so I hope it's fast enough).

Also fixes a small bug where restoring from maximised via a drag didn't change the icon from restore back to maximise (a `function()` instead of `() =>` oversight).

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -18,7 +18,7 @@
         'openfin.scroll'
     ]);
 
-    angular.module('openfin.main', ['openfin.currentWindow']);
+    angular.module('openfin.main', []);
     angular.module('openfin.showcase', ['openfin.selection', 'openfin.quandl']);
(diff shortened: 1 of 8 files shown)


WPFerg:
I'm getting some unexpected behaviour.
For example, an app with one window: The window's opening on compact view, then expanding it and closing it in quick succession (and the opposite for if it opens in expanded view). Reopening the app has the correct position as last time (when it was expanded), but is still compact. I'm happy to classify it as part of #322 and its mysteries.
The compact view does apply once adequate time has been given so I'm happy to merge.

PR #383, opened 2016-02-18
Compact search improvements

akhmorgan:


src/sidebars/sidebar.less
```diff
@@ -100,20 +100,25 @@
...
     background: transparent;
   }
+  .favourites {
+    margin-top: 34px;
+  }
   .favourites, #favScroll, .favourite {
     width: 100%;
   }
   .search {
+    transition: all 0.1s linear 0.2s;
```
owennw: This feels a bit too fast; can you use the `@transition` variable instead?

src/sidebars/search/search-controller.js
```diff
@@ -32,6 +32,10 @@
                 // Down
                 this.changePointer(1);
             }
+            else if (event.keyCode === 27) {
+                // Escape
+                this.$scope.sidebarCtrl.favouritesClick();
+            }
         }
```
owennw: This also applies in the main view. However, I think I'm okay with that.

owennw:
There's a very small hangover for when the app is expanding -- you can see the favourites tab grow vertically upwards. Besides that all good!

src/sidebars/sidebar.less
akhmorgan: Yeah that's fine

PR #390, opened 2016-02-19
Add config service

### LINKED ISSUE #332: Create config file for shared values

WPFerg:
Fixes #332

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -15,7 +15,8 @@
         'openfin.star',
         'openfin.tearout',
         'openfin.minichart',
-        'openfin.scroll'
+        'openfin.scroll',
+        'openfin.config'
     ]);
 
     angular.module('openfin.main', []);
@@ -34,4 +35,5 @@
     angular.module('openfin.currentWindow', []);
     angular.module('openfin.window', []);
     angular.module('openfin.scroll', []);
+    angular.module('openfin.config', []);
 }());
--- a/src/app.js
+++ b/src/app.js
@@ -33,7 +33,6 @@
     angular.module('openfin.store', []);
     angular.module('openfin.parent', ['openfin.window']);
     angular.module('openfin.currentWindow', []);
-    angular.module('openfin.window', []);
     angular.module('openfin.scroll', []);
     angular.module('openfin.config', []);
 }());
--- a/src/app.js
+++ b/src/app.js
@@ -3,7 +3,6 @@
 
     angular.module('OpenFinD3FC', [
         'ngAnimate',
-        'openfin.parent',
         'openfin.main',
         'openfin.showcase',
         'openfin.toolbar',
@@ -15,8 +14,7 @@
         'openfin.star',
         'openfin.tearout',
         'openfin.minichart',
-        'openfin.scroll',
-        'openfin.config'
+        'openfin.scroll'
     ]);
 
     angular.module('openfin.main', []);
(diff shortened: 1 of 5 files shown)


src/config-service.js
```diff
@@ -0,0 +1,72 @@
+(function() {
+
+    const RESIZE_NO_LIMIT = 50000;
+
+    class ConfigService {
+        constructor() {
+        }
```
owennw: This can be removed.

src/config-service.js
```diff
@@ -0,0 +1,72 @@
...
+
+        nameCreator() {
```
owennw: maybe `createName` is a more appropriate method name?

src/window-service.js
```diff
@@ -195,7 +147,13 @@
 
             var mainWindow;
             if (name) {
-                mainWindow = new fin.desktop.Window(createConfig(name, isCompact), () => {
+                var config;
+                if (isCompact) {
+                    config = this.configService.getCompactConfig(name);
+                } else {
+                    config = this.configService.getWindowConfig(name);
+                }
```
owennw: A ternary is more succinct here.

src/app.js
```diff
@@ -34,4 +35,5 @@
     angular.module('openfin.currentWindow', []);
     angular.module('openfin.window', []);
     angular.module('openfin.scroll', []);
+    angular.module('openfin.config', []);
 }());
```
owennw: I don't think this is necessary?

src/window-service.js
WPFerg: I decided against that because the line would get quite long and potentially confusing. I don't mind, though

src/config-service.js
```diff
@@ -0,0 +1,72 @@
...
+                resizable: false,
+                showTaskbarIcon: false,
+                saveWindowState: false,
+                maxWidth: 230,
+                maxHeight: 100,
+                url: 'tearout.html'
+            };
+        }
+    }
+    ConfigService.$inject = [];
```
owennw: It could be worth having a common (pseudo-)private method that contains the config common to all 3. At the same time, it not be worth it.

owennw:
Much cleaner! Just a few small comments.

src/app.js
WPFerg: You're right, I'll remove it

src/config-service.js
WPFerg: Here's the commonality:
```
{
           name: name || this.createName(),
           autoShow: false,
           frame: false
}
```
Just for three options, I'm not sure it's worth the overhead. Were it much more, then I'd agree

src/window-service.js
owennw: It can be on multiple lines.

WPFerg: Done

src/app.js
```diff
@@ -32,6 +33,6 @@
     angular.module('openfin.store', []);
     angular.module('openfin.parent', ['openfin.window']);
     angular.module('openfin.currentWindow', []);
-    angular.module('openfin.window', []);
     angular.module('openfin.scroll', []);
+    angular.module('openfin.config', []);
```
owennw: I think you forgot to remove this.

WPFerg: I can remove it, but I'll need to add an exclusion into the gruntfile. If I'm doing so, I'd also like to remove `openfin.parent` since that isn't used in the child windows.

akhmorgan:
looks good to me

src/app.js
owennw: An exclusion for the concat?

PR #391, opened 2016-02-22
Closed windows restore - ui

bjedrzejewski:
This is the UI part for closed windows restore


src/closedWindows/closedCard-directive.js
```diff
@@ -0,0 +1,14 @@
+(function() {
+    'use strict';
+
+    angular.module('openfin.closedCard')
+        .directive('closedCard', [() => {
+            return {
+                restrict: 'E',
+                templateUrl: 'closedWindows/closedCard.html',
+                scope: {
+                    cwindow: '='
```
akhmorgan: I'd prefer to have this variable with the full name of `closedWindow` rather than `cwindow`. This is probably personal preference though.

akhmorgan:
Looks good to me. I had one small comment but that's it.

src/closedWindows/closedCard-directive.js
owennw: I think I agree, I wasn't sure what `cwindow` was for a while (in fact it's not a window either, but instead a JSON object representing the window state)

bjedrzejewski:
@akhmorgan @owennw windows open now, but closing open window often causes the whole app to close. Can you help investigating? This is something to do with window-service I think... Also removing open window from recently closed- still left to do.

bjedrzejewski:
It's the second open/close operation that closes the app.

owennw:
The problem is now fixed in #398.

bjedrzejewski:
@akhmorgan I have renamed the variable to cardWindow to make it clearer.

PR #395, opened 2016-02-22
Tearout from compact

owennw:


src/store-service.js
```diff
@@ -58,8 +58,8 @@
             this.update(stock);
         }
 
-        toggleCompact(isCompact) {
-            this.store.compact = isCompact;
```
akhmorgan: I'm wary of strict toggle functions. I would prefer to have the ability to pass in an optional boolean and if it's present just set `compact` to that. However I suppose that's an easy addition if it becomes necessary in the future

akhmorgan:
looks good to me.

PR #398, opened 2016-02-22
Dispose of old event listeners correctly

owennw:
Correctly disposes of old event listeners when windows (main and tearout) are closed.


owennw:
Resubmitting because Travis.

PR #399, opened 2016-02-22
Dispose of old event listeners correctly

owennw:
copy of #398

### DIFF
--- a/src/sidebars/favourites/tearout.js
+++ b/src/sidebars/favourites/tearout.js
@@ -16,6 +16,7 @@
 
                         var windowService = window.windowService;
                         var tearoutWindow = windowService.createTearoutWindow(window.name);
+                        var me = {};
 
                         function initialiseTearout() {
                             var myDropTarget = tearElement.parentNode,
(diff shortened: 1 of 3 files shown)


src/sidebars/favourites/tearout.js
```diff
@@ -16,6 +16,7 @@
 
                         var windowService = window.windowService;
                         var tearoutWindow = windowService.createTearoutWindow(window.name);
+                        var me = {};
```
bjedrzejewski: Why do you call this var `me`?

owennw: That's mostly a hangover from the original tearout example, but not something that's changed in this PR.

bjedrzejewski: Yeah I have seen it was already there, if you think it makes sense to give it better name, we can do it, otherwise I am VERY happy for you to merge it (as it fixes bug in my feature :D )

owennw: Let's do that for another PR :)

PR #404, opened 2016-02-23
Set window to open when recovered

owennw:
This removes the entry from the recently closed windows when the window is restored.

### DIFF
--- a/src/closedWindows/closedCard-controller.js
+++ b/src/closedWindows/closedCard-controller.js
@@ -4,6 +4,7 @@
     class ClosedCardCtrl {
         openClosedWindow(name) {
             window.windowService.createMainWindow(name, false);
+            window.storeService.open(name).openWindow();
         }
     }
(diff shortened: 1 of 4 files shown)


src/closedWindows/closedCard-controller.js
```diff
@@ -3,7 +3,9 @@
 
     class ClosedCardCtrl {
         openClosedWindow(name) {
-            window.windowService.createMainWindow(name, false);
+            var store = window.storeService.open(name);
+            window.windowService.createMainWindow(name, store.isCompact());
+            store.openWindow();
```
bjedrzejewski: Does this not belong in the createMainWindow method? Putting it in a controller is not very nice if anything that creates new window will have to call it (store.openWindow)

owennw: Hmm, perhaps yeah.

PR #407, opened 2016-02-23
Tweak closed cards styling

owennw:


src/closedWindows/closedCard.less
```diff
@@ -1,22 +1,25 @@
...
 .closed-stock {
-  font-size: 13px;
+  font-size: 12px;
+  font-weight: 400;
 }
 
 .closed-card {
-  height: 60px;
-  border-bottom: 1px solid @separator-colour;
+  height: 70px;
```
bjedrzejewski: If you have 5 closed tabs, this won't work.

bjedrzejewski: image

owennw: Yes I was thinking that the window would auto-resize and we just hadn't done that yet, but I notice before it allowed 5 so I'll push a commit to keep that with the new sizes.

owennw:
You could've just merged it outright, there's no need to merge dev back into this first.

PR #411, opened 2016-02-24
Bug fixes

owennw:
Fixes:

Error with parent window trying to find the old bitflux file
Not showing default stocks on first time start up
Expanding from compact view on a torn out window


bjedrzejewski:
I will check out and have a play. Code looks good.

bjedrzejewski:
@owennw where is the fix for: 

> Not showing default stocks on first time start up 

?

owennw:
It's passing `newWindow.name` into the window created CB store service, instead of `name`.

PR #413, opened 2016-02-24
When search expands, auto focus on the input field.

### LINKED ISSUE #373: autofocus on search input when it is shown

bjedrzejewski:
Closes #373

### DIFF
--- a/src/sidebars/sidebar-controller.js
+++ b/src/sidebars/sidebar-controller.js
@@ -7,13 +7,16 @@
     };
 
     class SidebarCtrl {
-        constructor() {
+        constructor($scope) {
             this._favouritesClass = classes.expanded;
             this._searchClass = classes.contracted;
 
             this._showSearches = false;
             this._showFavourites = true;
             this._searchSmall = true;
+            this.$scope = $scope;
+
+            this._watch();
         }
 
         searchClass() {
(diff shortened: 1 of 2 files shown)


akhmorgan:
looks good to me.

src/sidebars/sidebar-controller.js
```diff
@@ -51,8 +54,25 @@
                 this._showFavourites = true;
             }
         }
+
+        _showSearchesChanged() {
+            var showSearches = this.showSearches();
+            if (showSearches) {
+                //Couldn't get it to work reliable in a synchronous/promise fashion
+                setTimeout(function() {
```
owennw: `setTimeout` shouldn't be used in angular apps. Instead inject `$timeout` and use that. Also would prefer the ES6 style function syntax.

src/sidebars/sidebar-controller.js
```diff
@@ -51,8 +54,25 @@
                 this._showFavourites = true;
             }
         }
+
+        _showSearchesChanged() {
+            var showSearches = this.showSearches();
+            if (showSearches) {
+                //Couldn't get it to work reliable in a synchronous/promise fashion
```
owennw: Not sure if you need the apologetic comment :P

bjedrzejewski: This is a disguised 'todo' for someone other than myself :D

owennw:
Seems like the feature doesn't work in compact mode, unfortunately. But it's good in normal!

akhmorgan:
It won't work in compact view because it's getting the input by id and the compact view has it's own search input, which is the second element in the sidebar html with that id. So it's only ever getting the full-view search input.

WPFerg:
Perhaps having something in the directive on expand would be more straight forward, then?

bjedrzejewski:
@akhmorgan where is that second input in the code? Which file?

akhmorgan:
@bjedrzejewski in sidebar.html you'll see two `<search>` elements

bjedrzejewski:
@akhmorgan ok I got it now. I have a nice fix I think.

bjedrzejewski:
@WPFerg @akhmorgan I will move to react for a bit now. If anyone want's to fix it (I failed without jQuery for now), feel free to add a commit!

WPFerg:
I think I've got a cleaner fix doing something different, I'll close this and open a new PR with it.

PR #437, opened 2016-02-25
Autofocus on search click

WPFerg:


akhmorgan:
that should do it. Looks good to me

PR #439, opened 2016-02-25
Keep window on screen on window start up

### LINKED ISSUE #434: You can tear out windows so that they are off screen
- [x] If you take a tile, and drag it to the bottom right of your desktop, the new window appears entirely off screen.

- [x] i don't think the new windows should appear with it's top left corner at the mouse location - it should be offset.

WPFerg:
Fixes #434

### DIFF
--- a/src/window-service.js
+++ b/src/window-service.js
@@ -142,6 +142,7 @@
 
                 newWindow.show();
                 newWindow.bringToFront();
+                this.snapToScreenBounds(newWindow);
             };
 
             var mainWindow;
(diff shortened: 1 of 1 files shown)


src/window-service.js
```diff
@@ -203,24 +225,6 @@
...
-            } else {
-                _window.updateOptions({
-                    resizable: true,
-                    minHeight: 510,
-                    minWidth: 918,
-                    maximizable: true
-                });
-            }
-        }
-
```
owennw: Ha, this must have been a merge fail at one point, good spot!

src/window-service.js
```diff
@@ -177,6 +178,27 @@
...
         }
 
+        snapToScreenBounds(targetWindow) {
+            fin.desktop.System.getMonitorInfo((info) => {
+                targetWindow.getBounds((bounds) => {
+
+                    if (bounds.left < info.virtualScreen.left) {
+                        bounds.left = info.virtualScreen.left;
+                    } else if (bounds.left + bounds.width > info.virtualScreen.right) {
+                        bounds.left = info.virtualScreen.right - bounds.width;
```
owennw: I take it you can't move `bounds.right` and have that shunt the whole window across?

src/window-service.js
```diff
@@ -177,6 +178,27 @@
...
+                    }
+
+                    if (bounds.top < info.virtualScreen.top) {
+                        bounds.top = info.virtualScreen.top;
```
owennw: Does this `info(.virtualScreen)` have any info on the taskbar position?

WPFerg: It does, http://cdn.openfin.co/jsdocs/stable/tutorial-system.getMonitorInfo.html, should I add in taskbar compensation too?

src/window-service.js
WPFerg: There's no `bounds.right`, unfortunately

src/window-service.js
owennw: I think so yes. If the top of the window ended up behind the taskbar it would be unmovable.

WPFerg: Added. It shrinks the size of the working area.

Issue #435, opened 2016-02-25
tear-out should not require you to drag outside of the window


ColinEberhardt:
Currently you have to drag a tile until it is entirely outside of the windows bounds. I think it should tear out as soon as you drag beyond the favourites pane bounds (possibly plus a small tolerance).

i.e. copy Chrome ;-)

owennw:
Would this also be the behaviour for tear in? I.e. the card will only tear in when intersected with a favourites pane.

owennw:
Closed by #440

PR #440, opened 2016-02-25
Tearout from favourites tab

owennw:
Dragging a card outside of the favourites tab and releasing will now create a new window. Note that tearing in still applies to the whole window. Perhaps it should be changed for symmetry too? #435

### DIFF
--- a/src/sidebars/favourites/tearout.js
+++ b/src/sidebars/favourites/tearout.js
@@ -193,7 +193,8 @@
                                 var hoverTargets = hoverService.get();
 
                                 for (var i = 0, max = hoverTargets.length; i < max; i++) {
-                                    var overDropTarget = geometryService.elementIntersect(tearoutWindow, window, hoverTargets[i].hoverArea);
+                                    var overDropTarget = geometryService.elementIntersect(
+                                        tearoutWindow, window, hoverTargets[i].hoverArea);
 
                                     if (overDropTarget) {
                                         if (!store) {
@@ -208,7 +209,8 @@
 
                             me.boundsChangingEvent = () => {
                                 // Check if you are over a drop target by seeing if the tearout rectangle intersects the drop target
-                                insideFavouritesPane = geometryService.elementIntersect(tearoutWindow, window, document.getElementsByClassName('favourites')[0]);
+                                insideFavouritesPane = geometryService.elementIntersect(
+                                    tearoutWindow, window, document.getElementsByClassName('favourites')[0]);
 
                                 if (insideFavouritesPane) {
                                     reorderFavourites();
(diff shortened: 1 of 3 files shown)


src/sidebars/favourites/tearout.js
```diff
@@ -167,11 +167,9 @@
                                                 dragService = null;
                                             } else {
                                                 // Create new window instance
-                                                var mainApplicationWindowPosition = geometryService.getWindowPosition(window);
-
                                                 var compact = store.isCompact();
                                                 windowService.createMainWindow(null, compact, (newWindow) => {
-                                                    newWindow.resizeTo(mainApplicationWindowPosition.width, mainApplicationWindowPosition.height, 'top-left');
+                                                    newWindow.resizeTo(window.outerWidth, window.outerHeight, 'top-left');
```
WPFerg: Does this resize the new window to the size of the window you dragged out from?

src/sidebars/favourites/tearout.js
```diff
@@ -191,13 +189,11 @@
...
 
-                            function reorderFavourites(tearoutRectangle) {
+                            function reorderFavourites() {
                                 var hoverTargets = hoverService.get();
 
                                 for (var i = 0, max = hoverTargets.length; i < max; i++) {
-                                    var dropTargetRectangle = geometryService.rectangle(
-                                        geometryService.elementScreenPosition(geometryService.getWindowPosition(window), hoverTargets[i].hoverArea)),
-                                        overDropTarget = tearoutRectangle.intersects(dropTargetRectangle);
+                                    var overDropTarget = geometryService.elementIntersect(tearoutWindow, window, hoverTargets[i].hoverArea);
```
WPFerg: This and https://github.com/ScottLogic/bitflux-openfin/pull/440/files#diff-3709fe15a75397a1b30d1833b8c85403R211 are quite long lines, could you split them up?

src/sidebars/favourites/tearout.js
owennw: Yes, just unravelling the pointless call to the geometry service, the code is identical though.

src/sidebars/favourites/tearout.js
owennw: They're both within the line character limit =/

WPFerg:
I'll hold off from merging this for now until we decide whether it's worth applying this to tearing in.

owennw:
How shall we decide?

WPFerg:
I'm not sure, I think it's probably best if it's symmetrical. How difficult would it be to do?

owennw:
I'll merge this as I think it provides useful functionality on its own. With it merged we can re-test and see whether implementing the symmetrical idea is worthwhile.

WRT the difficulty, it's a matter of finding the favourites tab element in the global window service when comparing intersections of windows. Not sure how cumbersome that would be.

Issue #424, opened 2016-02-25
Improve the release process


ColinEberhardt:
We need to know:
 - [x] When a user reports an issue, which version it relates to
 - [x] Provide documentation of the install process (where is it installed to?)

owennw:
Interestingly, because of the way we're hosting that app at the moment, there's only one version a user can use. What is our plan for hosting the app for a release?

ColinEberhardt:
I think github pages is fine for release.

Even though a user can only see one version, it would make sense to track the version that a bug was reported against. Something like help -> about (but less ugly!)

WPFerg:
@ColinEberhardt Where should the install process documentation go? `README.md`?

ColinEberhardt:
@WPFerg yeah, README makes sense.

I think the outline of this file should be roughly as follows:

- A very brief overview of what this project is in just a few sentences, maybe a screenshot
- End-user documentation, i.e. installation instructions
- Developer documentation (the current README contents, although the use of `--target` needs to be included!)

WPFerg:
Re `target`: #448, but I'll add screenshots/installation instructions :smile:

WPFerg:
#453 and #443 solve this.

PR #443, opened 2016-02-26
Add version info

WPFerg:
Adds a version number in the lower left corner for #424 
image

### DIFF
--- a/src/sidebars/sidebar.less
+++ b/src/sidebars/sidebar.less
@@ -35,6 +35,16 @@
   -webkit-animation-name: favourites;
 }
 
+.version {
+  bottom: 15px;
+  left: 70px;
+  position: absolute;
+  color: @icon-colour;
+  font-weight: @font-light;
+  font-size: 11px;
+  cursor: default;
+}
+
 .favClosed {
   bottom: 15px;
   left: 246px;
--- a/src/sidebars/version-directive.js
+++ b/src/sidebars/version-directive.js
@@ -1,12 +1,13 @@
 (function() {
     'use strict';
+    const VERSION = {version: '8.0.0'};
 
     angular.module('openfin.version')
-        .value('version', '0.1.0')
+        .value('version', VERSION)
         .directive('insertVersion', ['version', (version) => {
             return {
                 restrict: 'A',
-                template: version
+                template: version.version
             };
         }]);
 }());
--- a/src/sidebars/version-directive.js
+++ b/src/sidebars/version-directive.js
@@ -1,0 +1,12 @@
+(function() {
+    'use strict';
+
+    angular.module('openfin.version')
+        .value('version', '0.1.0')
+        .directive('insertVersion', ['version', (version) => {
+            return {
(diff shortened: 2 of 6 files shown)


src/sidebars/version-directive.js
```diff
@@ -0,0 +1,12 @@
+(function() {
+    'use strict';
+
+    angular.module('openfin.version')
+        .value('version', '0.1.0')
```
owennw: What's the rationale behind this number? And do we plan on the release grunt task to automate incrementing it, or will it be manual?

src/sidebars/sidebar.less
```diff
@@ -35,6 +35,16 @@
...
 
+.version {
+  bottom: 15px;
+  left: 70px;
+  position: absolute;
+  color: @icon-colour;
+  font-weight: @font-light;
+  font-size: 11px;
+  cursor: default;
+}
```
owennw: Is it worth getting this signed off by Rui?

src/sidebars/version-directive.js
WPFerg: No real rationale, just a number sub-1 since we're not on public release yet (and it'll change shortly anyway). I'll look into grunt tasks for releasing.

src/sidebars/sidebar.less
WPFerg: I'm not sure, it's just a small number in a corner of the page, but I'm happy to do so

owennw: Yeah just a quick skype message and pointing him to this issue I guess.

akhmorgan: It's a very minor fix to take it out or change it if he doesn't like it. So I'd say merge first, ask questions later in this case

src/sidebars/version-directive.js
akhmorgan: I'd say that grunt should increment this. Devs will no doubt forget at least once.

WPFerg: There's a grunt task that I've spotted that does this and also tags Git appropriately with releases, maybe it's worth using that?

bjedrzejewski: @WPFerg I strongly believe that we should match this number to our GH number, not sub it artificially. I don't see a problem with a first release being version 11.0.0 or something like this. Otherwise this number doesn't help much.

src/sidebars/sidebar.less
bjedrzejewski: @akhmorgan agreed

src/sidebars/version-directive.js
WPFerg: @bjedrzejewski No problem. If this grunt task works, that's what it'll default to, anyway

WPFerg: Added the bumper. I've kept it separated from `deploy`, since I can't verify it works, but `grunt release` bumps the version.

owennw:
I'm going to take it on faith that this works, as testing it isn't the easiest..

PR #445, opened 2016-02-26
Change default data amount to 1200 points- about 3 years.

bjedrzejewski:
I think we should merge this first, to close the issue, rather than be dependent on bitflux adding features. This can be customized later as a separate feature.

Performance is good (can't see a difference).

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "angular-animate": "^1.4.8",
     "angular-resource": "^1.4.8",
     "d3fc": "^5.2.0",
-    "BitFlux": "ScottLogic/BitFlux#cf81eab098cf6c6f4a4bd36269d9cfe21d4b449b",
+    "BitFlux": "bjedrzejewski/BitFlux#dddca873ab952c7d9f792c54155b4b73594e7ab3",
     "jquery": "^1.11.1",
     "malihu-custom-scrollbar-plugin": "^3.1.3",
     "moment": "^2.10.6"


package.json
```diff
@@ -31,7 +31,7 @@
     "angular-animate": "^1.4.8",
     "angular-resource": "^1.4.8",
     "d3fc": "^5.2.0",
-    "BitFlux": "ScottLogic/BitFlux#cf81eab098cf6c6f4a4bd36269d9cfe21d4b449b",
+    "BitFlux": "bjedrzejewski/BitFlux#dddca873ab952c7d9f792c54155b4b73594e7ab3",
```
owennw: Which version of BitFlux is this repo updated to?

owennw:
Would it be difficult for you to make a change to BitFlux to add this to their API and PR that in? @jleft

package.json
bjedrzejewski: Latest dev

bjedrzejewski:
I think expanding API would be ideal. This could go together with the API for setting and getting indicators.

package.json
owennw: The current BitFlux we're using is https://github.com/ScottLogic/BitFlux/commit/cf81eab098cf6c6f4a4bd36269d9cfe21d4b449b, so your change will actually be updated the version of BitFlux too. Do we want that?

bjedrzejewski: @owennw I would say say, but I don't have a strong preference.

akhmorgan: why is this now pointing to a forked version rather than the main bitflux repo?

akhmorgan: ah never-mind. Got it.

jleft:
We have an open PR to add this to the BitFlux API, it's a fairly straightforward change to make.

https://github.com/ScottLogic/BitFlux/pull/660

owennw:
Given the PR for BitFlux has now been merged, and this PR changes our version to use the latest dev anyway, can you re-work this PR to use the BitFlux repo and access the new API method?

bjedrzejewski:
This will be reworked using new api

PR #452, opened 2016-02-26
High level class comments

owennw:
Didn't want to go overboard here -- just added comments for classes that don't have a strict coupling to UI elements.


ColinEberhardt:
That level of documentation looks fine to me - thanks :-)

PR #454, opened 2016-02-26
Open repo when clicking on version number

### LINKED ISSUE #412: Include link to GH page in the app

WPFerg:
An attempt to resolve #412 without having to add an icon, and it shouldn't be too hard to add an icon in any case.

### DIFF
--- a/src/sidebars/sidebar-controller.js
+++ b/src/sidebars/sidebar-controller.js
@@ -7,7 +7,8 @@
     };
 
     class SidebarCtrl {
-        constructor() {
+        constructor(currentWindowService) {
+            this.currentWindowService = currentWindowService;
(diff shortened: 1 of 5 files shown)


src/sidebars/sidebar-controller.js
```diff
@@ -51,6 +51,10 @@
                 this._showFavourites = true;
             }
         }
+
+        openGithub() {
+            fin.desktop.System.openUrlWithBrowser('https://github.com/ScottLogic/bitflux-openfin');
+        }
```
owennw: Looks good, but can we run the `fin` API call through the current window service? Should be safer by using `ready` too.

WPFerg: Done

Issue #451, opened 2016-02-26
Window Screen Snapping should take into account taskbar


WPFerg:
If a window would spawn off-screen or over-taskbar, move it to not be either.

See #447 #439 for additional context.

PR #455, opened 2016-02-26
Re-attempt for taskbar compensation

WPFerg:
#451

### DIFF
--- a/src/window-service.js
+++ b/src/window-service.js
@@ -204,9 +204,11 @@
 
                 for (var monitor of monitors) {
 
+                    let monitorRect = monitor.monitorRect;
+
                     // If the window's top-left is within the monitor's bounds, use that + stop
-                    if (x >= monitor.monitorRect.left && x <= monitor.monitorRect.right &&
-                            y >= monitor.monitorRect.top && y <= monitor.monitorRect.bottom) {
+                    if (x >= monitorRect.left && x <= monitorRect.right &&
+                            y >= monitorRect.top && y <= monitorRect.bottom) {
 
                         callback(monitor);
                         return;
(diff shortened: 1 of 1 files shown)


src/window-service.js
```diff
@@ -196,20 +196,53 @@
...
+                        var midY = monitor.monitorRect.top + monitor.monitorRect.bottom / 2;
+                        var distance = Math.pow(midX - x, 2) + Math.pow(midY - y, 2);
+                        if (distance < closestDistance) {
+                            closestDistance = distance;
+                            closestMonitor = monitor;
+                        }
+
+                    }
+                }
+                callback(closestMonitor);
```
owennw: Maybe `monitor.monitorRect` should be a variable? And is there anything in the geometryService (or that should be added to it) that could help out here?

src/window-service.js
```diff
@@ -196,20 +196,53 @@
...
                     }
 
-                    if (bounds.top < info.virtualScreen.top) {
-                        bounds.top = info.virtualScreen.top;
-                    } else if (bounds.top + bounds.height > info.virtualScreen.bottom) {
-                        bounds.top = info.virtualScreen.bottom - bounds.height;
+                    if (bounds.top < monitor.availableRect.top) {
+                        bounds.top = monitor.availableRect.top;
+                    } else if (bounds.top + bounds.height > monitor.availableRect.bottom) {
+                        bounds.top = monitor.availableRect.bottom - bounds.height;
```
owennw: Similarly maybe `monitor.availableRect` should be a variable.

src/window-service.js
WPFerg: Possibly, the `intersects` function there is pretty similar, but `Rectangle`'s constructor requires `left, top, width, height`, opposed to `left, top, bottom, right`, so it mightn't be worth the overhead converting for quite a simple operation.

owennw: Fair enough. It does feel like a function that belongs in the geometry area, but it's not a big deal for the size and scope of the app.

PR #456, opened 2016-02-26
Make link more obvious

owennw:
Seems a bit more logical, without a tooltip in place.


WPFerg:
I don't mind either way, but would a tooltip give a better indication that it opens github?

owennw:
Yeah I don't know. Feels like we need some UX guidance.

akhmorgan:
octocat icon rather than 'github' text?

owennw:
Perhaps. Waiting on Rui to see what he prefers.

owennw:
Closing to resubmit aligned with #460

Issue #433, opened 2016-02-25
Compact view search styling issues


ColinEberhardt:
A couple of little issues:

 - [x] when you mouse-over a stock the star disappears.
 - [x] you can select items, which just seems a bit strange (see #430)
 - [x] it is not obvious how to close the search, perhaps clicking the title bar should close?

WPFerg:
Closed by #458 and #481

Issue #388, opened 2016-02-19
Hover state of search results in compact view is incorrect


owennw:
The hover colours were removed for the favourites tab, which also did the same for the search tab. Not sure if they should have hover state or not, but the stars disappear when hovered.

Issue #430, opened 2016-02-25
Should we allow selection of search results?


ColinEberhardt:
I can search for a stock, then select it rather than favourite it, then close the search pane. This leads to the application being in a somewhat inconsistent state - where none of the favourite stocks are selected.

owennw:
Why is that inconsistent? There is a text label above the chart that displays the currently selected stock.

ColinEberhardt:
Good point @owennw we'll keep the current behaviour.

owennw:
It could be that the label isn't obvious enough though! It's hard for me to tell given I've been using the app a lot.

PR #458, opened 2016-02-29
Compact style fixes

akhmorgan:
[#388](https://github.com/ScottLogic/bitflux-openfin/issues/388) & most of [#433](https://github.com/ScottLogic/bitflux-openfin/issues/433)


src/sidebars/search/search-controller.js
```diff
@@ -21,7 +21,12 @@
         }
 
         select(stock) {
-            this.selectionService.select(stock);
+            if (!this.store) {
+                this.store = window.storeService.open(window.name);
+            }
```
WPFerg: I don't think you need to do this here, since you're not using `this.store`

src/sidebars/search/search-controller.js
```diff
@@ -21,7 +21,9 @@
         }
 
         select(stock) {
-            this.selectionService.select(stock);
+            if (stock.favourite) {
+                this.selectionService.select(stock);
+            }
```
owennw: Not sure if I understand this. The selection can be changed from the compact view if the stock is a favourite, but why?

owennw:
One question, the rest looks good!

src/sidebars/search/search-controller.js
akhmorgan: That's basically because it felt weird to not be able to do that. So if you open the search in regular view, you can see your favourites and the one that's selected so being able to still select those favourites seems to make sense. It's easily removed if this is just my opinion though.

akhmorgan: Actually I may have misinterpretted the ticket. Perhaps selection should just be disabled across the board in compact. Easily done.

owennw: Yes I think so. The only time I think it should work is when double clicking a favourite.

src/sidebars/search/search-controller.js
akhmorgan: yes, I spotted that and it was removed in a later commit. However a subsequent commit put it back in as I did end up using it in the end

src/sidebars/search/search-controller.js
```diff
@@ -21,7 +21,12 @@
         }
 
         select(stock) {
-            this.selectionService.select(stock);
+            if (!this.store) {
+                this.store = window.storeService.open(window.name);
+            }
+            if (!this.store.isCompact() && stock.favourite) {
+                this.selectionService.select(stock);
+            }
```
owennw: Why can only favourites be selected? This is a change in functionality that isn't style.

akhmorgan: yeah, it's not style. The branch name might be a bit misleading but was based off the name of the issue #433, which includes a change in functionality in stating that search results should not be selectable. However favourited stocks are shown when you initially open the search, and it seems like those should be selectable. This change is to keep that existing functionality in full view that favourited stocks which are shown in search can still be selected, but nothing in compact view can be selected and in full view search results that aren't favourites already can't be selected. If we don't want the favourites in the initial search screen to be selectable, then I would say that they shouldn't be displayed when you first open search. It seems weird to open a view which shows the same favourites as in the favourites view but now you can't select them.

owennw: >and in full view search results that aren't favourites already can't be selected

In full view everything in the search tab can be selected, favourites or search results alike.

The way I understood it would work was for the search tab:

Full view - everything can be selected
Compact view - nothing can be selected

Though I might have misunderstood something.

akhmorgan: The way Colin put it was that you shouldn't be able to select search results, which to me would mean you can't select things that aren't in your favourites, because it seems strange to be able to select stuff straight from search. If you allow selection from search you can easily get into a state where the full view is showing the graph for something not in your favourites. Do we want that to be possible?

akhmorgan: We should probably just get this clarified later.

owennw: Yes that's fine, because there's the title above the charts for which stock is selected. See #430.

akhmorgan: Direct selection of search results still seems weird to me. It also seems to reduce the usefulness of having favourites at all. So I can search for something, select it, then I go back to the regular view look at the chart. I then select one of my favourites to look at, but then want to go back to the other chart that I was just viewing, if I forget the name of the stock I had selected from search, then there's no way to go back to that chart because the stock is not favourited. Seems like potential for bad UX to me. I'm happy to change it though, if I'm in the minority in that opinion.

owennw: I kind of see it like bookmarks. The favourites are the stocks you care about most, and you get to see the sparkline preview and current prices of them all the time. But you can also view the detailed data without having to add the stock to your favourites.

>So I can search for something, select it, then I go back to the regular view look at the chart.
That shouldn't be possible, because in compact view I don't think anything in search can be selected. After all, what would be the point of allowing that -- the user can't see anything change.

akhmorgan: > So I can search for something, select it, then I go back to the regular view look at the chart.

Yeah, in compact view it's not possible, but in regular view it is. I see your point about bookmarks, but I don't like the fact that if you select something not in your favourites to look at, then select one of your favourites, you have no way of getting back to the first selection without going back and searching for it again. Unless we want to track recent searches or something, but that seems like more work than it's worth.

akhmorgan: Anyway, in the interest of getting this merged I'm happy to push a change to make it so that you can select anything in full view. Unless anyone else wants to weigh in?

WPFerg: I'd say you should be able to select anything in full view *only*, to get a 'preview' of sorts. In compact view, to me at least, you'd only be searching to add it to favourites.

owennw:
Looks fine to me, except the merge issue?

WPFerg:
I'm getting this when I start up the app, is this what it should look like?
image

akhmorgan:
@WPFerg is the issue the highlight? Or is there actually no toolbar/searchbar? I'm assuming the latter is cropping. Re: highlight: that's not a change I made.

WPFerg:
@akhmorgan Yeah, it was the highlight. I think it's because of the additional `dark` class on the selection and `background: transparent` on it.

akhmorgan:
hmm. Ok, I thought that was a deliberate change made by someone on dev, as it only appeared after a rebase. However it seems to just be on this branch so I'll fix it.

src/sidebars/sidebar.less
```diff
@@ -153,21 +153,22 @@
   #search-scroll, #favourite-scroll{
      bottom: 34px;
   }
+
   #favourite-scroll {
-     top: 34px;
+	top: 34px;
+    .dark, .darkens:hover {
```
WPFerg: There's an indentation mismatch here, but it'll be good to go after that

PR #466, opened 2016-02-29
By default fetch 1200 data points.

bjedrzejewski:


owennw:
Travis did complete and pass, but didn't report back due to #465

Issue #475, opened 2016-03-01
Compact view styling broken


owennw:
Broken by #469 (bottom border line appears) and #460 (when search tab opened it appears over the search input field).

PR #476, opened 2016-03-01
Hide closed window; fix search result positioning #475

WPFerg:
#475

### DIFF
--- a/src/sidebars/sidebar.less
+++ b/src/sidebars/sidebar.less
@@ -150,8 +150,13 @@
     -webkit-app-region: drag;
   }
   #search-scroll, #favourite-scroll{
-     top: 34px;
      bottom: 34px;
+  }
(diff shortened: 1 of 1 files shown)


src/sidebars/sidebar.less
```diff
@@ -187,4 +192,7 @@
     margin-top: 7px;
     margin-right: 0px;
   }
+  .closed-window-selection {
+    display: none;
+  }
```
owennw: In the original design spec this icon should be moved to the toolbar and resized (#405), but I see that's assigned to @akhmorgan. If he hasn't done that already it looks like something orthogonal to other compact view changes -- maybe you can take it?

WPFerg: I can do, I'd say that's another PR, though.

PR #479, opened 2016-03-01
Fix Bitflux selection not occurring on compact select+expand

WPFerg:
Performs an additional check on start up for a selection, which happens on contact view expand. This is done because if you expand from contact view, it uses the data generator instead of stocks.

### DIFF
--- a/src/showcase/showcase-directive.js
+++ b/src/showcase/showcase-directive.js
@@ -16,6 +16,15 @@
                     chart.periodsOfDataToFetch(configService.getBitfluxStockAmount());
                     chart.proportionOfDataToDisplayByDefault(configService.getInitialBitfluxProportion());
 
+                    // If there's already a selection, run the chart and use that.
+                    // This occurs when the user selects a stock in compact view
+                    // and it expands.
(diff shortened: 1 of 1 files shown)


src/showcase/showcase-directive.js
```diff
@@ -16,6 +16,15 @@
                     chart.periodsOfDataToFetch(configService.getBitfluxStockAmount());
                     chart.proportionOfDataToDisplayByDefault(configService.getInitialBitfluxProportion());
 
+                    // If there's already a selection, run the chart and use that.
+                    // This occurs when the user selects a stock in compact view
+                    // and it expands.
+                    if (scope.selection && scope.selection()) {
+                        firstRun = false;
+                        chart.run(element[0].children[0]);
```
owennw: The code in the watch only calls `run` for the first run -- does it need to be called every time here?

WPFerg: This is only run when it's initially linked (which is on app start, compact expand), so would only be run once anyway

WPFerg:
@owennw Brought this up to latest dev

PR #480, opened 2016-03-01
Initial Indicator + Indicator Persist. 

### LINKED ISSUE #431: The initial chart is a bit boring!
Let's chuck in a bunch of indicators for the default view ...

WPFerg:
closes #431 + #438

### DIFF
--- a/src/showcase/showcase-directive.js
+++ b/src/showcase/showcase-directive.js
@@ -12,10 +12,15 @@
                 link: (scope, element) => {
                     var chart = bitflux.app().quandlApiKey(quandlService.apiKey()),
                         firstRun = true,
-                        store;
+                        store,
+                        initStore = () => {
+                            if (!store && window.storeService) {
+                                store = window.storeService.open(window.name);
+                            }
+                        };
 
-                    if (window.storeService) {
-                        store = window.storeService.open(window.name);
+                    initStore();
+                    if (store) {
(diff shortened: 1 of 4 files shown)


src/showcase/showcase.html
```diff
@@ -1 +1 @@
-<div id="showcase-container" selection="selection()"></div>
+<div id="showcase-container" selection="selection()" ng-click="save()"></div>
```
WPFerg: This was the best way I could think of to force an update of the indicators, any other ideas would be great

src/showcase/showcase-directive.js
```diff
@@ -11,23 +11,49 @@
...
                             if (newSelection !== previousSelection) {
                                 chart.changeQuandlProduct(newSelection);
                             }
                         }
                     });
+
+                    scope.$watchCollection(() => store.indicators(), (newValues) => {
+                        if (!store && window.storeService) {
+                            store = window.storeService.open(window.name);
+                        }
```
owennw: Given this is used about 3 times now perhaps it could be a function?

Across the app it's used quite a lot too.. we should probably do something about that.

WPFerg: Done

owennw:
Looks good, besides the 'out-of-date' message.

WPFerg:
Brought up to latest dev now

PR #481, opened 2016-03-01
Compact search exit

akhmorgan:
adds close icon to search in compact view to make a more obvious way for users to exit the compact search view.

### DIFF
--- a/src/sidebars/search/search.less
+++ b/src/sidebars/search/search.less
@@ -84,7 +84,7 @@
       #close-button {
         position: relative;
         float: right;
-        top: -17px;
+        top: -16px;
         right: 7px;
       }
     }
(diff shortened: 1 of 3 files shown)


src/sidebars/search/search.less
```diff
@@ -74,6 +78,17 @@
...
     display: block;
   }
+  .compact-search.expanded {
+    icon {
+      display: block;
+      #close-button {
+        position: relative;
+        float: right;
+        top: -16px;
+        right: 7px;
```
WPFerg: Maybe a `cursor: pointer;` so it gives a better indication it can be clicked? Looks good aside from that, though

akhmorgan: That would be inconsistent with the other icons, but I'm happy to do it.

akhmorgan: Actually, the other icons should probably have `cursor:pointer` too, really.

WPFerg: True, those icons might be a special case, though, due to their window 'control-ey' nature. I'll merge it in anyway, and it can be something for later if UX picks it up.

PR #484, opened 2016-03-02
Setting shadow on all windows.

### LINKED ISSUE #365: Add drop shadow back to windows

bjedrzejewski:
This closes #365


owennw:
Looks great! Okay to merge when all the checks are complete.

src/window-service.js
```diff
@@ -268,14 +268,16 @@
...
                 _window.updateOptions({
                     resizable: true,
                     minHeight: 510,
                     minWidth: 918,
-                    maximizable: true
+                    maximizable: true,
+                    shadow: true
                 });
             }
         }
```
owennw: I don't think you need to update the shadow options here, as they were never set to false.

bjedrzejewski:
@owennw fixed, ready to go imo. It looks really nice now!

src/window-service.js
```diff
@@ -268,7 +268,8 @@
                     resizable: false,
                     minHeight: 500,
                     minWidth: 230,
-                    maximizable: false
+                    maximizable: false,
+                    shadow: true
```
owennw: :( Sorry my fault here, wasn't clear in my previous comment. I don't think you need this one either. Given that it might be worth a squash too...

bjedrzejewski:
@owennw now it's good to go then :)

owennw:
Thanks for that, my bad!

Issue #486, opened 2016-03-02
Improve icon layout in compact view


owennw:

bjedrzejewski:
@owennw scope creep? :D

owennw:
Given there's a massive cross in the compact view I'd not say so :P

bjedrzejewski:
Oh, wow. That's a feature.

Issue #191, opened 2016-01-12
Refactor less to use 'design units' and style components with mixins


bjedrzejewski:

owennw:
#699

PR #488, opened 2016-03-02
Refactor compact LESS; introduce (first!) mixin (take 2)

owennw:


owennw:
#486

WPFerg:
Also part of #191's large refactor

src/assets/styles/mixins.less
```diff
@@ -0,0 +1,15 @@
+﻿.right-alignment(@icon-order) {
+  right: (15 + (@icon-order * 30));
+}
+
+.svg-position {
```
ColinEberhardt: The other mixing looks good, but not so sure about `svg-position`, what does it do exactly? You should name mixing in a similar way that you would name functions.

owennw: Would `position-svg` be better (that is what it does)? It specifies position and top for any svg (could be button but also includes the logo) in the toolbar.

WPFerg: `.toolbar-offset-top`?

src/assets/styles/style.less
```diff
@@ -39,23 +46,20 @@ body {
 
 #showcase-title {
   color: @chart-teal;
-  font-size: 20px;
   position: absolute;
   right: 75px;
   z-index: 10;
-  top: -12px;
+  top: -10px;
```
bjedrzejewski: Did you mean to actually change the CSS as well?

owennw: Yes; I removed the `font-size` here because it wasn't actually applying to any fonts (they're specified in the children), but was adding extra padding to the text that I needed to take into account.

bjedrzejewski:
@owennw is this good to merge now?

src/assets/styles/mixins.less
ColinEberhardt: I like your suggestion @WPFerg - it feels like a better description of what this actually does.

src/assets/styles/mixins.less
```diff
@@ -0,0 +1,44 @@
...
+  right: (15 + (@icon-order * 30));
+}
+
+.toolbar-offset-top {
+  position: fixed;
+  top: 15px;
+}
+
+// 400 is 'normal'
+.font(@size, @weight: 400, @line-height: 1.5) {
```
WPFerg: `@font-medium` is `500`, is it possible to use a variable as a default?

owennw: possibly, but it seemed the default throughout the app was indeed 400.

WPFerg: Fair enough, since it's keeping it consistent then I'm happy

PR #489, opened 2016-03-02
Use bitflux 1.2 release

WPFerg:


package.json
```diff
@@ -33,7 +33,7 @@
     "angular-animate": "^1.4.8",
     "angular-resource": "^1.4.8",
     "d3fc": "^5.2.0",
-    "BitFlux": "ScottLogic/BitFlux#9ddec65e495b7201e07fa9b7951f2f36b3448a52",
+    "BitFlux": "ScottLogic/BitFlux#2de16a8504fa981e9df094640495ae53ac576438",
```
owennw: Do they tag their releases? It would be nice to see a tag here, so we can tell more specifically what version is being used.

WPFerg: Done

PR #490, opened 2016-03-02
Added GPL Licence

### LINKED ISSUE #429: Determine a suitable licence
Previously we have defaulted to MIT - I'd like to check whether it is suitable in this case.

ColinEberhardt:
closes #429 

The GPL documentation [encourages the addition of the notice to every source file](http://www.gnu.org/licenses/gpl-faq.html#NoticeInSourceFile), but does not necessitate this. I'm not keen on adding repetitive boiler-plate to our code base, so we'll go with this more minimal version!

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,22 +1,619 @@
-The MIT License (MIT)
+GNU GENERAL PUBLIC LICENSE
+   Version 3, 29 June 2007
 
-Copyright (c) 2015 Nick Owen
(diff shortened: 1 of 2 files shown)


bjedrzejewski:
I take it, that we don't want to use LGPL to limit commercial and close-source re-use?

ColinEberhardt:
@bjedrzejewski - as discussed off line, I think GPL is good for now - we might split out parts in future and licence under MIT

PR #491, opened 2016-03-02
Shadow is back on closed window.

### LINKED ISSUE #485: There is no shadow from the recently closed popup div

bjedrzejewski:
The shadow is thrown to top left, rather than bottom right. Given that left is dark, this actually looks very nice to me.

Also closes #485


owennw:
The border looks better there, I think!

However, I can't see it in the compact view. At the moment I'm going through the LESS #488 so I'll do that while I'm refactoring.

Issue #419, opened 2016-02-25
Selectively apply pixel snapping


ColinEberhardt:
Some icons are a bit blurry

![image](https://cloud.githubusercontent.com/assets/1098110/13315596/858cfadc-dba3-11e5-8b65-2d7ffe12124a.png)

I think 'crisp edges' would resolve this.

WPFerg:
I can't seem to get `crispEdges` working in this instance. The candlestick is a background image, and I can't see a way to edit `shape-rendering` there. The only way around this that I can see is extract `candlestick-series.svg` from Bitflux, add in `shape-rendering='crispEdges'` and then override the `background-image`, which is fairly convoluted

ColinEberhardt:
I think we should also pixel align the logo

![image](https://cloud.githubusercontent.com/assets/1098110/13319177/c19f4912-dbb7-11e5-9321-a3c616cd72f8.png)

WPFerg:
The closest to pixel alignment I've gotten involves making the logo slightly smaller, but produces a better result. 
![image](https://cloud.githubusercontent.com/assets/7962948/13319565/7dc7bfd2-dbba-11e5-9ede-d8ed2ddb1f62.png)

ColinEberhardt:
Rather than SVGs, I think these would be better as PNGs that have been rendered to specific sizes, which we then render aligned to pixel boundaries. That way Rui can be in charge of getting the font rendering / aliasing to work nicely!

owennw:
Closed by #495

PR #495, opened 2016-03-02
Replace SVGs with PNG images

owennw:
Towards #419, don't know if this is sufficient to close it though.


WPFerg:
I think the ones done are good, but I think #419 also needs the Bitflux icons updated, too. I'm happy to merge this one in, though

bjedrzejewski:
@owennw we can't merge this yet. For example the start icon png is 15px, but we are forcing it to render as 16px. This brings back the sub-optimal rendering problem. Is it the css problem, or icon png should be larger? There are issues like that with most icons (magnifying glass and star on the top) etc.

owennw:
Working on it. I also think I've lost the closed tabs icon in compact view..

bjedrzejewski:
@owennw did Rui give you updated pngs?

Issue #497, opened 2016-03-03
UX Issues [Windows Menu]


WPFerg:
**Closed Windows Menu**
- [X] Increase text size of date to 12px,
- [X] Increase text size of stock symbols to 14px,
- [X] Add arrow pointing down towards icon.
- [X] Add a separator line above the icon (similar to the ones in-between cards), but reduce its top and bottom margins
- [x] Padding-bottom is not the same as top. Should have around 15 px for both.
- [ ] When opening a previously closed window, the top-right corner buttons did not appear, and the overall application was frozen. (*edit: I can't reproduce this*)

owennw:
Is point 5 still outstanding?

WPFerg:
Nope

owennw:
I'd suggest splitting point 6 into a different issue, not with the milestone, and closing this one.

Issue #494, opened 2016-03-02
UX Issues [Chart]


WPFerg:
*Chart*:
- [x] Purple on RSI doesn't match visual style (Purple colour should be #9D97E5)
- [x] Current price line and label should have the same blue teal colour as depicted in the designs (they have a different blue colour in the newest version.
- [x] Line weight in the summary chart is too think, perhaps 1px less?
- [x] Indicator “close” buttons are rounded, but should be squared, with a 1 px teal border.
- [x] The default timespan for the chart is too broad, making the chart too compressed. I suggest a default of 31 days.

*Summary Chart*
- [x] should be aligned with the separator line for the Closed Windows icon,
- [x] extending the main chart a bit further down
- [x] reducing spacing between the two charts to around 20px below the date labels

owennw:
Closed by #498

PR #498, opened 2016-03-03
Chart style updates

WPFerg:
Fixes 7/8 in #494. Indicator close buttons need a new SVG/png to implement, so I'm waiting on that.

Also fixes an issue in #497

### DIFF
--- a/src/showcase/showcase-changes.less
+++ b/src/showcase/showcase-changes.less
@@ -119,7 +119,7 @@
 
       path {
         stroke: @chart-fade !important;
-        stroke-width: 2px;
+        stroke-width: 1.5px;
       }
     }
 
--- a/src/showcase/showcase-changes.less
+++ b/src/showcase/showcase-changes.less
@@ -155,7 +155,7 @@
 }
 
 .y-axis {
-  .closeLine.tick {
+  .close-line.tick {
     > path {
       stroke: @chart-teal !important;
       fill: @chart-teal !important;
@@ -187,7 +187,7 @@
   stroke-dasharray: 8px;
 }
 
-.plot-area .annotation > line {
+.plot-area .annotation.close-line > line {
   stroke: @chart-teal;
 }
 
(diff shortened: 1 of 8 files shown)


src/showcase/showcase-changes.less
```diff
@@ -179,11 +179,15 @@
     }
 }
 
+.multi.rsi.indicator {
+    stroke: @tertiary-purple;
+}
```
owennw: I'm somewhat surprised not to see a companion change for this? Were we just always using the wrong colour?

owennw:
When you say waiting on that, have you requested that change from anyone?

WPFerg:
@owennw Rui

src/showcase/showcase-changes.less
WPFerg: I couldn't find any references to `@tertiary-purple`, and for some reason, I don't think Less is overriding it (or the selector's changed in Bitflux). I'll have a look

WPFerg: Okay, for some reason it reset to D3FC's default, and `@tertiary-purple` was applied, but for some reason the order of precedence changed and the showcase CSS is being overriden by D3FC CSS.
Even though on Bitflux it works in the correct order of precedence.

owennw: But the styles are correct after this PR?

WPFerg: Yeah, I think it's just this instance where it happens

src/sidebars/favourites/minichart/minichart.less
```diff
@@ -2,7 +2,7 @@
   position: absolute;
   padding-top: 10px;
   width: 100%;
-  overflow: hidden;
+  overflow: visible;
```
owennw: What's the default for overflow?

WPFerg: I believe `visible`, but in this instance it might be best to be explicit

PR #500, opened 2016-03-03
Use adjusted values for OHLC data

WPFerg:


bjedrzejewski:
I am happy to merge that.

PR #504, opened 2016-03-03
Star refactor

akhmorgan:


src/sidebars/star/star-controller.js
```diff
@@ -15,33 +15,35 @@
             this.selectionService = selectionService;
 
             this.starHovered = false;
-            this.check = false;
+
+            this.stock = $scope.stock;
```
owennw: It looks like the `$scope` variable was a hangover from before... perhaps it doesn't need to be assigned to `this.$scope` above? (I know that's not strictly part of this refactor though.)

owennw:
Overall much better. Might be worth a squash, perhaps, given the console.logs around etc? But not too concerned.

src/sidebars/star/star-controller.js
akhmorgan: I'm not certain I understand what you mean here. Are you saying that because `$scope` is only used in the contructor, it doesn't need to be assigned to `this.$scope`?

owennw: Yes.

Issue #506, opened 2016-03-04
Application title is still 'OpenFin D3FC'


ColinEberhardt:
If you mouse-over the app in the task bar it still has the old project name

PR #509, opened 2016-03-04
Rename app

owennw:
I'll also rename the angular modules here too, soon.


WPFerg:
#506

akhmorgan:
looks good to me.

Issue #508, opened 2016-03-04
Sometimes windows open in an un-responsive state


ColinEberhardt:
I've had this happen a couple of times now. In this case i re-opened a previously closed window. As you can see it doesn't have a title bar and the entire window is un-responsive.

![image](https://cloud.githubusercontent.com/assets/1098110/13523213/fc7fe9c6-e1eb-11e5-9670-f0bb73c0ed0c.png)

You cannot interact with it in any way

WPFerg:
I think the stack trace is as follows:

```
TypeError: Cannot set property '_owner' of undefined
    at addDesktopEventCallback (<anonymous>:1253:25)
    at Object.fin.desktop.Window.addEventListener (<anonymous>:3379:9)
    at ToolbarCtrl.onReady (app.js:2113)
    at app.js:2083
    at whenReady (<anonymous>:2543:9)
    at Object.fin.desktop.main (<anonymous>:2601:5)
    at Object.ready (app.js:453)
    at new ToolbarCtrl (app.js:2081)
    at Object.invoke (angular.min.js:41)
    at T.instance (angular.min.js:86)
```

WPFerg:
@ColinEberhardt Can you retry off of the latest version and see if it has been fixed for you?

WPFerg:
I'm happy to close this now since we've released a few new versions and had it tested a bit more with it not being discovered.

ColinEberhardt:
Thanks @WPFerg - I can also confirm that it hasn't re-surfaced for a while.

PR #518, opened 2016-03-04
Fix non-responsive state

WPFerg:
#508 
`this.maximisedEvent` etc were bound after the `ready` event was bound, meaning that the listener callbacks weren't defined.


WPFerg:
Closing as git seemed to do something there

PR #519, opened 2016-03-04
Fix non-responsive state (Take 2)

WPFerg:
#508 
this.maximisedEvent etc were bound after the ready event was bound, meaning that the listener callbacks weren't defined if OpenFin was already ready.


owennw:
So was this previously a race condition? I've never seen this bug, which makes me curious.

WPFerg:
Yes, and it was only specific circumstances

PR #531, opened 2016-03-04
Confirmation window

akhmorgan:


WPFerg:
A few things from using (probably good to run them past Rui)
* The cancel button doesn't look too great (Perhaps have the OK button's style on cancel and have OK with inverted BG/text?)
* Should there be behaviour on pressing Enter/Tab?
* Should the confirm box go away when clicking anywhere on the box that isn't OK/Cancel/X?

src/sidebars/star/star.html
```diff
@@ -1,2 +1,20 @@
 ﻿<img class="star" ng-src="assets/png/{{starCtrl.favouriteUrl()}}.png" ng-click="starCtrl.click(); $event.stopPropagation();"
      ng-mouseenter="starCtrl.mouseEnter()" ng-mouseleave="starCtrl.mouseLeave()" draggable="false" />
+
+<div class="confirmation-backdrop " ng-show="starCtrl.confirmationShow" ng-click="starCtrl.hideModal()">
```
WPFerg: Maybe swap out `ng-show` for `ng-if` so all the confirm boxes aren't lingering in the DOM.

akhmorgan:
@wpferg I agree. This all needs to be run past Rui. This is just the initial implementation to get the ball rolling.

src/sidebars/star/star.html
akhmorgan: Good point. I'll try that.

owennw:
Something in here affects the styling of the search tab in main view. The search results appear too high up and interfere with the search input box.

src/sidebars/star/star.less
```diff
@@ -1,3 +1,49 @@
...
+    margin: 100px auto;
+    .modal-content {
+      .header{
+        padding: 15px;
+        border-bottom: 1px solid #e5e5e5;
+        .title {
+          padding-top: 1px;
+          font-size: 20px;
+          font-weight: 500;
+          line-height: 1;
```
owennw: There's a mixin `.font` for the size, weight and line height.

src/sidebars/star/star.less
```diff
@@ -1,3 +1,49 @@
...
+
+.confirmation-backdrop {
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  z-index: 1050;
+  display: none;
+  overflow: hidden;
```
owennw: I think there's a mixin for overflow too...

src/sidebars/star/star-controller.js
```diff
@@ -47,6 +52,19 @@
...
         }
 
+        deselect() {
+            if (!this.store) {
+                this.store = window.storeService.open(window.name);
+            }
+            this.stock.favourite = false;
+            this.store.remove(this.stock);
+            this.hideModal();
+        }
```
owennw: You can probably call `deselect` from `click` too?

src/sidebars/star/star-controller.js
```diff
@@ -32,7 +34,10 @@
         }
 
         click() {
-            if (!this.check || confirm('Are you sure you wish to remove this stock (' + this.stock.code + ') from your favourites?')) {
+            if (this.check) {
+                this.confirmationShow = true;
+            }
+            if (!this.check) {
```
owennw: This can be an `else`.

src/sidebars/star/star-controller.js
akhmorgan: Done.

src/sidebars/star/star.less
akhmorgan: Done.

src/sidebars/star/star.less
akhmorgan: And done.

src/sidebars/star/star.html
```diff
@@ -1,2 +1,20 @@
 ﻿<img class="star" ng-src="assets/png/{{starCtrl.favouriteUrl()}}.png" ng-click="starCtrl.click(); $event.stopPropagation();"
      ng-mouseenter="starCtrl.mouseEnter()" ng-mouseleave="starCtrl.mouseLeave()" draggable="false" />
+
+<div class="confirmation-backdrop " ng-if="starCtrl.confirmationShow" ng-click="starCtrl.hideModal()">
+  <div class="confirmation-modal">
```
WPFerg: Clicking on the modal itself closes it. Would a `$event.stopPropagation()` here fix this?

src/sidebars/star/star.html
```diff
@@ -1,2 +1,20 @@
...
+      <div class="header">
+        <span class="title">StockFlux</span>
+        <icon name="close" icon-click="starCtrl.hideModal()"></icon>
+      </div>
+      <div class="message">
+        <p ng-bind="starCtrl.confirmationMessage"></p>
+      </div>
+      <div class="buttons">
+        <button class="btn btn-primary ng-binding" ng-click="starCtrl.deselect(); $event.stopPropagation();">OK</button>
+        <button class="btn btn-default ng-binding" ng-click="starCtrl.hideModal(); $event.stopPropagation();">Cancel</button>
```
owennw: I thought `btn` and `btn-primary` etc were bootstrap classes, but we're not using bootstrap. Are they doing something here?

bjedrzejewski:
Clicking on the window itself shouldn't close it.

bjedrzejewski:
@akhmorgan @owennw do you think it would be better to always show the dialog in the middle of the window?

src/sidebars/star/star.html
bjedrzejewski: I think we are inheriting these bootstrap customized classes from d3fc/bootstrap. It is not such a bad idea I think! We are going to be the same as the showcase.

akhmorgan:
@bjedrzejewski I think that sort of thing is more a question for UX, which can then be followed up in another PR. This PR is just to get the core functionality in.

bjedrzejewski:
@akhmorgan Ok, I would say then- I am happy to merge if one more person is happy to merge. @owennw - looking at you

owennw:
The spec states that the dialog should appear close to the star icon, with the bubble like the closed window list.

Issue #523, opened 2016-03-04
When you click and drag the mouse through search panel stars disappear


bjedrzejewski:
Click, hold, don't release, drag to the next search result, see that:

![image](https://cloud.githubusercontent.com/assets/9947570/13525604/ee991ed2-e1f9-11e5-9fd5-59517eaa7045.png)

PR #544, opened 2016-03-04
Fix disappearing star on dragover

WPFerg:
#523


owennw:
Good to merge when ready.

PR #553, opened 2016-03-07
Confirmation window

owennw:
To replace #531


bjedrzejewski:
@owennw @akhmorgan I think that Adam should do the squash, so he gets his name on that work :)

PR #557, opened 2016-03-07
Only shunt window on top edge

WPFerg:


owennw:
What if there's a large bottom taskbar?

WPFerg:
Similar to Chrome, the movable region is occluded.

bjedrzejewski:
@WPFerg I think we should merge it once you get that in line with dev.

WPFerg:
Done

PR #561, opened 2016-03-08
Add search loading message

### LINKED ISSUE #222: There should be friendly 'searching' message
Sometimes searching can take a few second (slow network, many results). A message saying something like 'searching' would be very useful.

WPFerg:
closes #222

### DIFF
--- a/src/sidebars/search/search.html
+++ b/src/sidebars/search/search.html
@@ -7,6 +7,9 @@
   </div>
   <div id="search-scroll" class="side-scroll custom-scrollbar">
       <div class="sidetab" ng-show="sidebarCtrl.showSearches()">
+          <div ng-show="searchCtrl.isLoading && !searchCtrl.noResults" class="loading-message results-message">
+              Loading search results...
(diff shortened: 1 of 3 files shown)


bjedrzejewski:
Reviewing :)

src/sidebars/search/search.html
```diff
@@ -18,7 +21,7 @@
                   </div>
               </div>
           </div>
-          <div class="no-results" ng-show="searchCtrl.noResults">
+          <div class="results-message no-results" ng-show="searchCtrl.noResults">
```
owennw: Might be a bit late here, but is the `no-results` class still being used?

WPFerg: No, but I wanted to keep it so there's at least some differentiation between the two messages.

owennw: Ah I see: `loading-message` is also a description?

WPFerg: Correct

PR #567, opened 2016-03-08
Scrollbar height fix. Closes #564

### LINKED ISSUE #564: When scroll-bar appears not all favorites are shown
This is a regression issue.

image

bjedrzejewski:
Closes #564

### DIFF
--- a/src/assets/styles/custom-scroll.less
+++ b/src/assets/styles/custom-scroll.less
@@ -13,7 +13,9 @@
   }
 
   &#favourite-scroll {
-    bottom: 40px;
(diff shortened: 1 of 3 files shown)


owennw:
Looks fine on the favourites tab, but on the main view's search tab I can't get to the bottom of those results. (seems like I can in compact view though!)

bjedrzejewski:
@owennw I didnt know there is also a bug in the search view! Let me add a commit to fix that as well.

bjedrzejewski:
@owennw updated, should work on both now. Thanks for spotting this another issue!

Issue #326, opened 2016-02-11
Brief display of "no favourites" message when app opens


owennw:
If possible I'd like to delay the visual rendering until angular has finished intialising? Or something to that effect.

WPFerg:
This is because although the favourites are added by Angular, it takes time to fetch the data from Quandl. The easiest way to wait is to count the responses from Quandl, until `responses === stocks`, then show

WPFerg:
An easy-ish way to do this would be to replace "no stocks" with a loading message. I think the additional wait for Quandl to return before showing the window would make the app feel more unresponsive.

ColinEberhardt:
I'm not sure whether to open a new issue for this - but when you tear out a tile, you also see the effect of the application 'bootstrapping'.

Perhaps you should make have a class that is part of the HTML template itself that hides the application, then remove this class once Angular has bootstrapped?

WPFerg:
@ColinEberhardt Do you mean have a big, empty window whilst the application is bootstrapping? Also, in tear out cases, angular will have already bootstrapped, due to windows being pooled.

ColinEberhardt:
I'm not sure what's best, a big empty window, or a loading indicator - either way it should be something that looks like it is 'by design', rather than the current behaviour which is just the way it worked 'out of the box' ;-)

WPFerg:
There's a PR in for some loading text, which could work as an interim if we wanted to add a loading spinner.

PR #568, opened 2016-03-08
Show loading text in favourites list

WPFerg:
The solution I have in mind for #326 to give a decent middle ground between responsiveness and information shown.

### DIFF
--- a/src/sidebars/favourites/favourite-controller.js
+++ b/src/sidebars/favourites/favourite-controller.js
@@ -15,6 +15,7 @@
             this.$timeout = $timeout;
 
             this.store = null;
+            this.receivedFirstQuandlResponse = false;
             this.stocks = [];
             this.update();
             this._watch();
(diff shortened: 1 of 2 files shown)


bjedrzejewski:
Checking it out now

src/sidebars/favourites/favourite-controller.js
```diff
@@ -66,6 +67,12 @@
 
                 this.favourites = this.store.get();
 
+                if (this.favourites.length === 0) {
+                    this.$timeout(() => {
+                        this.receivedFirstQuandlResponse = true;
+                    });
+                }
+
```
bjedrzejewski: What is this code for? Doesn't length 0 mean no response possibly?

WPFerg: In case there's a window with no favourites in storage. This is possible, but probably another issue's scope.

bjedrzejewski:
@WPFerg can you merge the pull request and then we can put it on dev?

WPFerg:
@bjedrzejewski I've updated it to be level with dev, just CI to go :)

PR #569, opened 2016-03-08
Corrected the style for search scrollbar

bjedrzejewski:


owennw:
Hm I didn't notice this. Are these numbers random or coupled to something else? Can we start using variable/mixins to make this nicer and prevent future problems?

bjedrzejewski:
At the moment this is to keep it in line with the design, this number is not present anywhere else other than fav tab (but this is coincidence possibly).

bjedrzejewski:
I was thinking about using- @toolbar-height, but I am not sure if we have a guarantee that this is conceptually linked.

owennw:
So it's not a calculated number based on other heights or something like that?

bjedrzejewski:
It can be made into calculated number by refactoring the way the other stuff (search bar) is written. This is a separate refactoring though.

bjedrzejewski:
@owennw changed it. Once travis is happy, can merge.

Issue #516, opened 2016-03-04
When toggling compact view there is quite a bit of UI 'jank'


ColinEberhardt:
Needs smoothing ....

owennw:
#387, #326

WPFerg:
I believe a decent fix for this would be removing the `ng-if` on the Showcase-Controller.

akhmorgan:
I think it's going to be a little more complicated than that. My first thought is to add a flag to say when the window is transitioning between compact and full, then disable all css animations and transitions whilst that flag is present.

WPFerg:
Closed by #571

PR #571, opened 2016-03-08
Remove animations on compact change

WPFerg:
Closes (and uses the approach in) #516

### DIFF
--- a/src/main/main-directive.js
+++ b/src/main/main-directive.js
@@ -2,12 +2,21 @@
     'use strict';
 
     angular.module('stockflux.main')
-        .directive('main', [() => {
+        .directive('main', ['$timeout', ($timeout) => {
             return {
                 restrict: 'E',
                 templateUrl: 'main/main.html',
                 controller: 'MainCtrl',
-                controllerAs: 'mainCtrl'
+                controllerAs: 'mainCtrl',
+                link: (scope, element) => {
+                    scope.$on('compactChanging', () => {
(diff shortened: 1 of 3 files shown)


src/main/main-directive.js
```diff
@@ -2,12 +2,21 @@
...
                 controller: 'MainCtrl',
-                controllerAs: 'mainCtrl'
+                controllerAs: 'mainCtrl',
+                link: (scope, element) => {
+                    scope.$on('compactChanging', () => {
+                        element.addClass('disable-transitions');
+
+                        $timeout(() => {
+                            element.removeClass('disable-transitions');
+                        }, 100);
```
owennw: I wonder if there should be two events: `before-compact-change` and `after-compact-change`?

That way you might not need the `$timeout`, and you can broadcast on either side of the compact change in the toolbar-controller. Or do async things mess that up?

WPFerg: I've just tried it, and it doesn't work. Probably something to do with how angular's events are processed.

bjedrzejewski:
@WPFerg can you update? I will test it and if it works great, will merge

WPFerg:
@bjedrzejewski Done

bjedrzejewski:
@WPFerg ups - travis cried

WPFerg:
@bjedrzejewski Looks like it fell at `npm install`, I did a `grunt ci` and it passed, give it a go anyway and we'll try again if you think it's ready to merge.

PR #578, opened 2016-03-09
Tearin indicator

### LINKED ISSUE #357: Create visual indication of tear in

WPFerg:
closes #357.

### DIFF
--- a/src/sidebars/favourites/favourites.less
+++ b/src/sidebars/favourites/favourites.less
@@ -5,12 +5,17 @@
 @favourite-padding-bottom: 10px;
 @top-height: 30px;
 @fav-split-color: #323A48;
+@highlight-colour: #447272;
 
 .favourites {
   float: right;
   height: 100%;
   background-color: @favourites-background;
   transition: @transition;
+
+  &.active {
+    background-color: #447272;
+  }
 }
 
 .hover-area {
--- a/src/sidebars/favourites/favourites.less
+++ b/src/sidebars/favourites/favourites.less
@@ -14,7 +14,11 @@
   transition: @transition;
 
   &.active {
-    background-color: #447272;
+    background-color: @highlight-colour;
+
+    .favourite, .drop-target, .hover-area {
+      background-color: transparent;
+    }
   }
 }
 
--- a/src/sidebars/favourites/geometry-service.js
+++ b/src/sidebars/favourites/geometry-service.js
@@ -77,12 +77,6 @@
     }
 
     class GeometryService {
-        windowsIntersect(openFinWindow, _window) {
-            var nativeWindow = openFinWindow.getNativeWindow();
-
-            return intersectHelper(getWindowPosition(nativeWindow), getWindowPosition(_window));
-        }
-
         elementIntersect(openFinWindow, _window, element) {
             var nativeWindow = openFinWindow.getNativeWindow();
 
--- a/src/sidebars/favourites/tearout-directive.js
+++ b/src/sidebars/favourites/tearout-directive.js
@@ -113,7 +113,7 @@
                                     dragService.overAnotherInstance((overAnotherInstance) => {
                                         if (overAnotherInstance) {
                                             dragService.moveToOtherInstance(scope.stock);
-                                            dragService.setOtherInstance(null);
+                                            dragService.clearOtherInstance();
                                             store.remove(scope.stock);
                                             dragService = null;
                                         } else {
@@ -163,7 +163,7 @@
                         function boundsChangingEvent() {
                             if (insideFavouritesPane()) {
                                 reorderFavourites();
-                                dragService.setOtherInstance(null);
+                                dragService.clearOtherInstance();
                             } else {
                                 // Check intersections to set the tear in indicator states.
                                 dragService.overAnotherInstance();
--- a/src/sidebars/favourites/tearout-directive.js
+++ b/src/sidebars/favourites/tearout-directive.js
@@ -113,7 +113,7 @@
                                     dragService.overAnotherInstance('.favourites', (overAnotherInstance) => {
                                         if (overAnotherInstance) {
                                             dragService.moveToOtherInstance(scope.stock);
-                                            dragService.clearOtherInstance();
+                                            dragService.destroy();
                                             store.remove(scope.stock);
                                             dragService = null;
                                         } else {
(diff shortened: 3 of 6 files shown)


src/sidebars/favourites/favourites.less
```diff
@@ -5,12 +5,17 @@
...
+@highlight-colour: #447272;
 
 .favourites {
   float: right;
   height: 100%;
   background-color: @favourites-background;
   transition: @transition;
+
+  &.active {
+    background-color: #447272;
```
bjedrzejewski: Did you not just defined: `@highlight-colour` ?

WPFerg: Oversight of the year, I'll fix that ;)

bjedrzejewski: Yeah I though so :)

src/sidebars/favourites/tearout-directive.js
```diff
@@ -160,6 +161,10 @@
                         function boundsChangingEvent() {
                             if (insideFavouritesPane()) {
                                 reorderFavourites();
+                                dragService.setOtherInstance(null);
```
owennw: Outside of the drag service all the calls to this method pass `null`. Further to that I don't think anything outside of the class should be able to specify another instance. I think I'd prefer a `clearOtherInstance` to start with, but I'll have a think of whether I prefer another pattern.

WPFerg: Yeah, I know what you mean, I'll add a shortcut function. To be honest, I think there's only one time it needs to be called. I'll try reducing the number of times I call it (so it's just before the service is destroyed)

WPFerg: @owennw I've attempted a further abstraction on the inner state by adding a `destroy` function and adding a new function for `overThisInstance`.

src/sidebars/favourites/tearout-directive.js
```diff
@@ -108,9 +108,10 @@
                                         store = window.storeService.open(window.name);
                                     }
 
-                                    dragService.overAnotherInstance((overAnotherInstance) => {
+                                    dragService.overAnotherInstance('.favourites', (overAnotherInstance) => {
                                         if (overAnotherInstance) {
                                             dragService.moveToOtherInstance(scope.stock);
+                                            dragService.destroy();
                                             store.remove(scope.stock);
                                             dragService = null;
```
owennw: It feels weird having both a `dragService.destroy()` and a `dragService = null` like that. I feel like this API is homing in, just not quite sure what it should be...

src/window-service.js
```diff
@@ -112,8 +123,11 @@
                 var deferred = this.$q.defer();
                 promises.push(deferred.promise);
                 mainWindow.getState((state) => {
-                    if (state !== 'minimized' && this.geometryService.windowsIntersect(this.tearoutWindow, mainWindow.getNativeWindow())) {
-                        this.otherInstance = mainWindow;
+                    var nativeWindow = mainWindow.getNativeWindow();
+                    var element = nativeWindow.document.querySelector(selector || 'body');
+
+                    if (!result && state !== 'minimized' && this.geometryService.elementIntersect(this.tearoutWindow, nativeWindow, element)) {
```
owennw: I reckon with this change the `windowsIntersect` method is never called now.

src/sidebars/favourites/tearout-directive.js
WPFerg: I don't think we need a `dragService = null;`, unless there was a reason for it when you wrote it? A long time ago, I know, so I'm not sure if you'll be ableo to remember

src/sidebars/favourites/tearout-directive.js
```diff
@@ -158,8 +159,11 @@
                         }
 
                         function boundsChangingEvent() {
-                            if (insideFavouritesPane()) {
```
owennw: Is `insideFavouritesPane` needed now?

src/window-service.js
WPFerg: Correct. Will remove.

src/sidebars/favourites/tearout-directive.js
WPFerg: It's used in one other place, I'll update that reference to the newer one

src/sidebars/favourites/tearout-directive.js
owennw: I assume it was for disposing purposes? As the directive is about to destroy itself.. guess it's not needed.

src/sidebars/favourites/geometry-service.js
```diff
@@ -77,12 +77,6 @@
     }
 
     class GeometryService {
-        windowsIntersect(openFinWindow, _window) {
-            var nativeWindow = openFinWindow.getNativeWindow();
-
-            return intersectHelper(getWindowPosition(nativeWindow), getWindowPosition(_window));
-        }
```
owennw: Similarly does this snowball? Are `intersectHelper` and `getWindowPosition` still used?

src/sidebars/favourites/tearout-directive.js
```diff
@@ -157,12 +157,11 @@
...
-                            // Check if you are over a drop target by seeing if the tearout rectangle intersects the drop target
-                            insideFavouritesPane = geometryService.elementIntersect(
-                                tearoutWindow, window, document.getElementsByClassName('favourites')[0]);
-
-                            if (insideFavouritesPane) {
+                            if (dragService.overThisInstance(TEAR_IN_SELECTOR)) {
                                 reorderFavourites();
+                            } else {
+                                // Check intersections to set the tear in indicator states.
+                                dragService.overAnotherInstance(TEAR_IN_SELECTOR);
```
owennw: Similarly here it feels wrong to call a method that returns a bool, but only for the purpose of changing state in the drag service...

src/sidebars/favourites/geometry-service.js
WPFerg: They are, each only in one place (`elementIntersect`). However, I think it'll be good to keep them separate because of the separation of concerns.

owennw: Yeah sounds good. Geometry service is left oddly thin now.

src/sidebars/favourites/tearout-directive.js
WPFerg: I think I've got a solution to this, to hopefully make it all contained in the drag service. I'll update in due course

WPFerg: I've brought out the main logic I was using `overAnotherInstance` for into its own function to do the update, and updated `overAnotherInstance` to use it (so hopefully this function would provide a foundation for other-instance checks if needed).

PR #581, opened 2016-03-09
Remove the bitflux tearout judder

### LINKED ISSUE #528: Moving a tearout card over Bitflux causes judder
Moving the tearout card around the place is fine and smooth, but if it goes over the Bitflux chart, it starts juddering a bit.

Edit: Experimentation -- Seems to be related to number of days loaded on the chart

WPFerg:
Closes #528

### DIFF
--- a/src/main/main-controller.js
+++ b/src/main/main-controller.js
@@ -2,8 +2,30 @@
     'use strict';
 
     class MainCtrl {
-        constructor() {
+        constructor($scope, $timeout) {
             this.store = null;
+            this.$scope = $scope;
+            this.$timeout = $timeout;
+
+            this.tearingOut = false;
+            this.watch();
+        }
+
+        watch() {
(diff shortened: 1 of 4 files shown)


bjedrzejewski:
@WPFerg can you make it merge-able?

WPFerg:
@bjedrzejewski done :)

src/main/main-controller.js
```diff
@@ -2,8 +2,30 @@
...
+
+            this.tearingOut = false;
+            this.watch();
+        }
+
+        watch() {
+            this.$scope.$on('tearoutStart', () => {
+                this.$timeout(() => {
+                    this.tearingOut = true;
+                });
```
owennw: Why do you need the `$timeout`s?

WPFerg: It doesn't apply the class otherwise

PR #582, opened 2016-03-09
Change closed windows icon when closed a window

owennw:
This PR introduces the feature whereby all closed windows icons are toggled to active when any app window is closed.

When one of these icons is then opened to view the new closed window it will revert back to the default behaviour. However, the closed window icons in any and all other app instances will remain active -- this is left as a todo for a future PR.

### DIFF
--- a/src/closedWindows/closedWindowList/closedWindowList-controller.js
+++ b/src/closedWindows/closedWindowList/closedWindowList-controller.js
@@ -7,8 +7,13 @@
             this.$timeout = $timeout;
             this.closedWindows = [];
             this.closedTabsShow = false;
+            this.overriddenIcon = '';
 
             this._watch();
+        }
+
+        override() {
+            return this.overriddenIcon;
         }
 
         refreshClosedWindows() {
@@ -16,6 +21,7 @@
         }
 
         click() {
+            this.overriddenIcon = '';
             this.refreshClosedWindows();
             this.closedTabsShow = this.closedWindows.length > 0;
         }
@@ -38,6 +44,10 @@
 
             this.$scope.$on('$destroy', () => {
                 window.windowService.removeClosedWindowListener(listener);
+            });
+
+            this.$scope.$on('windowClosed', () => {
+                this.overriddenIcon = this.$scope.icon + '_active';
             });
         }
     }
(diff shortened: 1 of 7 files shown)


owennw:
Note: this is slightly broken in compact view!

WPFerg:
I'm not 100% sure on the idea of an `override` for the icon directive, as it sort of removes the self-containedness of the directive (where it sets the url on a predetermined path according to the name). That said, I can't really think of another option aside from changing `<icon>` to an `` in this case, and I'm not sure thats 'better'.

owennw:
Yeah I was concerned that `icon` has outlived its usefulness, especially now we don't have hover states. But at the same time.. I think it's saving work. Better ideas are v welcome.

src/services/currentWindow-service.js
```diff
@@ -25,6 +25,10 @@
                 $rootScope.$broadcast('updateFavourites', event.stock);
             });
 
+            window.addEventListener('windowClosed', () => {
+                $rootScope.$broadcast('windowClosed');
+            });
```
WPFerg: This sort of thing is getting quite common place. Not in the scope of this PR but I think it's worth considering bringing the `window`/`$rootScope` event rebinding into the `currentWindowService` for a future release.

src/closedWindows/closedWindowList/closedWindowList-controller.js
```diff
@@ -39,6 +45,10 @@
             this.$scope.$on('$destroy', () => {
                 window.windowService.removeClosedWindowListener(listener);
             });
+
+            this.$scope.$on('windowClosed', () => {
+                this.overriddenIcon = this.$scope.icon + '_active';
+            });
```
WPFerg: There's two window close listeners here, do we need both?

owennw: I'm not sure if I see what you mean -- I see only one listener for `windowClosed`..

WPFerg: Lines 30 to 43 are adding a window close listener from previously merged PR.

owennw: While it's not tidy and needs a redesign, those listeners are listening for a `closedWindowChange` event, the new one is waiting for `windowClosed`

WPFerg:
Closing this in favour of another PR based upon this one.

PR #589, opened 2016-03-10
Confirmation style

akhmorgan:


src/sidebars/scrollbar-directive.js
```diff
@@ -7,6 +7,12 @@
                 restrict: 'C',
                 link: (scope, element) => {
                     var scrollPadding = 'scroll-padding';
+                    scope.$on('disableScrolling', function(event) {
+                        element.mCustomScrollbar('disable');
+                    });
+                    scope.$on('enableScrolling', function(event) {
+                        element.mCustomScrollbar('update');
+                    });
```
owennw: Doesn't look like you need either of the `event` variables here. Would prefer to see `() => ` too :)

src/sidebars/star/star.html
```diff
@@ -1,18 +1,14 @@
...
-        <span class="title">StockFlux</span>
-        <icon name="close" icon-click="starCtrl.hideModal()"></icon>
-      </div>
+<div class="confirmation-backdrop" ng-if="starCtrl.confirmationShow" ng-click="starCtrl.hideModal()">
+  <div class="bubble-head" ng-style="{top:starCtrl.modalBubbleTop()+'px'}"></div>
+  <div class="confirmation-modal" ng-class="{flipped: starCtrl.modalFlip()}"  ng-click="$event.stopPropagation();" ng-style="{top:starCtrl.modalTop()+'px'}">
+    <div class=confirmation-content>
       <div class="message">
-        <p ng-bind="starCtrl.confirmationMessage"></p>
+        <span>Are you sure you wish to remove this stock from your favourites?</span>
```
owennw: We no longer put the stock code in the confirmation message.. but given the location of the window I guess that's no longer necessary anyway.

src/sidebars/star/star.less
```diff
@@ -15,34 +19,40 @@
   outline: 0;
   display: block;
   .confirmation-modal {
+    &.flipped{
+      box-shadow: -2px -2px 2px 0px @dark-background;
+    }
+    box-shadow: -2px 2px 2px 0px @dark-background;
```
owennw: I'd quite like to see a `box-shadow(@shadow-position, @blur, @colour)` mixin, which could be used from the search.less and mixins.less files too.

src/sidebars/star/star.html
```diff
@@ -1,18 +1,14 @@
...
       </div>
       <div class="buttons">
-        <button class="btn btn-primary ng-binding" ng-click="starCtrl.deselect(); $event.stopPropagation();">OK</button>
-        <button class="btn btn-default ng-binding" ng-click="starCtrl.hideModal(); $event.stopPropagation();">Cancel</button>
+        <button class="btn btn-primary ng-binding" ng-click="starCtrl.deselect(); $event.stopPropagation();">Confirm</button>
```
owennw: I'm still a bit wary of using these bootstrap classes when our application doesn't explicitly depend on it, and it's more chance that BitFlux uses it.

src/sidebars/star/star.html
akhmorgan: Yeah, this is based off the spec.

owennw: Yeah I think it was only added in the first place as a stop gap.

src/sidebars/star/star.html
owennw: Reading this again I see it isn't a very constructive comment.. mostly because I don't know what to say about it.

src/sidebars/star/star.less
akhmorgan: Is that really massively better? So this would become something like `.shadowMixin(@shadow-position:(-2px,-2px), @blur: (2px, 0px), @colour: @dark-background)` or `.shadowMixin((-2px,-2px), (2px, 0px), @dark-background)`

owennw: Looking at it the blur might always be the same, so it would just be:

`.box-shadow(-2px -2px, @dark-background)`

src/sidebars/star/star.html
akhmorgan: I see your point here. However it's not really an issue with this PR, considering what's already in dev. I could take it out, but then I would end up more or less just be copying the bootstrap styles into our less.

src/sidebars/star/star.less
owennw: And it also takes us closer towards the goal of only using mixins for styling (#191)

src/sidebars/star/star.html
owennw: Yeah I agree, it's nothing to do with this PR so it's fine to leave as is.

PR #590, opened 2016-03-10
Remove upwards search jank transition + tidy LESS

owennw:


WPFerg:
The upwards search goofiness has gone, but the search results are lower down than they look like they should be in compact view. Do you want to add that in to this PR or have a different one?

owennw:
Ah thanks hadn't noticed that, let's put it in this one.

WPFerg:
LGTM, feel free to merge when Travis fires

PR #597, opened 2016-03-11
Closed window icon

### LINKED ISSUE #406: Change recover closed windows icon to blue when there is a new unseen closed window.
When the new card has been viewed the icon should revert to default. This will be across all open windows.

WPFerg:
Replacement for #582; (hopefully) closes #406 
Adds an additional 'seen' flag onto the windowService, which is updated on the children when they're informed there's a different closed window state.

### DIFF
--- a/src/closedWindows/closedWindowList/closedWindowList-controller.js
+++ b/src/closedWindows/closedWindowList/closedWindowList-controller.js
@@ -18,8 +18,16 @@
 
         refreshClosedWindows() {
             this.closedWindows = window.storeService.getPreviousClosedWindows();
+            this.updateSeen();
+        }
+
+        updateSeen() {
             var seen = window.windowService.getClosedWindowSeen();
             this.overriddenIcon = seen ? '' : (this.$scope.icon + '_active');
+
+            if (!seen && this.closedTabsShow && document.hasFocus()) {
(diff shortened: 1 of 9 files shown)


bjedrzejewski:
@WPFerg can you make it up to date? I will test and merge then.

WPFerg:
@bjedrzejewski Done

bjedrzejewski:
@WPFerg last small question. If the last opened are already open and you close a window. Would you still want the color to change (this happens)? I think it would be better if it did not change if the list is open. This should be really simple change if you agree that its a good idea. If you don't we can just merge now.

WPFerg:
That's a good point. I think it's probably worth doing, but only if the window has focus (if the list is open on a background window, then it'd clear instantly without you knowing). I'll have a try with that, and get back to you.

WPFerg:
@bjedrzejewski I've pushed up another commit that does this, and also clears if the list is open and you alt-tab and give the window focus.

bjedrzejewski:
@WPFerg if all goes well in the test I will merge.

WPFerg:
@bjedrzejewski I'm guessing you meant to merge this ;)

bjedrzejewski:
@WPFerg im blind sometimes :)

PR #601, opened 2016-03-11
Tearout jank removed, closes #560

### LINKED ISSUE #560: When tearing out there is a quite a bit of window movement 'jank'
When I tear out a tile I can see the window reposition itself a few times. Would be good to get this buttery smooth!

bjedrzejewski:
Closes #560

### DIFF
--- a/src/window-service.js
+++ b/src/window-service.js
@@ -214,7 +214,7 @@
      * Class that creates and governs OpenFin windows.
      */
     class WindowCreationService {
-        constructor($rootScope, storeService, geometryService, $q, configService) {
+        constructor($rootScope, storeService, geometryService, $q, configService, $timeout) {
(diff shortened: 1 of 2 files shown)


src/window-service.js
```diff
@@ -242,15 +243,23 @@
 
                 this.windowTracker.add(newWindow);
 
+                var self = this;
+                var showFunction = function() {
+                    self.$timeout(() => {
+                        newWindow.show();
+                        newWindow.bringToFront();
+                    });
```
WPFerg: Normally `$timeout`s are used to make angular update. What does it do here, does it break when it's removed?

bjedrzejewski: Timeout is the magic that makes it work. I will change to the lambda, thanks for the tip!

PR #613, opened 2016-03-14
Check windowService exists before adding listener

### LINKED ISSUE #612: New install on windows 7 comes up empty
It doesn't show the new stocks.

Will says he knows how to fix :)

WPFerg:
Closes #612

### DIFF
--- a/src/closedWindows/closedWindowList/closedWindowList-controller.js
+++ b/src/closedWindows/closedWindowList/closedWindowList-controller.js
@@ -43,8 +43,10 @@
         _watch() {
             var listener = () => this.$timeout(() => this.refreshClosedWindows());
             var addListener = () => {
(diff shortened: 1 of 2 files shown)


src/closedWindows/closedWindowList/closedWindowList-controller.js
```diff
@@ -43,8 +43,10 @@
         _watch() {
             var listener = () => this.$timeout(() => this.refreshClosedWindows());
             var addListener = () => {
-                window.windowService.addClosedWindowListener(listener);
-                this.refreshClosedWindows();
+                if (window.windowService) {
```
bjedrzejewski: Should we not add a retry after some time?

WPFerg: The watcher below should be doing that

Issue #574, opened 2016-03-09
Fail gracefully if Quandl times out


owennw:

owennw:
See #70 - make sure we gracefully fail if there is an incorrect API kye

mbssantos:
Could be useful to clarify what to do when:

- Quandl down on start
- Quandl goes down mid use
- Single / some requests fail during use
- Wrong API key used || request limit exceeded

Logic I implemented tries a normal request and:
- If that fails, we try using a fallback option: 
 - Currently an anonymous request but can be easily changed to use a backup key since there's the 50 / day request limit
- If that one also fails the error message is shown to the user on the favorites bar

WPFerg:
Regarding the first two, the long term plan is to have a set of backup data to use in that instance (#559), but that's an 'epic', so in the interim I'd group it with some requests fail.

If some requests fail, I think some form of ratio over time might be good. Say (for example) if >30% fail within 5 minutes, that would be a decent indicator to show a message about difficulties getting data. Or, simpler, if a request fails, show an error; and on the next succeeding request, hide that error.

The fallback for API/request limit seem to be okay to me.

Issue #70, opened 2015-12-11
If API key is wrong, revert to no API key (error handling)


bjedrzejewski:

mbssantos:
There doesn't seem to be any decent way of knowing when BitFlux throws a connection error.
Our options seem limited to:
* Pooling the DOM shortly after a user selection (horrible)
* Listening in on the network (horrible)
* Ask the BitFlux team to extend their API
 * Talking to @jleft it seems simple enough if we can creep it in

WPFerg:
Is there a way to do it in the `quandlService` (which is used for the minicharts)? If the request doesn't succeed, and has message A/B/C, do X/Y/Z?

mbssantos:
@jleft has added the method - `historicFeedErrorCallback(cb)` - to BitFlux's api
It's is called whenever there is a Quandl error and is passed the HTTP error
For example:
`function callback(err) {
    // do something with the error
}`

Passing in the callback will disable all of BitFlux's error notifications from Quandl
The callback can be unset by passing in `null` - `historicFeedErrorCallback(null)` - then the default behavior will resume.

The callback can optionally return a string, which will be shown in an overlay over the chart.

With this change we can catch the error, try again with no Quandl API key and finally show an error if that also fails.

PR #617, opened 2016-03-14
Greaceful quandle failover

mbssantos:
Fix for #574 but not for #70 
show error message on Search and on Favorites when Quandl requests fail


bjedrzejewski:
@mbssantos can you squash the cleanup commits before being ready to merge?

src/sidebars/search/search.html
```diff
@@ -7,6 +7,9 @@
   </div>
   <div id="search-scroll" class="side-scroll custom-scrollbar">
       <div class="sidetab" ng-show="sidebarCtrl.showSearches()">
+          <div class="results-message connection-error" ng-show="!searchCtrl.isConnected">
+              An error occured while retrieving data, please contact support.
```
bjedrzejewski: We need to change this message as there is no support for the app.

src/services/quandl-service.js
```diff
@@ -12,6 +12,34 @@
...
+        };
+
+        proto.notify = function(eventName, value) {
+            this.registry[eventName].forEach(function(entry) {
+                entry.fn.call(entry.scope, value);
+            });
+        };
+
+        return Obs;
+    })();
```
WPFerg: Could this be expressed using the `class` syntax? Out of curiosity, what's the advantage of using this over `$rootScope.$on`/`$broadcast`?

src/services/quandl-service.js
```diff
@@ -113,16 +166,46 @@
...
-                };
-
-                cb(stock);
-            });
+                });
+            }, (request) => {
+                // only use the failsafe once per call
+                if (!isRetry) {
+                    this._notify('CONNECTION_STATUS_CHAGED', (useApiKey = false));
+                    this.getData(stockCode, cb, true);// eslint-disable-line
```
WPFerg: Why is ESLint disabled here?

src/sidebars/search/search.html
mbssantos: How about "An error occurred while retrieving data, please try again later"?

src/services/quandl-service.js
mbssantos: Can do, habit too over me there!
I made it like this so it's tech independent, meaning we won't have to touch it if we move to react.

src/services/quandl-service.js
mbssantos: Good spot, leftover from a console.log I had

src/sidebars/search/search.html
bjedrzejewski: Sounds good!

src/services/quandl-service.js
WPFerg: That's fair enough, it's probably a wiser move (I did something somewhere else in a more reacty-fluxy way), less to migrate

src/sidebars/favourites/favourite-controller.js
```diff
@@ -129,25 +131,25 @@
 
                             // Repeat the check as in the mean time a stock for this favourite could have been added.
                             if (this.stocks.map((stock1) => { return stock1.code; }).indexOf(favourite) === -1) {
-                                var data = stock.data[0],
+                                var data = stock && stock.data && stock.data[0],
```
akhmorgan: I'm surprised that this works. Doesn't this set `data` to a boolean? How then are things like `data.close` accessed later?

mbssantos: in this case it'll only set it to a boolean if either of those is one.
In normal use cases it will either be what we want (the data object) or undefined.
Further down there's  `if(data)...` that cares care of either adding the stock data or creating an error entry

mbssantos:
Closing this PR and opening another to merge with dev instead

PR #621, opened 2016-03-14
Greaceful quandle failover

mbssantos:
Fix for #574 but not for #70 
Show error message on Search and on Favorites when Quandl requests fail


src/services/quandl-service.js
```diff
@@ -113,16 +163,46 @@
...
+        getData(stockCode, cb, isRetry = false) {
+            var self = this;
+            var startDate = period().format('YYYY-MM-DD');
+            var url = (QUANDL_URL + QUANDL_WIKI + stockCode + '.json?' + (useApiKey ? API_KEY_VALUE : '') + '&start_date=' + startDate);
+
             return this.stockData().get({ code: stockCode }, (result) => {
-                var stock = {
-                    name: result.dataset.name,
+                if (!useApiKey && !isRetry) {
+                    this._notify('CONNECTION_STATUS_CHAGED', (useApiKey = true));
```
WPFerg: Should this be `CONNECTION_STATUS_CHANGED`?

src/sidebars/search/search-controller.js
```diff
@@ -9,6 +9,9 @@
             this.selectionService = selectionService;
             this.currentWindowService = currentWindowService;
 
+            this.quandlService.on('CONNECTION_STATUS_CHAGED', this._setConnectionStatus, this);
```
WPFerg: You might be able to replace `_setConnectionStatus` with something like, `(status) => { this.isConnected = status; }`, that way you don't need to pass references to `this` around

mbssantos: it'd look terrible if you had a few of those together though.
Would also be weird reusing code, imo this format makes it nice and straightforward.

WPFerg: Yeah, my primary reason for suggesting it was because the function's only used once, but I don't mind too much.

WPFerg:
Just so I know what to check, is there a timeout (if your network is down, for example) to say that Quandl is unreachable? That and invalid keys are the only thing I can really check

PR #625, opened 2016-03-14
Make the upgrade code tolerant of pre-release versions

WPFerg:


src/store-service.js
```diff
@@ -130,8 +130,8 @@
             var thisVersion = parseVersion(this.version);
             var storedVersion = parseVersion(localStorage.getItem(VERSION_KEY));
 
-            // If there's more than x.y.z, someone's probably tampered.
-            if (storedVersion.length > 3) {
+            // If there's more than x.y.z-rc.a, someone's probably tampered.
+            if (storedVersion.length > 4) {
```
bjedrzejewski: Does this check gives us anything other than potential problems?

bjedrzejewski: Isn't this - return thisVersion[0] !== storedVersion[0]; enough? :)

WPFerg: Probably not, aside from being able to check a malformed version number.

bjedrzejewski: I think we should remove it, this is beyond the scope of the method 'should upgrade'. Unless there is a good reason to keep it.

WPFerg: I've removed it.

PR #628, opened 2016-03-14
Failsafe for Quandl requests - fixes #574

### LINKED ISSUE #574: Fail gracefully if Quandl times out

mbssantos:
Now up to date with dev.
Includes improvements from previous feedback!


bjedrzejewski:
@mbssantos will review tomorrow, thanks for cleanup of the commits!

src/services/quandl-service.js
```diff
@@ -113,16 +163,42 @@
...
+        * @param cb {Function} callback to be called on success and error
+        * @todo use alternative API key instead of defaulting to No key
+        * @todo should we show a warning to the user when we swap to anonymous?
+        */
+        getData(stockCode, cb, isRetry = false) {
             return this.stockData().get({ code: stockCode }, (result) => {
-                var stock = {
-                    name: result.dataset.name,
+                if (!useApiKey && !isRetry) {
+                    this._notify('CONNECTION_STATUS_CHAGED', (useApiKey = true));
```
owennw: I feel like this should be saying `CONNECTION_STATUS_CHANGED`.

bjedrzejewski: Oh yeah, the typo. Took me 30 seconds of starring at it to realize what you mean :)

src/services/quandl-service.js
```diff
@@ -45,6 +70,16 @@
...
     function processResponse(json) {
+        if (!json || json.quandl_error) {
+            return {
+                success: false,
+                error: {
+                    code: json.quandl_error.code,
+                    message: json.quandl_error.message
+                }
+            };
+        }
```
owennw: I'm not sure what this is doing -- `processResponse` previously was not returning anything, but was appending a processed response to the input JSON object. Where this method is called no variable is assigned to the return value, either.

src/services/quandl-service.js
owennw: I'm quite wary of a line doing two things, in this case the `(useApiKey = true)`. I think I'd rather see `useApiKey` being passed in vanilla, but modified explicitly on the line above.

src/services/quandl-service.js
```diff
@@ -12,6 +12,31 @@
...
+
+            this.registry[eventName].push({
+                fn: fn,
+                scope: scope
+            });
+        }
+
+        notify(eventName, value) {
+            this.registry[eventName].forEach((entry) => {
+                entry.fn.call(entry.scope, value);
```
owennw: Do we need to save the scopes and use `call`? Can we get around this with good use of lambdas and calling `entry.fn(value)` directly?

src/services/quandl-service.js
```diff
@@ -75,9 +111,18 @@
 
     class QuandlService {
         constructor($resource) {
+            this.observable = new Observable();
             this.$resource = $resource;
         }
 
+        on() {
+            this.observable.on.apply(this.observable, arguments);
```
owennw: Similarly to the comment above, can we circumvent `apply` here and pass the `arguments` through directly? Also, why not explicitly specify the arguments in this method's signature?

src/services/quandl-service.js
```diff
@@ -113,16 +163,42 @@
...
+            }, (request) => {
+                // only use the failsafe once per call
+                if (!isRetry) {
+                    this._notify('CONNECTION_STATUS_CHAGED', (useApiKey = false));
+                    this.getData(stockCode, cb, true);
+                } else {
+                    // pass data on so an error message can be shown
+                    cb({
+                        success: false,
+                        code: request.status,
```
owennw: There might be some confusion with the `code` property of this object here. The success object has `name, code, data, success` properties, and the failure object has `message, code, details, success`. Maybe this isn't a problem though, but it might get confusing?

src/sidebars/favourites/favourite-controller.js
```diff
@@ -129,25 +131,25 @@
 
                             // Repeat the check as in the mean time a stock for this favourite could have been added.
                             if (this.stocks.map((stock1) => { return stock1.code; }).indexOf(favourite) === -1) {
-                                var data = stock.data[0],
+                                var data = stock && stock.data && stock.data[0],
```
owennw: Do you need the first check for `stock` here? It seems below in the `else` `code` and `message` are being read off the `stock` object without a prior check, so it's assumed the object is not null/undefined.

src/services/quandl-service.js
mbssantos: Oh, well spotted, will correct
On the bright side, they all had the typo!

src/sidebars/favourites/favourite-controller.js
owennw: In fact, isn't there a `success` property on this `stock` object that we can check here?

src/sidebars/favourites/favourite-controller.js
```diff
@@ -156,6 +158,19 @@
             });
         }
 
+        // pseudo private
+        _addError(newError) {
+            var err = this.errors, max = err.length;
+            for (var i = 0; i < max; i++) {
```
owennw: Can this be refactored to look more like `this.errors.forEach((error) => ... )`?

src/services/quandl-service.js
mbssantos: Forcing the use of lambdas can make it awkward to reuse the functions sent anywhere else.
It'll also makes it really verbose on the subscribers side if we add more events.

I'm happy to change it if you feel strongly about it though :)

src/services/quandl-service.js
mbssantos: Used apply so we could change the base "on" method signature without needing to change this one or any others who use it.
Same goes for the _notify

src/sidebars/favourites/favourite-controller.js
bjedrzejewski: Maybe using a map?

src/services/quandl-service.js
mbssantos: Overall I don't see an issue; One is a stock the other is an error.
I understand how "details" might be too generic as a field. Especially since it can be hard to know which code is which (http vs quandl), same for message.
Will rename fields to be more specific and replace "details" with "data"

src/sidebars/favourites/favourite-controller.js
mbssantos: In theory checking the success property should work for almost all cases.
I wrote this thinking there's a chance success data will be missing, although admittedly it'd show an empty code and message.
I felt this provided a bit more robustness, since a malformed success message would break our app.
I've added a stock check to the _addError to be on the safe side

src/sidebars/favourites/favourite-controller.js
mbssantos: Using a forEach would be less efficient though since we'd have to iterate through all elements regardless of when we found it.
On the other hand, it probably doesn't matter since there should rarely be more than 2 showing at any given time.

I'll actually change it to an object and use the error code as the key, this'll remove the "for" altogether. 
This way it'll also match the observable structure

src/sidebars/search/search.less
```diff
@@ -33,6 +33,11 @@
     .font(14px);
   }
 
+  .connection-error {
+    width: 90%;
+    padding: 12px;
+   }
```
owennw: Small whitespace nitpick!

src/services/quandl-service.js
mbssantos: I was using that return for debug - removed now!

mbssantos:
Simplified work based on all feedback

src/services/quandl-service.js
```diff
@@ -133,8 +173,9 @@
         }
 
         // Queries Quandl for all stocks matching the input query
-        stockSearch() {
-            return this.$resource(QUANDL_URL + 'datasets.json?' + API_KEY_VALUE + '&query=:query&database_code=WIKI', {}, {
+        stockSearch(usefallback = false) {
+            usefallback = Math.random() > 0.5;
```
owennw: Not sure if this should be here.. !

bjedrzejewski:
@owennw @mbssantos are we good to merge this now?

owennw:
Just looking at it/testing. I'll merge/update when done.

src/services/quandl-service.js
```diff
@@ -45,6 +48,10 @@
     }
 
     function processResponse(json) {
+        if (!json || json.quandl_error) {
+            return;
+        }
+
```
owennw: Both the call to `filterByDate` and `processResponse` are preceded by a line looking like `json = angular.fromJson(data)`. I think it might make more sense (and reduce duplication!) to move the sanitising checks above into a new method wrapping the `fromJson` call, instead of having them in both these methods.

owennw:
Looking good!

I have a couple of comments which I'll list here. However for the sake of getting this PR in and a release ready I think we should look at them and decide which are important or not, and raise issues for any tidy up for after the merge/release.

- I see a bit of duplication in the code between:

    1. The `_addError` method in the favourites and search controller
    2. The associated HTML for the error message.

For (ii) I think a case of adding a new directive _should_ be simple enough. For (i) I'm not sure what an appropriate solution would be (an errors controller? Some stateful service?).

- I'm unsure about the error code and message. If I try opening the app without an internet connection I get this:
no-results
I get a similar result when attempting a search, in which case the "occurrences" number is always 1. I'm not sure how useful the "No message" message is, and the code for both errors is -1. Also the "no favourites to display" message appears, which could be confusing for a user.

owennw:
Another small thing:

When the internet connection does return (or the API works again) the favourites will not update until a new favourite is added/removed via the search tab. Possibly a spin off to another issue though.

bjedrzejewski:
@owennw @mbssantos I would make the message simpler, as this does not seem quite right.

The internet coming back / restoring - this is a separate issue I guess? Gracefully restore?

owennw:
too-many-requests

Bad API key reverting to default, too many requests.

owennw:
Given the other error the code and message seem useful now. Still not sure if the 'Ocurrences' is worth having, and would like to see the "You have no favourites ..." message removed, as this may prompt the user to think that the error is due to that, and then attempt to add new favourites from the search, which won't work.

owennw:
@rmarcalo Do you have any thoughts on the view of the error messages?

rmarcalo:
Is there any action the user must take? An error message should either tell the user what they should do to correct the problem, or mention that such problem is going to be corrected automatically.

mbssantos:
The aim of this pr was to create a fallback when a normal request fails.
This error message is shown when the fallback also fails.

That can happen on network and / or quandl errors.
The user can check their internet connection, wait until quandl is back up or wait until the key is usable again.

owennw:
@mbssantos raises a good point, this is specifically about a fallback and not error messages. I'm not sure if the original thought was to have better communication with the user too. @ColinEberhardt do you have any thoughts around what we wanted in this release?

From a code point of view it's clean enough to go in and be tidied later if required, we're more focused on UX here I think.

rmarcalo:
Something along the lines of "An error occurred while retrieving data. Please check your internet connection or wait for our data services to be re-established." would be more user friendly and useful (no error codes, or occurrences ).

mbssantos:
Changed error message per @rmarcalo suggestion.

PR #630, opened 2016-03-15
Fix restoring a window setting it to default size; closes #619

### LINKED ISSUE #619: Restoring a window sets it to the default size
Resize a window, maximise, restore. It'll be back to the default height, not the height it was pre-maximise.

WPFerg:


src/toolbar/toolbar-controller.js
```diff
@@ -59,7 +59,6 @@
 
         normalSizeClick() {
             this.window.restore();
-            this.window.resizeTo(defaultWidth, defaultHeight, 'top-right');
```
akhmorgan: I'm fairly sure this was in for a good reason. I think that there's a way to end up with the `restore()` method only returning to the compact size without this line, if I remember rightly. I'll test and see if that's still an issue.

WPFerg: Ah, that makes sense, I'll look at a more comprehensive fix

akhmorgan: Actually the issue I was talking about no longer seems to be present. As I recall the steps to break it were:

- click maximise icon

- click to compact icon

- click expand icon

- click restore icon. This would end up back in a psuedo compact view.

but now clicking the expand icon in a compact window that was previously maximised does not return it to the maximised state so this isn't an issue. Although the fact that it doesn't return to a maximised state may be an issue unto it's own.

WPFerg: I think the issue in this case is that for some reason, that when reverting from compact, it becomes unresizable. I'll close this PR for now, since the implementation is likely to be very different.

WPFerg:
Looks like it was fixing an issue in #411.

PR #643, opened 2016-03-15
Minichart is self-contained; show error if not enough data

### LINKED ISSUE #632: Minichart looks weird if there's not enough history to render
Only seems to appear on certain stocks, for example:

image

WPFerg:
Closes #632 
There was only one bit of history for that stock, so it threw an error in that instance.

I've also made the minichart more self-contained, as it felt weird to have `ng-controller` outside of the `minichart` node.

### DIFF
--- a/src/sidebars/favourites/minichart/minichart-controller.js
+++ b/src/sidebars/favourites/minichart/minichart-controller.js
@@ -3,6 +3,7 @@
 
     class MinichartCtrl {
         constructor(quandlService, $timeout) {
+            this.showMinichart = true;
             this.quandlService = quandlService;
             this.$timeout = $timeout;
         }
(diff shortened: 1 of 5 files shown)


src/sidebars/favourites/minichart/minichart-controller.js
```diff
@@ -16,6 +17,12 @@
                         var width = extent.width,
                             height = extent.height;
                         var data = result.data;
+
+                        if (data.length < 2) {
```
owennw: What does this mean? Why 2?

WPFerg: There needs to be 2 data points to draw the line (otherwise it just does what is in the issue).

WPFerg:
Closing and recreating for `R2`

src/sidebars/favourites/minichart/minichart-controller.js
owennw: Makes sense. Perhaps worth a comment?

PR #644, opened 2016-03-15
Minichart is self-contained; show error if not enough data

WPFerg:
See #643 
Example: stock code SKX.


WPFerg:
FYI @owennw this has the additional comment about the data length

Issue #634, opened 2016-03-15
You can see the selected stock on compact search results


WPFerg:
![image](https://cloud.githubusercontent.com/assets/7962948/13773540/9fa3a25c-ea91-11e5-9b78-5f0ee5576f5a.png)

PR #645, opened 2016-03-15
Don't give the selected stock a background in compact search

WPFerg:


src/sidebars/search/search-compact.less
```diff
@@ -24,6 +24,14 @@
         }
       }
     }
+
+    &.search {
+      .search-result {
+        &.selected.dark {
+          background-color: @light-background;
```
owennw: Why not a single line? `&.search .search-result.selected.dark`?

WPFerg: Good point. I've pushed up a new commit that fixes something, so I don't think it applies now.

src/sidebars/search/search-compact.less
```diff
@@ -24,6 +24,18 @@
         }
       }
     }
+
+    &.search {
+      .search-result {
+        &.selected.dark {
```
owennw: Same comment! `&.search .search-result.selected.dark` can still be the top line here.

WPFerg: Whoops, done

bjedrzejewski:
@WPFerg which issue is this related to?

WPFerg:
#634

PR #648, opened 2016-03-15
Only select a search item for bitflux when favouriting if there's no favourites

### LINKED ISSUE #646: Favouriteless window with focus on a non favourite will change focus to newly added to favourite
Minor inconsistency in behaviour:

- If a user has a favourited stock selected, favouriting other stocks will not change the current chart displayed to the user.

- If a user has a stock selected and has no favourites, favouriting stocks will make the currency chart displayed change to the added favourite.

[bandicam 2016-03-15 15-49-38-607.avi](https://files.zenhub.io/56e82f96a0b9fd8b56595235)

WPFerg:
closes #646

### DIFF
--- a/src/sidebars/favourites/favourite-controller.js
+++ b/src/sidebars/favourites/favourite-controller.js
@@ -100,7 +100,8 @@
                 if (updatedStock) {
                     if (this.stocks.length === 0) {
                         // If there aren't any stocks, we could be adding one...
-                        if (updatedStock.favourite) {
+                        // But only if there's no selection already
+                        if (updatedStock.favourite && this.selection() === '') {
                             this.selectionService.select(updatedStock);
                         } else if (oldSelectedStock.code === updatedStock.code) {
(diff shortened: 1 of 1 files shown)


src/sidebars/favourites/favourite-controller.js
```diff
@@ -100,7 +100,8 @@
                 if (updatedStock) {
                     if (this.stocks.length === 0) {
                         // If there aren't any stocks, we could be adding one...
-                        if (updatedStock.favourite) {
+                        // But only if there's no selection already
+                        if (updatedStock.favourite && !this.selection()) {
```
owennw: This is relying on the empty object in the selection service, right? So also relying on the behaviour of `!` with `''`? If so that might not be obvious for reading/debugging in the future.

WPFerg: Correct, I can change it to `=== ''` if that helps

PR #653, opened 2016-03-16
Fix blurred tearouts breaking the window; closes #538

### LINKED ISSUE #538: Alt-Tab whilst tearing out breaks the tearout window
You can mouse up and move the mouse around without the card following/expanding/etc

Same with the start button

WPFerg:
A few notes:

* This is triggered on any blur event on the tearout window -- as a result of alt-tab, start keyboard button, right click. Maybe the `Escape` key should also trigger it?
* The event for `blur` is added/removed after the window is shown; it stops situations where you tearout, return and try to tearout and it instantly returns (probably due to hiding/showing the window).
* In this case, tearing out 'cancels' the tearout, so it doesn't tear between windows or create a new window.

### DIFF
--- a/src/sidebars/favourites/tearout-directive.js
+++ b/src/sidebars/favourites/tearout-directive.js
@@ -87,6 +87,8 @@
                             clearIncomingTearoutWindow();
                             appendToOpenfinWindow(tearElement, tearoutWindow);
                             displayTearoutWindow();
+
+                            tearoutWindow.addEventListener('blurred', onBlur);
                         }
 
                         function handleMouseMove(e) {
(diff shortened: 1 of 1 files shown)


owennw:
Interesting. I had initially thought that we would just cancel all keyboard events, i.e. on the blur event stop things happening. Is that possible/desired?

In reality why a user would begin tearing out and then press a key is beyond me (besides wanting to watch the world burn :D ). Perhaps an `Escape` press to stop tearing out? It feels like it would be more of an accidental keypress than anything.

owennw:
(code looks good though)

WPFerg:
It's definitely an edge-case thing, I've done it a few times during dev work and it's a bit of a pain to get it back to normal working order.

Regarding stopping the error from happening/stop things happening, you can't get `mouseup` events on non-focused windows, so you could potentially alt-tab and then mouse up with everything breaking. I think just cancelling the tearout is the safest course of action.

PR #668, opened 2016-03-17
Search updates

### LINKED ISSUE #639: Loading search results messages remains after clearing a none matching search
Do a new search that doesn't match e.g. \# and then delete the search term.
Results section shows a "Loading search results..." message rather than being cleared.

### LINKED ISSUE #641: Search results display recently moved stock tabs
User searches for a stock
User goes to their favourites
Drags favourites into a new window
Returns to search menu
All the dragged favourites will be present in the search list[bandicam 2016-03-15 13-07-55-736.avi](https://files.zenhub.io/56e80c78a0b9fd8b56594dd2)

### LINKED ISSUE #663: Entering a character into the search field and then pressing delete can return an empty search result instead of your faves list
If you type a character into the search field and then clear the field quickly, then the search menu doesnt return your favourites list. It returns an empty list instead? Once you get the empty list, you can only restore your faves list in the search result by doing a new search and clearing that one (slowly)

WPFerg:
A fairly hefty refactor of the search component:

Major changes:
* The search part of the Quandl service now callbacks with the array of all stocks, as opposed to calling back for each stock
* Refactored the store service checking to a function `hasStore`, which returns `true` if the store exists. If it doesn't, it checks to see if the store service is available to be bound, and binds it. Otherwise, it returns `false`. Removes a bit of code duplication.
* The favourite updates are now comprehensively applied using more up-to-date data from the store service
* Closes #639 
* Closes #641 
* Closes #663

### DIFF
--- a/src/services/quandl-service.js
+++ b/src/services/quandl-service.js
@@ -20,10 +20,9 @@
     }
 
     function processDataset(dataset, query) {
-        var code = dataset.dataset_code;
         return {
             name: dataset.name,
-            code: code,
+            code: dataset.dataset_code,
             favourite: false,
             query: query
         };
--- a/src/services/quandl-service.js
+++ b/src/services/quandl-service.js
@@ -19,16 +19,14 @@
         return moment().subtract(28, 'days');
     }
 
-    function processDataset(dataset, query, cb) {
+    function processDataset(dataset, query) {
         var code = dataset.dataset_code;
-        var stock = {
+        return {
             name: dataset.name,
             code: code,
             favourite: false,
             query: query
         };
-
-        cb(stock);
     }
 
     function isValidResponse(json) {
@@ -87,11 +85,13 @@
 
         search(query, cb, noResultsCb, errorCb, usefallback = false) {
             this.stockSearch(usefallback).get({ query: query }, (result) => {
-                result.datasets.map((dataset) => {
-                    processDataset(dataset, query, cb);
+                var processedDataset = result.datasets.map((dataset) => {
+                    return processDataset(dataset, query);
                 });
 
-                if (result.datasets.length === 0) {
+                if (processedDataset.length > 0) {
+                    cb(processedDataset);
+                } else {
                     noResultsCb();
                 }
             }, (result) => {
@@ -109,7 +109,7 @@
 
         getMeta(stockCode, cb) {
             this.stockMetadata().get({ 'stock_code': stockCode }, (result) => {
-                processDataset(result.dataset, stockCode, cb);
+                cb(processDataset(result.dataset, stockCode));
             });
         }
 
--- a/src/sidebars/search/search-controller.js
+++ b/src/sidebars/search/search-controller.js
@@ -180,7 +180,8 @@
                             this.stocks[index].favourite = data.favourite;
                         }
                     // The stock doesn't exist, push it on if it's a favourite.
-                    } else if (data.favourite) {
+                    // if there's no search results.
+                    } else if (data.favourite && !this.query) {
                         this.stocks.push(data);
                     }
                 });
--- a/src/sidebars/search/search-controller.js
+++ b/src/sidebars/search/search-controller.js
@@ -172,20 +172,33 @@
                     }
                     var index = this.stocks.map((stock) => { return stock.code; }).indexOf(data.code);
                     if (index > -1) {
-                        if (!this.query) {
-                            // There are no search results, so remove the favourite.
-                            this.stocks.splice(index, 1);
-                        } else {
-                            // Update the stock's favourite
-                            this.stocks[index].favourite = data.favourite;
-                        }
-                    // The stock doesn't exist, push it on if it's a favourite.
-                    // if there's no search results.
+                        this.updateFavouriteStates();
                     } else if (data.favourite && !this.query) {
                         this.stocks.push(data);
                     }
                 });
             });
+        }
+
+        /*
+        * Updates the favourite states depending on whether or not they're in
+        * the window's store (this is the best way to ensure that they are
+        * favourites, as the updateFavourites event is broadcast after the store
+        * has been updated).
+        */
+        updateFavouriteStates() {
+            if (!this.store) {
+                this.store = window.storeService.open(window.name);
+            }
+
+            var favs = this.store.get();
+            if (this.query) {
+                this.stocks.forEach((stock) => {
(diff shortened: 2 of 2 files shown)


src/services/quandl-service.js
```diff
@@ -19,16 +19,14 @@
         return moment().subtract(28, 'days');
     }
 
-    function processDataset(dataset, query, cb) {
+    function processDataset(dataset, query) {
         var code = dataset.dataset_code;
```
akhmorgan: does this var need to exist?

WPFerg: Probably not, I'll remove it

src/sidebars/search/search-controller.js
```diff
@@ -19,15 +19,23 @@
             this._watch();
         }
 
+        hasStore() {
```
mbssantos: I'd expect this method to not change the state of the store.

WPFerg: Perhaps rename it to `getStore()`?

owennw: I agree that it shouldn't change the state. Either a `hasStore` and it doesn't open the store (though I would expect it to check `if (!this.store && !window.storeService)`?), or a `getStore` as you suggested.

WPFerg: I'll do `getStore`, as it allows reducing the boilerplate with setting up the store into one function :)

src/sidebars/search/search-controller.js
```diff
@@ -78,69 +86,83 @@
...
+                            this.noResults = false;
+
+                            stocks.forEach((stock) => {
+                                // not adding old stocks
+                                if (stock.query !== this.query) {
+                                    return;
+                                }
+
+                                var stockAdded = false;
+                                for (var i = 0; i < length; i++) {
```
owennw: This could be changed into a `favourites.forEach(...)`?

WPFerg: Will do, I think this was largely just indentation changes.

src/sidebars/search/search-controller.js
```diff
@@ -78,69 +86,83 @@
...
+                            this.isLoading = false;
+
+                            // removing stocks found with old query
+                            this.stocks = this.stocks.filter((result) => {
+                                return result.query === this.query;
+                            });
+
+                            // Only show the favourites if the query is empty
+                            if (this.query.trim() === '') {
+                                this.displayFavourites();
```
owennw: We already went to the store once here (line 90) to get the favourites. In `displayFavourites` we then go and get them again. Is this what we want?

WPFerg: Probably not. I'll move the favourites var down to where it's used (as part of the above refactor with `.forEach`).

src/sidebars/search/search-controller.js
```diff
@@ -78,69 +86,83 @@
...
                 }
             });
         }
 
+        displayFavourites() {
+            if (this.hasStore()) {
+                this.store.get().map((favourite) => {
+                    this.quandlService.getMeta(favourite, (stock) => {
+                        stock.favourite = true;
+                        this.stocks.push(stock);
```
owennw: This callback used to be run on a single stock at a time, and now it's run on an array of processed stocks, I think? Feels like something should be different here but isn't, and so maybe I'm missing something?

src/sidebars/search/search-controller.js
```diff
@@ -172,21 +194,33 @@
...
-                            // There are no search results, so remove the favourite.
-                            this.stocks.splice(index, 1);
-                        } else {
-                            // Update the stock's favourite
-                            this.stocks[index].favourite = data.favourite;
-                        }
-                    // The stock doesn't exist, push it on if it's a favourite.
-                    } else if (data.favourite) {
+                        this.updateFavouriteStates();
+                    } else if (data.favourite && !this.query) {
```
owennw: I think the comment(s) should return, they made the code more readable imo.

src/sidebars/search/search-controller.js
WPFerg: It's only the `search` function that change applies to; `getMeta` will only return the metadata of one stock (it's updated here: https://github.com/ScottLogic/StockFlux/pull/668/files#diff-c91ca5d182e35cf8cabdf93e69061912R112).

src/sidebars/search/search-controller.js
owennw: Looking through I believe the only use of the store is the `get` method, and after the check that is what we do (since saving new favourites is the responsibility of the star controller I think). Maybe that can be incorporated into the new method?

src/sidebars/search/search-controller.js
WPFerg: I'll add some more.

src/sidebars/search/search-controller.js
owennw: Yes I see. Even though it only returns one stock is that in an array or just the stock object?

WPFerg: Just the stock object, since that's what the API returns.

src/sidebars/search/search-controller.js
WPFerg: I'm not 100% sure what you mean here, do you mean something like:
```js
var store = this.getStore();
if (store) { // something }
```
Instead of having to refer to `this.store` in the `if` block?

owennw: Was thinking of just a method called `getFavourites` that does what your `hasStore` method did, but adds the extra line of returning the favourites? If it can't get them then can return null/undefined as per the suggestion for `getStore`

WPFerg: Ah, that's a good idea. I'll do that.

WPFerg:
I've pushed up a number of commits based upon feedback.

src/sidebars/search/search-controller.js
```diff
@@ -193,8 +193,12 @@
                     }
                     var index = this.stocks.map((stock) => { return stock.code; }).indexOf(data.code);
                     if (index > -1) {
+                        // If the stock in question doesn't exist on the list, check
+                        // the favourites and upadte accordingly
```
owennw: Typo!

owennw:
Looks good! Small typo but that's it.

WPFerg:
Fixed :)

Issue #297, opened 2016-01-29
Investigate keyboard and mouse selection of search results


owennw:
Now that the keyboard can be used to change selected search result the user experience might not be as expected.

By hovering the mouse over one search result, then using the keyboard to select others two separate results can appear highlighting, and then moving the mouse over others won't reduce that down to 1. 

Should selecting with the keyboard act as a pure 'selection', i.e. not change the displayed stock until a mouse click or enter button has been pressed? And how should this work with the mouse?

akhmorgan:
I'd say the behaviour at present is pretty self-explanatory to my mind. If anything I'd say that we should change the hover (mouse over) state of the search results to distinguish it from a selected result. So that it is clear that keyboard changes selected result, whereas mouse over doesn't until you click, which makes sense to me.

owennw:
Interesting. Would you apply the same styling rules to the favourite cards? Separating selected from hovered?

akhmorgan:
I suppose that would make sense for the sake of consistency, so yes.

WPFerg:
I agree. Perhaps a 50% faded background colour when hovering?

akhmorgan:
Yes, something like that should work. We should maybe run any ideas past Rui before proceeding though.

owennw:
Definitely.

mic-smith:
@rmarcalo are you happy with the change to separate the selected and hover state?

rmarcalo:
I believe the proposed solution might work. A 50% opacity for hover might be hard to notice in some monitors but it's definitely worth testing and tweaking as required.

mic-smith:
It's pretty subtle on the favourites. Looks like:
![image](https://cloud.githubusercontent.com/assets/17762186/13841328/2dac7b8e-ec20-11e5-82d3-334c3dd5c193.png)

![image](https://cloud.githubusercontent.com/assets/17762186/13841331/32384a66-ec20-11e5-8464-f9c2941f275e.png)

owennw:
@mic-smith I can't see the difference on the favourites! :/ Might be my monitor/contrast/eyes though.

mic-smith:
@owennw Yep it's not the most obvious. Going much lighter went too close to the normal background colour - at least on my monitor.
I tried a 50% alpha version which ended up with a green hint from the drop area below. This is more prominent to my eyes and similar to the hover effect on the collapsed favourites. But then ends up not being as consistent with the hover effect on the search in terms of just darkening the selection. 

[...]

PR #669, opened 2016-03-17
Lighter hover colour for sidebar for #297

mic-smith:


src/sidebars/favourites/favourites.less
```diff
@@ -26,6 +26,10 @@
       background-color: @highlight-colour;
     }
   }
+
+  .darkens:hover {
+    background: #1F262E;
```
bjedrzejewski: Can we use a variable here? If one does not exist for this color, let's create one.

owennw: Interestingly the other change uses a `fade`, but we don't want one here?

mic-smith: Sure. Will pull out a variable and I think there's at least one other colour in that file that's not using one too. Will fix them up too.

The favourites needed an override for the style as the fade version with 0.5 alpha meant the green of the drop-area style showed through for the hover style.

owennw: That makes sense.

mic-smith:
Closing as target branch needs changing

PR #678, opened 2016-03-17
Auto deploy to develop and version on gh-pages.

### LINKED ISSUE #676: Use directory structure for releasing to gh-pages
The idea is to use directory structure for gh-pages release.

There most likely will be:
-Latest release
-Development
-Old releases

Details will become more clear once the pr is in

bjedrzejewski:
Closes #676

Looking for early feedback

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+set -eo pipefail
+if ([ $TRAVIS_PULL_REQUEST == "false" ]  && [ "${TRAVIS_REPO_SLUG}" != "ScottLogic/StockFlux" ])
+then
+    #Clone the latest gh-pages
+    git clone https://github.com/ScottLogic/StockFlux.git --branch gh-pages gh-pages
+
+    #Get line with version from the file -> get the second word -> remove quotes around the value
+    VERSION=$(grep "version" package.json | awk -v N=$2 '{print $2}' | cut -d \" -f2)
+
+    #Get line with the release type (develop/master) from the file -> get the second word -> remove quotes around the value
+    TYPE=$(grep "release-type" package.json | awk -v N=$2 '{print $2}' | cut -d \" -f2)
+
+    if ([ -z '$TYPE' ] || [ -z '$VERSION' ])
+    then
+        echo 'Version or Type not set in package.json'
+        exit 1
+    fi
+
+    rm -rf './gh-pages/$TYPE'
+    cp -r './public' './gh-pages/$TYPE'
+    rm -rf './gh-pages/$VERSION'
+    cp -r './public' './gh-pages/$VERSION'
+    cd gh-pages
+
+    #Removing git history
+    rm -rf .git
+    git init
+
+    # inside this git repo we'll pretend to be a new user
+    git config user.name "Travis CI"
+    git config user.email "travis@scottlogic.com"
+
+    # The first and only commit to this new Git repo contains all the
+    # files present with the commit message "Deploy to GitHub Pages".
+    git add .
+    git commit -m "Deploy to GitHub Pages"
+
+    # Force push from the current repo's master branch to the remote
+    # repo's gh-pages branch. (All previous history on the gh-pages branch
+    # will be lost, since we are overwriting it.) We redirect any output to
+    # /dev/null to hide any sensitive credential data that might otherwise be exposed.
+    git push --force --quiet "https://${GH_TOKEN}@${GH_REF}" master:gh-pages > /dev/null 2>&1
+fi
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 set -eo pipefail
-if ([ $TRAVIS_PULL_REQUEST == "false" ]  && [ "${TRAVIS_REPO_SLUG}" != "ScottLogic/StockFlux" ])
+if ([ $TRAVIS_PULL_REQUEST == "false" ]  && [ "${TRAVIS_REPO_SLUG}" != "ScottLogic/StockFlux" ] && ([ $TRAVIS_BRANCH == 'dev' ] || [ $TRAVIS_BRANCH == 'master' ]))
 then
     #Clone the latest gh-pages
     git clone https://github.com/ScottLogic/StockFlux.git --branch gh-pages gh-pages
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -2,7 +2,9 @@
 'use strict';
 module.exports = function(grunt) {
     var target = grunt.option('target') || 'http://localhost:5000',
-        port = process.env.PORT || 5000;
+        port = process.env.PORT || 5000,
+        version = grunt.file.readJSON('package.json').version,
+        type = grunt.file.readJSON('package.json').type;
 
     grunt.initConfig({
         'gh-pages': {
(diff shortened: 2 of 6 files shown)


bjedrzejewski:
This is done now, waiting for proper feedback.

gruntfile.js
```diff
@@ -87,9 +89,15 @@ module.exports = function(grunt) {
         },
 
         download: {
+            //You need to make sure to create the correct zip here
```
owennw: The grunt `createZip` task: what will it do?

deploy.sh
```diff
@@ -0,0 +1,36 @@
+if ([ $TRAVIS_PULL_REQUEST == "false" ])
+then
+    #Clone the latest gh-pages
+    git clone https://github.com/ScottLogic/StockFlux.git --branch gh-pages gh-pages
+
+    #Get line with version from the file -> get the second word -> remove brackets
+    VERSION=$(grep "version" package.json | awk -v N=$2 '{print $2}' | cut -d \" -f2)
+
+    #Get line with the release type (develop/master) from the file -> get the second word -> remove brackets
```
owennw: The remove brackets comment confused me for a while.. I think it should be "quotes".

deploy.sh
```diff
@@ -0,0 +1,36 @@
+if ([ $TRAVIS_PULL_REQUEST == "false" ])
+then
+    #Clone the latest gh-pages
+    git clone https://github.com/ScottLogic/StockFlux.git --branch gh-pages gh-pages
+
+    #Get line with version from the file -> get the second word -> remove brackets
```
owennw: Brackets => quotes here too

deploy.sh
bjedrzejewski: well spotted, will fix and squash commits before merging

deploy.sh
```diff
@@ -0,0 +1,36 @@
...
+    VERSION=$(grep "version" package.json | awk -v N=$2 '{print $2}' | cut -d \" -f2)
+
+    #Get line with the release type (develop/master) from the file -> get the second word -> remove brackets
+    TYPE=$(grep "type" package.json | awk -v N=$2 '{print $2}' | cut -d \" -f2)
+
+    rm -rf ./gh-pages/$TYPE
+    cp -r ./public ./gh-pages/$TYPE
+    rm -rf ./gh-pages/$VERSION
```
owennw: Why do we need to delete the folder `./gh-pages/$VERSION', I wouldn't expect that folder to exist yet? Is this being defensive in case it already exists?

package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "bitflux-openfin",
   "version": "10.0.0-rc.1",
+  "type": "develop",
```
owennw: So to do a 'proper' release we need to manually change this line to `"type": "master"`? Is that correct?

If so, is there a way to automate that into the `grunt release(/deploy?)` task?

deploy.sh
bjedrzejewski: It can already exists if someone re-pushes. Very easy to happen, a pain to fix without that line.

deploy.sh
```diff
@@ -0,0 +1,36 @@
...
+    cp -r ./public ./gh-pages/$VERSION
```
owennw: If the `type` is "develop" we'll still be creating a version folder for it? That's probably okay for now, but do we actually want that? (I don't know)

package.json
bjedrzejewski: Yes, to do a proper release you want to change that to master and yes, automating it in grunt is a great idea! Will get on it.

deploy.sh
owennw: If someone repushes with the same version it feels like that's because a mistake was made? Is that correct?

gruntfile.js
```diff
@@ -87,9 +89,15 @@ module.exports = function(grunt) {
...
             openfinZip: {
-                src: ['https://dl.openfin.co/services/download?fileName=StockFlux&config=http://scottlogic.github.io/StockFlux/app.json'],
-                dest: './public/StockFlux.zip'
+                src: ['https://dl.openfin.co/services/download?fileName=StockFlux-' + version +
+                '&config=http://scottlogic.github.io/StockFlux/' + version + '/app.json'],
+                dest: './public/StockFlux-' + version + '.zip'
+            },
+            openfinTypeZip: {
+                src: ['https://dl.openfin.co/services/download?fileName=StockFlux-' + type + '&config=http://scottlogic.github.io/StockFlux/' + type + '/app.json'],
+                dest: './public/StockFlux-' + type + '.zip'
```
owennw: Does grunt support functions by the way? Looks like with an appropriate input parameter (type/version) this could be less duplicative.

deploy.sh
bjedrzejewski: Maybe a mistake, maybe some hotfix, maybe fixing a mistake. Any of these scenarios is easier with this line in, so I strongly recommend keeping it in.

deploy.sh
bjedrzejewski: I did it on purpose, as I thought that its great to keep developer versions, to ease testing. We can share any version with anyone for testing, showing stuff, trying. It does not limit us, it only empowers us :D

owennw:
I've looked it over -- as you say it's hard to actually test. Most of the comments I've raised are just questions (based on my lack of full understanding of this).

When it's all done a squash + rebase would be nice!

gruntfile.js
bjedrzejewski: I was thinking about it as well. I think that maybe its a better approach to make it into another issue- proper cleanup of grunt the same way you did with less.

bjedrzejewski:
Thanks a lot for that @owennw. I will automate away a bit more with grunt and squash/rebase etc at the end before merging.

deploy.sh
owennw: Yeah I see the reasons for it. My only concern is around our version folders accidentally being deleted. It puts the onus on the developer not to execute a silly grunt task to wipe out something old. Whereas without the line if a mistake is made then yes it might be hard to fix, but at least you know all the older stuff is safe? 

Maybe I'm overestimating the "problem" though. Wouldn't mind a third opinion here.

bjedrzejewski: @WPFerg @akhmorgan @mbssantos @mic-smith third opinion needed.

I will just remind one more thing- this will only ever happen after the PR is merged, never before- it needs more than one person to make a mistake.

deploy.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+if ([ $TRAVIS_PULL_REQUEST == "false" ])
```
mic-smith: The BitFlux deploy script has a few extra checks - e.g. not a fork.

Do we want this too?

bjedrzejewski: Yes, I didn't think about it, but I guess we don't want this to break travis on peoples forks. Will add.

deploy.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
```
mic-smith: We should fail on first error:

`set -e`
` set -o pipefail`

deploy.sh
WPFerg: I'm not sure. I think it cleaning the current version is a good idea. If there's no commit history, then everything would be overwritten anyway, and you'd want the folder to be as clean as possible. Re; fixing the version -- is it a case of re-pushing from the version-tagged commit?
Or am I misunderstanding it?

bjedrzejewski: @WPFerg if merge `"version": "10.0.0-rc.1",` twice it will overwrite the directory on second merge, which I think is good.

deploy.sh
bjedrzejewski: @mic-smith so you suggest just adding:

`set -e
set -o pipefail`

straight after the the shebang?

deploy.sh
owennw: I agree with that. I was more just (perhaps overly) concerned with accidentally overwriting an older version (like `"10.1.0"` or something).

deploy.sh
mic-smith: Yes on line after shebang.

akhmorgan:
I've looked over the PR, I didn't spot any issues, so all good from me

deploy.sh
mic-smith: Should we also check for none empty on the VERSION and TYPE? 
E.g. before we do ` rm -rf ./gh-pages/$TYPE` 
And maybe quote the args.

WPFerg:
I can't spot anything, either

deploy.sh
mic-smith: I can see the advantage for fixing up issues, but I'm a bit wary on the versions overwrite, especially if we changed the Travis config to build on other branches etc., but I think fine for this PR.  The planned structure looks good and we can tweak the actually deploy.sh as needed later

PR #684, opened 2016-03-18
Fixing quotes and logic in deploy.sh

bjedrzejewski:


deploy.sh
```diff
@@ -1,26 +1,26 @@
 #!/usr/bin/env bash
 set -eo pipefail
-if ([ "${TRAVIS_REPO_SLUG}" != "ScottLogic/StockFlux" ] && ([ $TRAVIS_BRANCH == 'dev' ] || [ $TRAVIS_BRANCH == 'master' ]))
+if ([ $TRAVIS_PULL_REQUEST == "false" ]  && [ "${TRAVIS_REPO_SLUG}" == "ScottLogic/StockFlux" ] && ([ $TRAVIS_BRANCH == "dev" ] || [ $TRAVIS_BRANCH == "master" ]))
```
WPFerg: Didn't the last PR get rid of `$TRAVIS_PULL_REQUEST == "false"`?

bjedrzejewski: Because the issue was not with that, but with the REPO_SLUG

PR #687, opened 2016-03-18
Checking for correct 'type' parameter in deploy.sh

bjedrzejewski:


deploy.sh
```diff
@@ -10,7 +10,7 @@ then
     echo "Version is: $VERSION"
 
     #Get line with the release type (develop/master) from the file -> get the second word -> remove quotes around the value
-    TYPE=$(grep "release-type" package.json | awk -v N=$2 '{print $2}' | cut -d \" -f2)
+    TYPE=$(grep "type" package.json | awk -v N=$2 '{print $2}' | cut -d \" -f2)
```
owennw: I know this works now, but what if in the future we add a new dependency with the word "type" in it? (same for the other keyword I guess)

bjedrzejewski: it takes the first one it finds

PR #699, opened 2016-03-21
Mixins

owennw:
Adds a size mixin, and a tiny refactor to the overflow CSS in the minichart.

Note I've not included the size mixin in the showcase overrides.

### DIFF
--- a/src/sidebars/favourites/minichart/minichart.less
+++ b/src/sidebars/favourites/minichart/minichart.less
@@ -3,13 +3,12 @@
 minichart {
   .position(absolute);
   padding-top: 10px;
-  width: 100%;
+  .size(@width: 100%);
   overflow: visible;
 }
 
 .minichart {
-  width: @chart-width;
-  height: @chart-height;
+  .size(@width: @chart-width, @height: @chart-height);
   overflow: visible !important;
 
   svg {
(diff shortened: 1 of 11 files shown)


src/sidebars/sidebar.less
```diff
@@ -2,28 +2,26 @@
 @flash-colour: #447272;
 @bottom-tab-height: 40px;
 
-@-webkit-keyframes search {
-  0% {
+.click-flash(@final-colour) {
+    0% {
```
WPFerg: Indentation

WPFerg:
One small whitespace comment but looks fine to me. I'm guessing you can't do something like this:

```less
.class {
  .size(@width: 10px, @height:10px);

  &.some-other-class {
    .size(@height: 10px);
  }
}
```
because `@width` is overridden? (Like [here](https://github.com/ScottLogic/StockFlux/pull/699/files#diff-da44fe9f3c295da97d70c9a6a6ced0adR6)?)

owennw:
Yes that is the case at the moment, which is maybe not the best. Any ideas on how to improve that?

WPFerg:
Theres [guards](http://lesscss.org/features/#css-guards-feature) but I'm not sure if they'd work for something like `.size( ... ) when (@width)`

src/sidebars/favourites/minichart/minichart.less
```diff
@@ -1,19 +1,22 @@
 ﻿@minichart-error-colour: white;
 
+.overflow {
+  overflow: visible;
```
WPFerg: Trying it out shows some suspicious line endings here

owennw:
Updated the commits now.

Issue #696, opened 2016-03-21
Improve the GH Pages deployment


WPFerg:
- [ ] Currently based off of a field in `package.json`, which is causing merge conflicts between `dev` and `master`. Perhaps push to a folder with the name `$TRAVIS_BRANCH` instead of what's in the `package.json` file?
- [ ] Each app.json points to the develop folder, instead of their own folders.

owennw:
How will this play with #680?

WPFerg:
It would close it as editing `type` would no longer be needed.

WPFerg:
Closed by #702

Issue #693, opened 2016-03-21
Cleanup gh-pages once everyone migrated to new versions


bjedrzejewski:
Gh-pages currently holds old directories not to break old demos.

Once these are no longer used, it should be cleaned.

WPFerg:
We can probably clean it now, and update it to point to the same paths as in `master/app.json`

PR #701, opened 2016-03-21
Tidy grunt file

owennw:

PR #702, opened 2016-03-21
GH Pages Build Updates

WPFerg:
* Uses the travis branch to determine the folder to push to #696 
* Builds two versions; the `TYPE` version, ie dev/master; and the `VERSION` version (ie 10.0.0)

* Will create a new `dev` folder instead of develop (bear in mind for #693)

### DIFF
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -3,8 +3,7 @@
 module.exports = function(grunt) {
     var target = grunt.option('target') || 'http://localhost:5000',
         port = process.env.PORT || 5000,
-        version = grunt.file.readJSON('package.json').version,
-        type = grunt.file.readJSON('package.json').type;
+        buildTarget = grunt.option('build-target') || 'dev';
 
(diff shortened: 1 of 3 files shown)


gruntfile.js
```diff
@@ -236,6 +231,17 @@ module.exports = function(grunt) {
...
                 }
+            },
+            'gh-pages': {
+                files: {
+                    'public/app.json': 'public/app.json'
+                },
+                options: {
+                    replacements: [{
+                        pattern: /\/develop\//g,
+                        replacement: '/' + buildTarget + '/'
```
WPFerg: Ideally the pattern should be more flexible than this (I'll look at doing some path-based-fancy-regex)

owennw:
Would you mind incorporating some of the changes from #701 here too? Removing old tasks and dependencies etc.

WPFerg:
@owennw Done -- I've removed `grunt-gh-pages`.

PR #715, opened 2016-03-22
Tearout card reordering is done by intersection area

### LINKED ISSUE #636: Re-organisation of favorites doesnt always work properly
- A user goes to move a favourite, they hold down their mouse button on it
- Realising they dont want to move it, they let go of the mouse button
- Provided they have moved the mouse a small amount, their favourite will be moved into the next position up the list, despite this not being their intention

A system closer to how chrome allows you to reorganise your favourites would be favourable. If you want to move a favourite up by one place, you should need to move the mouse purposefully up one place in the favourites[bandicam 2016-03-15 10-47-09-825.avi](https://files.zenhub.io/56e7ea48a0b9fd8b56594b07)

WPFerg:
Closes #636 -- intersection's done by intersection area with the card, as opposed to using the divider for intersection checks.

### DIFF
--- a/src/sidebars/favourites/geometry-service.js
+++ b/src/sidebars/favourites/geometry-service.js
@@ -45,6 +45,18 @@
                 otherOrigin.x < this.corner().x &&
                 otherOrigin.y < this.corner().y;
         }
+
+        areaOfIntersection(otherRectangle) {
+            if (this.intersects(otherRectangle)) {
+                var left = Math.max(this.left(), otherRectangle.left());
+                var right = Math.min(this.right(), otherRectangle.right());
+                var top = Math.max(this.top(), otherRectangle.top());
+                var bottom = Math.min(this.bottom(), otherRectangle.bottom());
+                return (right - left) * (bottom - top);
+            } else {
+                return 0;
+            }
+        }
     }
 
     // Helper function to retrieve the height, width, top, and left from a window object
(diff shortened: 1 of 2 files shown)


src/sidebars/favourites/geometry-service.js
```diff
@@ -45,6 +45,18 @@
...
         }
+
+        areaOfIntersection(otherRectangle) {
+            if (this.intersects(otherRectangle)) {
+                var left = Math.max(this.left(), otherRectangle.left());
+                var right = Math.min(this.right(), otherRectangle.right());
+                var top = Math.max(this.top(), otherRectangle.top());
+                var bottom = Math.min(this.bottom(), otherRectangle.bottom());
+                return (right - left) * (bottom - top);
+            } else {
```
owennw: I'm about to change this, but why the farthest left and closest right as opposed to vice versa?

src/sidebars/favourites/tearout-directive.js
```diff
@@ -137,7 +137,6 @@
                                         }
 
                                         // Remove drop-target from original instance
-                                        parent.removeChild(myHoverArea);
```
owennw: So we're not doing this any more?

owennw:
Code looks reasonable, haven't tested though (and won't as I'm in the middle of changing the files: if someone else could that would be great).

src/sidebars/favourites/tearout-directive.js
WPFerg: The equivalent of this would be `parent.remove()`, but since everything's removed anyway on tear out I didn't think it worth it

src/sidebars/favourites/geometry-service.js
WPFerg: That describes the rectangles' intersection (farthest left will always be greater than closest right if there's an intersection, etc)

jackarnstein:
Given it a good bit of testing and as far as I can tell this works really well! It makes the process alot smoother for the user.

The only problem with dragging now is that if you have many windows open, sometimes its hard to make tabs move into specifc other windows (they just seem to ignore them). This issue is however, totally unrelated to Wills fix right here and requires further investigation

ColinEberhardt:
We need an npm micro-module for rectangle intersections ...

https://github.com/d3fc/d3fc-label-layout/blob/master/src/intersect.js

:-P

WPFerg:
That may or may not be my inspiration, but it's alright, I wrote that too ;)

WPFerg:
Resolved some more conflicts, might be worth a retest

Issue #631, opened 2016-03-15
Line separator before stock cards


rmarcalo:
While I intentionally left this out of the design for initial aesthetic reasons, I think the favourite tab would be more balanced if a line separator (similar to the one before the Close Windows icon) was introduced below the Favourites Tab icon, thus separating this icon from the first stock card. The same line would be on the Search tab, right below the text input.

![visual spec - latest](https://cloud.githubusercontent.com/assets/13066377/13772614/05b7f5a2-ea8d-11e5-8b53-cd41d3be0f4a.png)

mic-smith:
Should the line be the same colour on both tabs and should they show on the collapsed versions?
This is what it looks like:

![image](https://cloud.githubusercontent.com/assets/17762186/13776189/bc247fc8-eaa0-11e5-91cf-7104553cb9e1.png)

![image](https://cloud.githubusercontent.com/assets/17762186/13776223/f2263512-eaa0-11e5-9a62-ee7afb3105fc.png)

![image](https://cloud.githubusercontent.com/assets/17762186/13776229/f766023c-eaa0-11e5-9fd1-3caf3a4bf34a.png)

owennw:
@rmarcalo

rmarcalo:
The search tab line could have the colour of the favourite tab background instead, to make it less striking.

PR #717, opened 2016-03-22
Add line separator on side bar for #631

mic-smith:


src/assets/styles/custom-scroll.less
```diff
@@ -4,6 +4,8 @@
 
 .side-scroll {
   padding-bottom: 0px;
+  border-top: 1px solid @dark-background;
+  transition: border 0.7s;
```
WPFerg: Could this be replaced by the `.transition` mixin?

owennw: There's also a specific bit of LESS in search.less for something similar:

  ```.sidetab > div:first-of-type search .search-result {
    border-top: @search-splitter-style;
  }```

Can something be refactored to make these more common?

mic-smith: @WPFerg Could replace with the transition mixin - though not sure what it adds and this is more explicit that it's only the border being animated?

@owennw That selector doesn't match anything. I'll remove

owennw: @mic-smith Doesn't that selector create the top border that we do have on the search tab?

WPFerg: @mic-smith It doesn't really add much, it just adds a bit of consistency -- but it's certainly not something I'd insist on

mic-smith:
Rebased on dev and repushed

owennw:
LGTM.

PR #718, opened 2016-03-22
Resizability on compact-created windows

### LINKED ISSUE #633: Expanding from compact will never revert to a maximised window
This is similar to #619 

If you have a maximised window and you click the compact icon, and then click the expand icon, you'd expect the window to revert to it's previous state of maximised, but it does not.

This is because the `maximised` flag in the toolbar controller is set based on the `maximise `and `restore` events on the window. Which makes sense, however the `restore `event, which sets the flag to false is triggered by the window becoming compact, and so the state is not maintained correctly.

### LINKED ISSUE #647: Resizing issues when restoring as compact window after restart.
When window is restored as compact (restarting the app) and then made not compact, resizing doesn't work.

WPFerg:
Closes #647 

* Adds a 'shared' config, which adds the shared values to main window/tearout card configs (less to maintain)
* Removes the compact window config in favour of checking to see if the window is indeed compact when it's about to be displayed, and then updating the compact flags. The positioning works between app runs because of `top-left`.
* Commit 3 closes #633 by adding a flag for `wasMaximised`, as the `restored` event fired when switching between maximised and compact view

### DIFF
--- a/src/config-service.js
+++ b/src/config-service.js
@@ -65,22 +65,6 @@
             });
         }
 
-        getCompactConfig(name) {
-            return this._getConfig(name, {
-                showTaskbarIcon: true,
-                saveWindowState: true,
-                url: 'index.html',
-                resizable: false,
-                maximizable: false,
-                minWidth: COMPACT_WINDOW_DIMENSIONS[0],
-                minHeight: COMPACT_WINDOW_DIMENSIONS[1],
-                maxWidth: COMPACT_WINDOW_DIMENSIONS[0],
-                maxHeight: COMPACT_WINDOW_DIMENSIONS[1],
-                defaultWidth: COMPACT_WINDOW_DIMENSIONS[0],
-                defaultHeight: COMPACT_WINDOW_DIMENSIONS[1]
-            });
-        }
-
         getTearoutConfig(name) {
             return this._getConfig(name, {
                 maximizable: false,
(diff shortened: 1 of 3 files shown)


akhmorgan:
Looks good to me. Tested locally and all seems fine.

WPFerg:
Resolved the merge conflicts, it might be worth re-testing

src/config-service.js
```diff
@@ -50,44 +39,42 @@
...
                 saveWindowState: true,
                 url: 'index.html',
-                resizable: false,
-                maximizable: false,
-                minWidth: COMPACT_WINDOW_DIMENSIONS[0],
-                minHeight: COMPACT_WINDOW_DIMENSIONS[1],
-                maxWidth: COMPACT_WINDOW_DIMENSIONS[0],
-                maxHeight: COMPACT_WINDOW_DIMENSIONS[1],
-                defaultWidth: COMPACT_WINDOW_DIMENSIONS[0],
-                defaultHeight: COMPACT_WINDOW_DIMENSIONS[1],
```
owennw: Are these constants still used?

WPFerg: No, but I'll change a line to make use of them

WPFerg: Done.
Also fixed the merge conflicts.
Again.

akhmorgan:
Re-tested. Looks good to me. Happy to merge, pending Travis' approval.

PR #719, opened 2016-03-22
removes unnecessary variables and tidies up sidebar controller displa…

akhmorgan:
Minor re-factor of sidebar controller to remove a few unnecessary variables and generally make the search/favourites toggling logic a bit nicer.


owennw:
LGTM.

Issue #706, opened 2016-03-21
Visual and Interaction Adjustments - Tile spacing


mic-smith:
Tile size and spacing updates from #685

PR #721, opened 2016-03-22
Tearout whole card

owennw:
This PR enables tearing out a favourite from anywhere within the card.

Because these cards also can be selected I've implemented a delay on holding the mouse button (of 200ms) to distinguish between a click and an intentional hold.

Within these 200ms a user could have already started dragging the card and get no visual reply, so a card will also tear if the mouse is moved outside of a 50px square from where it started. (Squares are quicker to code than circles, but it's not hard to change).

Questions:

- The card can be torn from the star icon -- is this okay?
- Should the cursor over the card be different now (perhaps the hand?)
- In compact view there's not much user feedback -- should we reinstate hover states?

Issues:

- [x] When the unfavouriting confirmation is open the card can still be dragged. It looks like there's code to prevent this, so I'll be looking into it further with a view to updating this PR or submitting a new one.
- [x] Confirmation window slightly offset
- [x] Scroll bar appearing ruins it all


akhmorgan:
I'd rather see the style completed before merging.

WPFerg:
I agree with @akhmorgan. Also upon testing, the tear in indicator appears when you tear between windows, but if you move the card back out the indicator doesn't go away.

owennw:
@WPFerg So it does! I'll add to this PR with the style changes for #706 then.

owennw:
This is done and ready for review now.

WPFerg:
Quite a small issue I've found when using it -- sometimes when I tear out a favourite card, the background flickers on mouse move.

akhmorgan:
looks good to me.

src/sidebars/favourites/favourites.less
```diff
@@ -1,11 +1,13 @@
-﻿@favourite-height: 100px;
+﻿@favourite-height: 110px;
 @favourite-padding-left: 0px;
-@favourite-padding-top: 5px;
-@favourite-padding-right: 4px;
-@favourite-padding-bottom: 10px;
+@favourite-padding-top: 10px;
+// Padding right is declared in the variables file.
```
WPFerg: Although moving it out is justified, I'm not too sure on having only `favourite-padding-right` moved out. it doesn't seem too consistent to me. Perhaps have a new variable, which is then assigned to `@favourite-padding-right` in this file?

owennw: Yes I felt similar about it. I wasn't sure what that variable would be called though: I couldn't think of a general yet specific name. 

They could all be moved out, but then is there a need for that?

src/assets/styles/style.less
```diff
@@ -15,7 +15,7 @@
 @import "../../sidebars/favourites/minichart/minichart.less";
 @import "../../sidebars/star/star.less";
 @import "../../showcase/showcase-changes.less";
-@import "custom-scroll.less";
+@import "../../sidebars/custom-scroll.less";
```
akhmorgan: Is there any particular reason why this has been moved? I know we're only using the scroll in the sidebar at the minute, but it still seems to me like it's applications are sufficiently abstract/general to warrant being in a high level directory. Also my personal preference is to have file moves like this in a separate PR unless they're directly related/necessary for the change that this PR is making.

owennw: The reason for the move was that we're only using it in the sidebar, and every time I went to find it I had to search around for not knowing where it was!

Happy to submit as part of another PR though, or leave where it is.

owennw:
Re: the background flicker. It appears to only happen when the card is torn out via the mouse movement detection, as opposed to the 200ms timeout. I'm investigating now and will hopefully fix it for this PR.

src/sidebars/favourites/favourites.less
WPFerg: There isn't really a need for that, no. Perhaps a brief explanation why in a comment instead? I can't think of a name that would go welll, either.

WPFerg:
One other quick comment from use, I think the delay on click to create the card is a bit too short. If you click not unusually slowly, you can get the card tearing out as you're mousing up. It doesn't break anything, but it feels to me like it should be selecting instead of starting the tearout, although this might be a comment for a later issue/UX fix.

owennw:
The issue of the background colour changing is tricky to pin down and I don't think it should prevent this PR from being merged:

- It only happens with specific (speedy) actions
- I didn't notice it until it was pointed out
- I don't think a user is likely to spend a lot of time dragging around a tearout card

I'll raise another issue for it when this is merged.

WPFerg:
Probably an important point, come to think of it: how should the positioning of new torn out windows work? Previously it was a hard-coded offset to the middle of the top tearout card icon. It doesn't feel right in this instance that the mouse is now in the top right corner of the top card.

owennw:
Updated now.

WPFerg:
Seems to work fine, do you want to bring it up to date and then post-travis, we can merge.

PR #723, opened 2016-03-22
Fix compact scrollbar going off the bottom

### LINKED ISSUE #722: Bottom stock cut off in compact
image
I suspect it's because of the `@height: inherit` part of the mixin in `sidebar-compact.less` -- change it to just `width: 100%`, so the scroll bar continues all the way over the search expansion box?

WPFerg:
Closes #722


WPFerg:
Closing for now, the height of the favourite card is likely to change, so would be invalidated fairly quickly.

PR #738, opened 2016-03-23
Fix mousedown on favourite removal confirm button triggering hide modal

### LINKED ISSUE #731: Unstarring favourites doesn't work
Broken on dev

WPFerg:
Closes #731 

Adds a `mousedown` event to the confirmation modal to stop the propagation. I'd tried doing some pointer event CSS magic to hopefully make this a bit nicer to approach, but it didn't work, so this was the next best alternative

### DIFF
--- a/src/sidebars/star/star.html
+++ b/src/sidebars/star/star.html
@@ -2,7 +2,8 @@
      ng-mouseenter="starCtrl.mouseEnter()" ng-mouseleave="starCtrl.mouseLeave()" title="{{starCtrl.tooltip()}}" draggable="false" />
 <div class="confirmation-backdrop" ng-if="starCtrl.confirmationShow" ng-mousedown="starCtrl.hideModal(); $event.stopPropagation();">
   <div class="bubble-head" ng-style="{top:starCtrl.modalBubbleTop()+'px'}"></div>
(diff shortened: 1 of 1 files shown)


src/sidebars/star/star.html
```diff
@@ -2,7 +2,8 @@
      ng-mouseenter="starCtrl.mouseEnter()" ng-mouseleave="starCtrl.mouseLeave()" title="{{starCtrl.tooltip()}}" draggable="false" />
 <div class="confirmation-backdrop" ng-if="starCtrl.confirmationShow" ng-mousedown="starCtrl.hideModal(); $event.stopPropagation();">
   <div class="bubble-head" ng-style="{top:starCtrl.modalBubbleTop()+'px'}"></div>
-  <div class="confirmation-modal" ng-class="{flipped: starCtrl.modalFlip()}"  ng-click="$event.stopPropagation();" ng-style="{top:starCtrl.modalTop()+'px'}">
+  <div class="confirmation-modal" ng-class="{flipped: starCtrl.modalFlip()}" ng-click="$event.stopPropagation();"
```
owennw: Do we still need the `ng-click` too?

WPFerg: Good point. Probably not since the issue was because the `click` moved to `mousedown`, and we mustn't have updated it.

WPFerg: Actually, it is needed. You can click through it and select another stock on non-compact mode (same with double click, and I'll push that fix up in a sec)

Issue #707, opened 2016-03-21
Visual and Interaction Adjustments - Collapse search by clicking icon


mic-smith:
Collapse search by clicking icon from #685

PR #740, opened 2016-03-23
clicking search icon now closes the search tab

akhmorgan:
Closes [#707](https://github.com/ScottLogic/StockFlux/issues/707)

### DIFF
--- a/src/sidebars/search/search.html
+++ b/src/sidebars/search/search.html
@@ -1,6 +1,6 @@
 <div ng-keydown="searchCtrl.onSearchKeyDown($event)">
   <div>
-      <img src="assets/png/search_tab.png" class="top-icon" title="Search Stocks" draggable="false" />
+      <img src="assets/png/search_tab.png" ng-click="sidebarCtrl.toggleClick(); $event.stopPropagation();" class="top-icon" title="Search Stocks" draggable="false" />
       <input class="sidetab" type="text" maxlength="20" placeholder="Enter stock name or symbol"
              ng-show="sidebarCtrl.showSearches()" ng-model="searchCtrl.query"/>
       <icon name="close" icon-click="sidebarCtrl.favouritesClick()" tooltip="Close Search"></icon>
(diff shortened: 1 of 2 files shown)


src/sidebars/sidebar-controller.js
```diff
@@ -40,6 +40,10 @@
             }
             this._showSearches = false;
         }
+
+        toggleClick() {
+            this._showSearches = !this._showSearches;
```
owennw: Do we want to track this with the analytics too?

src/sidebars/search/search.html
```diff
@@ -1,6 +1,6 @@
 <div ng-keydown="searchCtrl.onSearchKeyDown($event)">
   <div>
-      <img src="assets/png/search_tab.png" class="top-icon" title="Search Stocks" draggable="false" />
+      <img src="assets/png/search_tab.png" ng-click="sidebarCtrl.toggleClick(); $event.stopPropagation();" class="top-icon" title="Search Stocks" draggable="false" />
```
owennw: I'm curious about what the `stopPropagation` is needed for?

WPFerg:
If the search icon collapses/expands, do we need/should we have the close button on compact view?

akhmorgan:
@wpferg an excellent question. One which the issue doesn't really specify. Might be worth asking Rui what he thinks. It's easy to remove the close icon in a later PR if needs be.

src/sidebars/search/search.html
akhmorgan: I'll double check this. I don't think it is actually necessary any more. I think it might just be left over from the first implementation that I tried without the `toggle` function.

akhmorgan: Actually this is necessary. Otherwise once the search is open, the click on the icon will be superseded by the click on search, so `searchClick` is called instead of `toggle`. This results in the icon no longer closing the search

src/sidebars/sidebar-controller.js
WPFerg: I'd probably say so. We're tracking the changes between search and favourites, and this is one of them

akhmorgan: So, do we want something like `reportAction('Toggle', 'Sidebar');` ?

Issue #729, opened 2016-03-23
Questions about the new favourites


owennw:
[ ] The card can be torn from the star icon -- is this okay?
[ ] Should the cursor over the card be different now (perhaps the hand?)
[ ] In compact view there's not much user feedback -- should we reinstate hover states?

rmarcalo:
These are really good points.

1) We can try to let the card be torn from the star icon, I'll just be interested to test if that doesn't affect clicking the star in any way. It probably won't.

2) I think the hand icon could be very appropriate for hovering the card, this hand is maintained when hovering the start icon, but since this icon has an hover state, the interaction should be clear. I would add the hover state to the star icon even with it is full (as the bordered style can really work both ways in terms of feedback).

3) I can imagine the compact view looking like it doesn't react much. We could could introduce the hover states, as you mentioned, but prevent it from being selected (as that would produce no effect).

PR #741, opened 2016-03-23
Cursor and star tweaks

owennw:
Response to the second point of #729

### DIFF
--- a/src/sidebars/star/star-controller.js
+++ b/src/sidebars/star/star-controller.js
@@ -24,10 +24,10 @@
         }
 
         favouriteUrl() {
-            if (this.stock.favourite) {
-                return starUrls.on;
-            } else if (this.starHovered) {
+            if (this.starHovered) {
                 return starUrls.onHover;
+            } else if (this.stock.favourite) {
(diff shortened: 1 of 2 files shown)


WPFerg:
LGTM. One question: On the card tearout, should the cursor revert back to default, or should it still be `pointer`?

owennw:
¯\_(ツ)_/¯

@rmarcalo ?

src/sidebars/star/star-controller.js
```diff
@@ -24,10 +24,10 @@
         }
 
         favouriteUrl() {
-            if (this.stock.favourite) {
+            if (this.starHovered) {
+                return starUrls.onHover;
```
WPFerg: Suspected line ending problem here

rmarcalo:
Not sure what you mean by tearout itself, when you click and drag?

WPFerg:
Correct

rmarcalo:
The change of cursor is there to let user know they can interact. When you click and drag I think it should be maintained. This cursor change is also something I would like to test, as it may become too prevalent.

PR #742, opened 2016-03-23
Don't persist empty closed windows

### LINKED ISSUE #573: Restore a window with no favourites
Should you be able to restore a closed window if it has no favourites?

owennw:
Closes #573

### DIFF
--- a/src/store-service.js
+++ b/src/store-service.js
@@ -92,10 +92,21 @@
         closeWindow() {
             this.store.closed = Date.now();
             this.save();
-            this.$rootScope.$broadcast('closeWindow');
+
+            if (this.store.stocks.length > 0) {
+                this.$rootScope.$broadcast('closeWindow');
+            }
+
+            var closedArray = storage.filter((store) => store.closed !== 0);
+
+            // Trim any stores without stocks
+            var emptyStores = closedArray.filter((store) => store.stocks.length === 0);
+            emptyStores.forEach((emptyStore) => {
+                var index = storage.indexOf(emptyStore);
+                storage.splice(index, 1);
+            });
 
             // Trim the oldest closed store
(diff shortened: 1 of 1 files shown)


WPFerg:
Is there a way to prevent the window from being added in the first place? It looks like it removes windows if they don't have stocks after the fact. If you have 5 closed windows, close an empty window, you'd only have 4 on your list

owennw:
Not really.

A window is added, with say 1 stock. That stock is removed. At that point we can choose to remove the window from the store, but it will still be open, so the store will not be an accurate representation of the current state of the application.

WPFerg:
What about when you set the `closed` flag?

owennw:
Also I'm not sure if you would only have 4, as the empty stores are removed before the trim for 5.

src/store-service.js
```diff
@@ -90,12 +90,24 @@
...
+            closedArray = storage.filter((store) => store.closed !== 0);
+            var emptyStores = closedArray.filter((store) => store.stocks.length === 0);
+            emptyStores.forEach((emptyStore) => {
+                var index = storage.indexOf(emptyStore);
+                storage.splice(index, 1);
+            });
 
             // Trim the oldest closed store
-            var closedArray = storage.filter((store) => store.closed !== 0);
+            closedArray = storage.filter((store) => store.closed !== 0);
```
WPFerg: Instead of re-setting `closedArray`, maybe it's worth doing something like:

```js
restorableWindows = storage.filter((store) => store.closed !== 0 && store.stocks.length > 0);
var cleanableWindows = storage.filter((store) => store.closed !== 0 && store.stocks.length === 0);
```

I'm not sure that's much better. It's still 2 iterations through `storage`

PR #743, opened 2016-03-23
Restore transition for compact favourites scrollbar

### LINKED ISSUE #703: "Transition" a bit slow on removal of scroll bar
When you hover over a list, if you wait for the scroll bar to appear and then hover out, the transition is really slow for the bar leaving. The movement of the text is really quick. The problem here is that you end up with the scroll bar on top of the text and it looks messy. Could we speed up the removal of the bar? Or wait until the bar is gone before moving the text back in line with the left edges of the stock tabs?
scrollbarspeed

mic-smith:
Closes #703 

Not sure why it was disabled before for compact?

### DIFF
--- a/src/sidebars/favourites/favourite-compact.less
+++ b/src/sidebars/favourites/favourite-compact.less
@@ -10,8 +10,4 @@
       }
     }
   }
-
-  .favourite .top {
-    transition: none;
-  }
 }


WPFerg:
`grunt ci` seems to work fine, happy to merge.

Issue #726, opened 2016-03-23
When you have one tab in a window, you should be able to drag it to another window


jackarnstein:
Dragging the last/only tab from one window to another window should be possible.
Unsure if you should then destroy the empty window or if it should sit there empty?

owennw:
How do we want this to work?

In Chrome with a single tab the entire window moves until it is dragged into the tab-bar where the whole window collapses to another tab (before releasing moving the mouse out of the tab-bar restores the window again).

Do we want to have the normal interaction, but only when the card is torn in the old window closes? Or the entire window moves? Or when the last card is torn out the old window shrinks/closes (what if the user changes their mind?)

@rmarcalo

rmarcalo:
My ideal for this was the following (and it might me way too complex):

A few seconds (1-2) after you detach and move a card away the entire window disappears. As the cursor moves around only the card is shown. When the card stops moving (after 1-2 seconds) the window appears again, aligning perfectly to the position of the card. At this point, if the user releases the card the window is placed. If moves again the window is hidden again. If the card is dragged into another window then it starts to belong there. The same would happen for compact view.

If this is too complicated:

A few seconds (1-2) after you detach and move a card away the entire window disappears. As the cursor moves around only the card is shown. It only reappears again if the card is dropped (without being included into another window).

owennw:
It's good to know the ideal, I think we an aim for that and see how it goes!

PR #747, opened 2016-03-24
Tearout single stocks WIP

owennw:
This PR introduces tearing out the final stock in a window. As per #726 the window will fade after an amount of time after the card has been torn and left the tearout area, and will fade back in if the card stops moving.

Issues/bugs:

- [ ] The window shadow is still present after the window has faded
- [ ] The window sometimes fades in and out twice, instead of once
- [ ] Quickly tearing out a card and letting go of the mouse will result in a window with the favourite missing (but it's there in search and the window behaves as if it has one stock!). This could potentially be resolved by not keeping the old window and instead destroying it and recreating a new window. This feels wasteful but would work. (There would be no fade effect either).

### DIFF
--- a/src/config-service.js
+++ b/src/config-service.js
@@ -30,7 +30,7 @@
                 contextMenu: allowContextMenu,
                 autoShow: false,
                 frame: false,
-                shadow: true,
+                shadow: false,
                 resizeRegion: {
                     size: 7,
                     topLeftCorner: 14,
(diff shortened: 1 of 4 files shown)


owennw:
An idea I had to make this a bit neater is to extract all the timer code into its own class/component.

mic-smith:
I've a fix for the window with the favourites missing. Looking at other issues

akhmorgan:
This looks pretty good to me, in terms of functionality. I've tested it locally, and I haven't seen any issues.

akhmorgan:
Code looks fine to me.

mic-smith:
Shadows have been turned off and I think other issues are now fixed. Will merge

Issue #716, opened 2016-03-22
Stock card autoexpands when dragging on some screens


bjedrzejewski:
On some screens (Lenovo Yoga) the stock card just keeps expanding. This makes it impossible to demo on some of our sales people laptops!

![image](https://cloud.githubusercontent.com/assets/9947570/13950666/471bad6c-f024-11e5-9745-0841b8d55c91.png)

It happens with the touch screen on and off.

bjedrzejewski:
It does work on MS Surface. So it is not strictly touchscreens related.

WPFerg:
From the config file:
```
maxWidth: 230,
maxHeight: 100,
```
Weird.

bjedrzejewski:
The problem gets worse when you move the mouse- any ideas? What happens on drag?

Also, we replicated it on another Lenovo Yoga.

bjedrzejewski:
Not sure if related, I found that on localhost:9090

TypeError: Cannot read property 'open' of undefined
    at app.js:1
    at whenReady (<anonymous>:2543:9)
    at Object.fin.desktop.main (<anonymous>:2601:5)
    at Object.d [as ready] (app.js:1)
    at a._createClass.value (app.js:1)
    at app.js:1
    at m.$digest (angular.min.js:139)
    at m.$apply (angular.min.js:142)
    at g (angular.min.js:94)
    at x (angular.min.js:98)(anonymous function) @ angular.min.js:114

akhmorgan:
Tested on the Lenovo Yoga, and the issue appears to be tied to the display settings. The System display setting for the size of text, apps and other items is set to 150% as the default, recommended setting. Changing this to 100% removes the issue.

akhmorgan:
Adjusting display settings on a desktop to change the size of text and other items to 150%, produces the same bug.

akhmorgan:
The issue is not present at 125% scale.

akhmorgan:
The issue appears to be related to the use of `window.moveTo`

akhmorgan:
A stopgap fix has been implemented by repeatedly setting the bounds on the tearout, pending further investigation into the root cause of the issue, which may be outside of the scope of this app.

PR #753, opened 2016-03-31
modifies tearout to use setBounds rather than moveTo

akhmorgan:
Overcomes issue [#716](https://github.com/ScottLogic/StockFlux/issues/716)

This is not a nice fix. However, as the root cause of the bug remains unknown, this is an effective workaround.

### DIFF
--- a/src/sidebars/favourites/tearout-directive.js
+++ b/src/sidebars/favourites/tearout-directive.js
@@ -30,6 +30,7 @@
                             myHoverArea = parent.getElementsByClassName('drop-target')[0],
                             mouseOffset = { x: 0, y: 0 },
                             elementOffset = { x: 0, y: 0 },
+                            previousOffset = { x: 0, y: 0 },
                             currentlyDragging = false,
                             dragService,
                             dragTimeout;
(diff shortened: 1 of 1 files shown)


src/sidebars/favourites/tearout-directive.js
```diff
@@ -162,10 +173,12 @@
                             var _window = currentWindowService.getCurrentWindow();
 
                             var newCardOffset = configService.getTopCardOffset(store.isCompact());
-                            moveWindow(
-                                _window,
-                                currentMousePosition.x - newCardOffset[0] - 2,
-                                currentMousePosition.y - newCardOffset[1] - 1);
+                            _window.setBounds(
```
mic-smith: Could this still be a call to moveWindow? 
It looks like the window doesn't change size and this would keep the moveTo / setBounds switch isolated to one function.

src/sidebars/favourites/tearout-directive.js
```diff
@@ -272,9 +285,14 @@
 
                                             windowService.createMainWindow(null, compact, (newWindow, showFunction) => {
                                                 reportAction('Tearout', 'Created ' + scope.stock.code);
-                                                newWindow.resizeTo(window.outerWidth, window.outerHeight, 'top-left');
                                                 var newCardOffset = configService.getTopCardOffset(compact);
-                                                moveWindow(newWindow, e.screenX - newCardOffset[0], e.screenY - newCardOffset[1], showFunction);
+                                                newWindow.setBounds(
```
mic-smith: Same here - could this still be moveWindow?

Issue #595, opened 2016-03-11
Create a test 'environment'


ColinEberhardt:
Currently all of our releases are immediately public (hooray for continuous delivery!). We need a mechanism that allows us to distribute multiple versions of the application so that we can do some testing on a release before pushing it public.

bjedrzejewski:
@ColinEberhardt can't we use the release candidate mechanism? This automatically appears on gh-pages and is ready for testing.

bjedrzejewski:
You can have for example:

https://github.com/ScottLogic/StockFlux/blob/gh-pages/10.0.0-rc.2/StockFlux-10.0.0-rc.2.zip

Pointing to the `10.0.0-rc.2`

mic-smith:
Release candidate and dev versions available through gh-pages for testing before merging to master.

PR #757, opened 2016-04-05
Prevent gh-pages version overwrites from dev build. 

### LINKED ISSUE #713: Fix gh-pages version update to prevent dev and master build overwrites
Currently deploy.sh is generating the version directory based on the package.json for both dev and master. It does not distinguish between them so it can overwrite when they are set to the same version.

We need a prevent dev builds overwriting an existing master version. Possible options:
- Prefix dev versions
- Add a process to bump dev to the next snapshot version as part of release
- Don't build version directory on dev build.

mic-smith:
Updated deploy script to publish to gh pages the following:

| Branch | gh-pages version  |
|--- |---|
|master| master and _version_ |
| dev | dev |
| release-* |  _version_-rc |

_version_ is from package.json

Closes #713 #595


akhmorgan:
looks good to me

PR #761, opened 2016-04-06
Fixes issue with positioning of new windows on expansion from tearout

akhmorgan:


mic-smith:
Tested and looks good functionally. In terms of code, could do with moving the x, y adjustments into one place. However, it would be good to try and simplify the tearout code generally. We can do that on dev as a separate change.

PR #771, opened 2016-04-08
Fixes race condition

### LINKED ISSUE #769: Occasionally the favourites display as "loading favourites" on app startup
This seems to be a race condition. Only been able to reproduce on the gh-pages deployed versions.

Switching between versions master -> dev etc occasionally triggers. 
Also just restarting the 10.1-rc version on a slow machine occasionally fails.

Adding a favourite via the search then causes the rest to display correctly.

mbssantos:
Fixes #769 by sorting out the race condition.
A test version can be downloaded [here](http://mbssantos.github.io/StockFlux/dev_restart2/install.html)


mic-smith:
Not been able to reproduce issue with test version - looks good

PR #803, opened 2016-05-13
Moved actions to enum file

rjmcneill:


src/child/actions/sidebar.js
```diff
@@ -1,80 +1,73 @@
...
-export const SEARCH_INPUT = 'SEARCH_INPUT';
-export const SEARCH_ERROR = 'SEARCH_ERROR';
-export const CLEAR_SEARCH = 'CLEAR_SEARCH';
-export const SEARCH_FINISHED = 'SEARCH_FINISHED';
-export const FAV_CLICKED = 'FAV_CLICKED';
-export const TOGGLE_FAVOURITE = 'TOGGLE_FAVOURITE';
-export const SELECTION = 'SELECTION';
-export const UNSELECT = 'UNSELECT';
+
+import ACTIONS from '../enums/actions.js';
```
jleft: Not sure, but if the constants are to be stored in nested objects, then I'd be tempted to do the following, so to not import constants which aren't required here and allows you to be a bit more terse when using them (e.g. `ACTIONS.SIDEBAR.SEARCH_INPUT` vs `ACTIONS.SEARCH_INPUT`).

```js
import { SIDEBAR as ACTION_TYPES } from '../enums/actions.js';

//...

ACTION_TYPES.SOME_ACTION
```

src/child/enums/actions.js
```diff
@@ -0,0 +1,26 @@
+
```
jleft: I'm not sure if this might be better in a _constants_ directory, rather than _enums_?

Just curious, but I noticed new blank lines at the top of some of these files - not sure why?

src/child/enums/actions.js
```diff
@@ -0,0 +1,26 @@
...
+const ACTIONS = {
```
jleft: Please correct me if I'm wrong, but from what I've read - an action is a plain JS object, with a `type` property - I think these are _action types_, not _actions_.

PR #804, opened 2016-05-13
Added favourites listing when not searching

rjmcneill:


src/child/containers/sidebars/search/Search.js
```diff
@@ -64,11 +65,29 @@ class Search extends Component {
...
+
+                        {isSearching &&<div className="loading-message results-message">Loading search results...</div>}
+
+                        {!isSearching && !results && favourites && favourites.length > 0 && (favourites || []).map(stockCode =>
+                            <Favourite
+                                key={stockCode}
+                                stockCode={stockCode}
+                                bindings={bindings}
+                                selected={stockCode === selection.code}
+                                isFavourite={favourites.indexOf(stockCode) >= 0}
```
mbssantos: Do we still need the indexOf here?

src/child/containers/sidebars/search/Search.js
```diff
@@ -36,12 +37,12 @@ class Search extends Component {
...
-    onIconClick(stock) {
-        this.props.dispatch(toggleFavourite(stock.code));
+    onIconClick(stockCode) {
+        this.props.dispatch(toggleFavourite(stockCode));
     }
 
-    onClick(stock) {
-        this.props.dispatch(selectStock(stock.code, stock.name));
+    onClick(stockCode, stockName) {
+        this.props.dispatch(selectStock(stockCode, stockName));
```
mbssantos: Might be worthwhile to have a look at keeping the name somewhere and only pass codes around.

src/child/containers/sidebars/search/Search.js
mic-smith: We don't need this prop, the current Favourite propTypes are wrong. We should remove the isFavourite prop from the Favourite component propTypes and then remove the usage in Favourites.js. Looks like we'd based it on the search prop types at some point.

Also the favourites should get shown in the style of a search result (Just code, name and star), when shown here.

src/child/containers/sidebars/favourites/Favourites.js
```diff
@@ -41,6 +42,10 @@ class Favourites extends Component {
         // TODO: send content to tearout
     }
 
+    onQuandlResponse(stockCode, stockName) {
+        this.props.dispatch(quandlResponse(stockCode, stockName));
```
mbssantos: Since we now have the codes and names stored, we can afford to only pass codes around.

PR #805, opened 2016-05-13
Initial unit test setup

jleft:
```sh
> mocha --compilers js:babel-register --recursive

  shared/versionValue
  ✓ should be a valid semantic version

1 passing (12ms)
```


mic-smith:
Thanks @jleft  Looks good to me

PR #806, opened 2016-05-13
Add transition to search text

jdhodges:
Adds a transition to the search input when the search tab is expanded/collapsed.


mbssantos:
Can we use the Sidebar's "expanded' || 'contracted' classes instead of revealing it to their children?

jdhodges:
Should I pass this to search as a property? e.g.
`<Search isContracted={!sidebar.showSearch} />`
Or do you mean something else?

mbssantos:
I meant this logic seems similar to what's in the [Sidebar](https://github.com/ScottLogic/StockFlux/blob/react_redux/src/child/containers/sidebars/Sidebar.js) - see line 33 && 36.
Feels like we could have the css inherit from it.

PR #807, opened 2016-05-16
fix: selectors now clickable above showcase div

rjmcneill:


mbssantos:
Editing css that doesn't belong to us can lead to problems, can we sort this in our css?
display-block on our div seems to work.

Issue #770, opened 2016-04-08
It's possible to tear in to a window that is currently behind other windows


akhmorgan:
This seems a little bit weird.

Steps to reproduce:
1. create 2 windows
2. put an application in front of just one of the windows (let's call that window B). So we have one window still visible, window A, and one window which is now hidden behind some other app, window B.
3. Drag a favourite from the displayed window, window A, over to where window B would be if it were not hidden behind another application. Specifically aim for where the sidebar would be.
4. Release the tearout.

Observed behaviour:
Instead of getting a new window, the tearout will snap into window B and bring it to the front of the desktop.

I'm not sure what the desired behaviour is here exactly, but I think that if I were to release a tearout, seemingly not over any other StockFlux windows, then I would expect a new window to be created.

akhmorgan:
[StockFlux_Issue_770.zip](https://github.com/ScottLogic/StockFlux/files/210305/StockFlux_Issue_770.zip)

msuperina:
Solved by #809

Issue #534, opened 2016-03-04
Drag search result into favourite tab to favourite


owennw:

rmarcalo:
This would be a "hidden" nice to have. It is interesting as a playful interaction.

msuperina:
Solved by #809

PR #809, opened 2016-05-16
feat: Drag and drop

mbssantos:
Drag and drop functionality.
Allows user to reorder tiles and move them across windows without tearouts.
Annuls #770 and #534


jdhodges:
Really nice work, just something I noticed. The reorder tiles feature doesn't seem to be working correctly.

Here I'm taking the top tile and dragging it to the position under TSLA:
image

Dropping it here results in:
image

Two things to point out:
1) MSFT should now be in the second position.
2) There should only be one 'empty' space during the dnd operation (the destination of the dragged tile). The other tiles should transition up to fill the space left by the dragged tile.

mbssantos:
Addressed those two issues.
For future reference, this already inlcudes cross window drag & drop

jleft:
:+1:

src/child/containers/sidebars/favourites/Favourites.js
```diff
@@ -38,14 +58,56 @@ class Favourites extends Component {
         this.props.dispatch(selectStock(stockCode, stockName));
     }
 
-    onDrag() {
-        // TODO: send content to tearout
+    onDragOverFavourite(e, targetCode) {
+        const codes = this.props.favourites.codes;
+        const code = getCodeFromDT(e.dataTransfer.types);
+        // const index = codes.indexOf(targetCode);
```
jdhodges: Can we remove this?

rjmcneill:
:+1:

src/child/components/Favourite.js
```diff
@@ -18,6 +17,7 @@ class Favourite extends Component {
 
     componentDidMount() {
         const stockCode = this.props.stockCode;
+        this.addDragTarget(stockCode);
```
mic-smith: Should remove the event listeners created in this call in componentWillUnmount

jleft:
Merged with the intention that the 2 issues listed above will be addressed separately (removing redundant comment (https://github.com/ScottLogic/StockFlux/pull/809#discussion_r64353516) and removing event listeners (https://github.com/ScottLogic/StockFlux/pull/809#discussion_r64578393))

PR #810, opened 2016-05-16
feat: defaults to new favourite when unfavouriting a selected favourite

rjmcneill:


src/child/components/Favourite.js
```diff
@@ -35,6 +35,15 @@ class Favourite extends Component {
     onIconClick(e) {
         e.stopPropagation();
         this.props.bindings.onIconClick(this.props.stockCode);
+
+        // Select a new stock to display
+        for(var i = 0; i < this.props.favourites.codes.length; i++) {
```
jdhodges: Could we use `forEach` here instead?

src/child/containers/sidebars/favourites/Favourites.js
```diff
@@ -88,6 +88,7 @@ class Favourites extends Component {
                               stockCode={stockCode}
                               bindings={bindings}
                               selected={stockCode === selection.code}
+                              favourites={favourites}
```
mbssantos: Exposing the Favourites to the Favourite seems excessive, can we move the code to the toggleFavourite fn (and possibly rename it)?

PR #811, opened 2016-05-17
chore: simplified dev setup

ColinEberhardt:
Just a few simple changes that hopefully make it easier to get the dev build running

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ This is an 'evergreen' application, each time it launches the application code i
 
 # Development
 
-The application is built using AngularJS and ES2015, transpiled via Babel. The charts are rendered using [d3fc](https://d3fc.io/), a Scott Logic open source project which provides a number of components that allow the creation of bespoke interactive charts. The bulk of the charting code is adapted from [BitFlux](http://scottlogic.github.io/BitFlux/), which showcases the capabilities of d3fc.
+The application is built using React and ES2015, transpiled via Babel. The charts are rendered using [d3fc](https://d3fc.io/), a Scott Logic open source project which provides a number of components that allow the creation of bespoke interactive charts. The bulk of the charting code is adapted from [BitFlux](http://scottlogic.github.io/BitFlux/), which showcases the capabilities of d3fc.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -29,18 +29,27 @@ The displayed data is real and provided by [Quandl](https://www.quandl.com). The
 
 - Download or clone this repository locally
 - Ensure [Node.js](https://nodejs.org/), which includes npm, is installed
-- Ensure [Grunt](http://gruntjs.com/getting-started#installing-the-cli) is installed:
+
+
+Install dependencies:
 
 ```
-npm install -g grunt-cli
```
jleft: Grunt is still required for `postinstall`, when building BitFlux. (If this was removed from the Travis config, then the build would break; I think this is something which we could change in BitFlux to remove the dependency from StockFlux.)

openfin-launcher.js
```diff
@@ -0,0 +1,11 @@
+const openfinLauncher = require('openfin-launcher');
+
+openfinLauncher.launchOpenFin({
+    configPath: 'http://localhost:5000/app.json'
+})
+.then(() => {
+    console.log('success!');
+})
+.fail((error) => {
+    console.log('error!', error);
```
jleft: Nitpick: `console.error` ?

jleft:
Thanks @ColinEberhardt, I think this will make it a lot easier to get the dev build running :smile:

README.md
ColinEberhardt: As long as we keep grunt as a dev dependencies, the build should work just fine without having to ask the user to install it as a global module.

This is because npm adds `node_modules/.bin` to the path when running scripts:

https://docs.npmjs.com/misc/scripts#path

This is a great reason for adding npm scripts to run your build, tests, dev-servers etc ...

openfin-launcher.js
ColinEberhardt: good point, I'll update

ColinEberhardt:
@jleft - updated based on your feedback

jleft:
LGTM 👍  Thanks @ColinEberhardt

PR #814, opened 2016-05-18
feat: add compact view functionality

jdhodges:
This doesn't include 'double click favourite to open full view'.


src/child/containers/toolbar/Toolbar.js
```diff
@@ -37,10 +37,12 @@ class Toolbar extends Component {
     }
 
     onCompactClick() {
+        fin.desktop.Window.getCurrent().resizeTo(230, 500, 'top-right');
```
jleft: I'd use `ConfigService` to get these dimensions, so you DRY.

src/child/containers/toolbar/Toolbar.js
```diff
@@ -37,10 +37,12 @@ class Toolbar extends Component {
...
         this.props.dispatch(compact());
```
jleft: Unsure about this, so please correct me if I'm wrong, but it seems odd to resize the window, then dispatch an action. It feels like the resize is a side-effect.

I think using an async action creator or subscribing a listener on the store to call the resize method would be more appropriate.

I'm not sure which, but I'm leaning towards an async action creator. Assuming the `compact` and `expand` actions will always lead to a resize and as `resizeTo` is aysnc and will either succeed or fail, the async action creator seems like a good fit. For example, at the moment if `resizeTo` failed in `onCompactClick()` the state would still be updated to suggest that the window is in compact view, when it's not.

Any thoughts?

src/child/containers/App.js
```diff
@@ -18,13 +18,13 @@ require('script!../../../node_modules/BitFlux/node_modules/bootstrap/js/dropdown
...
-const App = ({ code, name }) => (
-    <div className="main">
+const App = ({ code, name, windowState }) => (
+    <div className={`main${windowState.isCompact ? ' compact' : ''}`}>
         <SideBar />
         <div className="main-content">
             <Toolbar />
             <img className={`no-favourites-image${code ? ' hidden' : ''}`} src={noFavourites} />
-            <div className={code ? '' : 'hidden'}>
+            <div className={code && !windowState.isCompact ? '' : 'hidden'}>
```
jleft: If this needs to be hidden when the app is in compact view, can this be done in the styles? (Not sure if `.right-content` (from main-compact.less) is appropriate? If not, then a new class like `.showcase` or `.chart` could work?)

```less
.compact .someClass {
  display: none;
}
```

src/child/containers/main/main-compact.less
```diff
@@ -0,0 +1,3 @@
+.compact .right-content {
```
jleft: Looks like this is unused?

src/child/containers/toolbar/Toolbar.js
```diff
@@ -78,7 +78,7 @@ class Toolbar extends Component {
         const { windowState } = this.props;
         return (
             <div className="toolbarWrapper drag">
-                <img id="logo" src={icon} className="logo" alt="Scott Logic" />
+                <img id="logo" src={icon} className={`logo${windowState.isCompact ? ' hidden' : ''}`} alt="Scott Logic" />
```
jleft: Similar comment to above, about hiding the chart - if this needs to be hidden, can it be done in the styles?

src/child/actions/window.js
```diff
@@ -43,3 +44,21 @@ export function close() {
         type: ACTION_TYPES.CLOSE
     };
 }
+
+export function resizeToCompact() {
```
jleft: :+1: this looks much nicer

src/child/actions/window.js
```diff
@@ -43,3 +44,21 @@ export function close() {
...
+    return dispatch => {
+        const compactWindowDimensions = configService.getCompactWindowDimensions();
+        fin.desktop.Window.getCurrent().resizeTo(compactWindowDimensions[0], compactWindowDimensions[1], 'top-right', () => {
+            dispatch(compact());
```
jleft: Probably don't need to export `compact()` and `expand()` now that we have `resizeToCompact()` and `resizeToDefault()`?

jdhodges:
image

Sometimes bitflux isn't rendering properly so looks like this is going to need some more work. I'm guessing this is a race condition that has been introduced after adding the async actions. I may need to add an intermediate step to manage the state while the request is being made, in the same way `isFetching` is used in the example [here](http://redux.js.org/docs/advanced/AsyncActions.html).

src/child/actions/window.js
```diff
@@ -43,3 +50,23 @@ export function close() {
         type: ACTION_TYPES.CLOSE
     };
 }
+
+export function resizeToCompact() {
+    return dispatch => {
+        dispatch(resizing());
+        const compactWindowDimensions = configService.getCompactWindowDimensions();
+        fin.desktop.Window.getCurrent().resizeTo(compactWindowDimensions[0], compactWindowDimensions[1], 'top-right', () => {
```
jleft: Also worth handling the error if `resizeTo` fails?

jdhodges:
This now triggers a `hasErrors` flag in `windowState` if the resize fails. This isn't reported to the user yet - we may want to show a toast style message in the future.

PR #815, opened 2016-05-23
feat: persist state in local storage

jdhodges:
Added the [singleton](https://github.com/webpack/style-loader#singleton) option to the webpack style loader as BitFlux wasn't rendering correctly (when loaded immediately on start). This broke the fonts so I ended up importing them. Happy to change if anyone knows a better way.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "react-dom": "^15.0.1",
     "react-redux": "^4.4.4",
     "redux": "^3.4.0",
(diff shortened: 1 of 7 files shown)


rjmcneill:
This looks good :+1: .
One future enhancement we could make would be to store the indicators in the state as well.
It would probably involve some messing about since that section is controlled by BitFlux, but it would look good if it worked.

jdhodges:
Closing as this is superseded by #833

PR #823, opened 2016-05-26
refactor: ensure event handlers are removed when unmounting favourite

jdhodges:
Fixes https://github.com/ScottLogic/StockFlux/pull/809#discussion_r64578393

This uses react's synthetic events, so now the event handlers are removed automatically when a favourite is unmounted. See [here](https://facebook.github.io/react/docs/interactivity-and-dynamic-uis.html#under-the-hood-autobinding-and-event-delegation).

### DIFF
--- a/src/child/components/Favourite.js
+++ b/src/child/components/Favourite.js
@@ -12,11 +12,14 @@ class Favourite extends Component {
     constructor(props) {
         super(props);
         this.onIconClick = this.onIconClick.bind(this);
+        this.onDragStart = this.onDragStart.bind(this);
+        this.onDragOver = this.onDragOver.bind(this);
+        this.onDragLeave = this.onDragLeave.bind(this);
+        this.onDragEnd = this.onDragEnd.bind(this);
     }
 
     componentDidMount() {
         const stockCode = this.props.stockCode;
-        this.addDragTarget(stockCode);
         quandlServiceGetStockData(stockCode)
             .then(response => {
                 const data = response.stockData.data[0];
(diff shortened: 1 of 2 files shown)


jleft:
Thanks @jdhodges, generally this looks okay to me, a couple of comments/questions:

+ I think you'll need need to use `bind` to preserve the context in the event listeners, e.g
```js
this.onIconClick = this.onIconClick.bind(this);
```
+ Have you investigated the performance impact? I think this could potentially be a fairly sensitive area.

---

@mbssantos - just wondering if you tried React's synthetic events and, if so, there are any drawbacks which we should be aware of?

jdhodges:
Do we need to bind the event listeners since none of them require a reference to the component, or is it good practice to preserve the context?

I haven't investigated the performance, but from what I can see in this [blog post](https://davidwalsh.name/event-delegate), it seems it may be better to pull all of these event listeners up to `Favourites`.
 
The advantage of this is we'd only need one 'set' of event listeners on the parent (rather than a set for each individual Favourite), on the other hand it isn't very consistent with our approach elsewhere and react could already be doing this automatically since it controls the internal event handler mapping.

jleft:
> Do we need to bind the event listeners since none of them require a reference to the component, or is it good practice to preserve the context?

I think it's a good practice as, rather than depending on the context from which the function was called, it's makes it clear what `this` will be (even if it's not used now, it might be in the future). It's also consistent with all/most of the other event listeners.

React will usually do this automatically, using the [autobinding feature](https://facebook.github.io/react/docs/reusable-components.html#no-autobinding) (also linked to above), but this doesn't apply when using ES6 class syntax, so it has to be done manually.

> **Autobinding:** When creating callbacks in JavaScript, you usually need to explicitly bind a method to its instance such that the value of this is correct. With React, every method is automatically bound to its component instance (except when using ES6 class syntax). React caches the bound method such that it's extremely CPU and memory efficient. It's also less typing!

[The recommended way is to do the binding in the components constructor](https://facebook.github.io/react/docs/reusable-components.html#no-autobinding):

> We recommend that you bind your event handlers in the constructor so they are only bound once for every instance:
> 
> ```js
> constructor(props) {
>   super(props);
>   this.state = {count: props.initialCount};
>   this.tick = this.tick.bind(this);
> }
> ```


---

> I haven't investigated the performance, but from what I can see in this blog post, it seems it may be better to pull all of these event listeners up to Favourites.

> The advantage of this is we'd only need one 'set' of event listeners on the parent (rather than a set for each individual Favourite), on the other hand it isn't very consistent with our approach elsewhere and react could already be doing this automatically since it controls the internal event handler mapping.

Thanks for explanation - that sounds good. Is it worth moving the event listeners in the _Favourites_ container to synthetic events too?

jdhodges:
I've had a go at using event delegation, but the results aren't pretty - we'd end up with our events looking like this:
```
    function stockCodeAccessor(elem) {
        return elem.getElementsByClassName('code')[0].textContent;
    }
    
    function stockNameAccessor(elem) {
        return elem.getElementsByClassName('name')[0].textContent;
    }

    onClick(e) {
        if (e.target && e.target.className === 'favouriteWrapper') {
            this.props.bindings.selectStock(stockCodeAccessor(e.target), stockNameAccessor(e.target));
        }
    }
```
Seems like premature optimisation to me.

Since we're not dealing with hundreds of favourites at once I think we'll be fine with our current approach.

src/child/containers/sidebars/favourites/Favourites.js
```diff
@@ -96,29 +98,23 @@ class Favourites extends Component {
...
-            if (!dropTarget.classList.contains('dragOver')) {
-                dropTarget.classList.add('dragOver');
-            }
-            e.preventDefault();
-        }, false);
+    onDrop(e) {
+        const codes = this.props.favourites.codes;
+        const code = e.dataTransfer.getData('text/plain');
+        this.props.dispatch(insertFavouriteAt(codes.length, code));
+        e.target.classList.remove('dragOver');
```
jleft: [`e.target`](https://developer.mozilla.org/en-US/docs/Web/API/Event/target) is not equivalent to the `dropTarget` in the previous implementation - I'm not sure if this is the intention? I think that `this` ([in this case: `this === e.currentTarget`](https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget)) would be more appropriate and consistent here, unless there is a reason to change?

This [StackOverflow answer](http://stackoverflow.com/a/21667010) explains the difference between `event.target` and `this`/`event.currentTarget` well.

src/child/components/Favourite.js
```diff
@@ -38,36 +41,29 @@ class Favourite extends Component {
...
     }
 
-    addDragTarget(stockCode) {
-        const dragTarget = document.getElementById(`stock_${stockCode}`);
-        dragTarget.addEventListener('dragstart', e => {
+    onDragStart(stockCode) {
+        return e => {
             // TODO: fade out window if it's last stock
-            dragTarget.classList.add('dragging');
+            e.target.classList.add('dragging');
```
jleft: Same comment about `this` vs `e.target` as made on the Favourites container.

jleft:
Thanks for the updates @jdhodges - this is looking good :+1: Just a couple more comments above, then I think this is good to merge.

src/child/components/Favourite.js
jleft: Has this not been addressed because using `e.target` is the intention (over `this`)? If so, what is the reasoning behind this?

PR #831, opened 2016-05-31
test: add tests for favourites reducer

jleft:


src/child/reducers/favourites.js
```diff
@@ -32,7 +32,7 @@ export default function favourites(state = { codes: [], names: {} }, action) {
         } else {
             codes.push(action.code);
         }
-        return Object.assign({}, state, { codes, move: {} });
```
rjmcneill: Yup, looks like leftover code from favourites d&d

PR #833, opened 2016-06-01
Added parent store and local storage

rjmcneill:
Supersedes #815


rjmcneill:

PR #836, opened 2016-06-07
feat: defaults to new favourite when unfavouriting a selected favouri…

rjmcneill:
…te in the search component

### DIFF
--- a/src/child/containers/sidebars/favourites/Favourites.js
+++ b/src/child/containers/sidebars/favourites/Favourites.js
@@ -2,7 +2,7 @@
 import React, { Component, PropTypes } from 'react';
 import { connect } from 'react-redux';
 
-import { selectStock, unselectStock, toggleFavourite, quandlResponse, insertFavouriteAt } from '../../../actions/sidebar';
+import { selectStock, quandlResponse, insertFavouriteAt } from '../../../actions/sidebar';
 import favTabImage from '../../../assets/png/favourites_tab.png';
 import Favourite from '../../../components/Favourite.js';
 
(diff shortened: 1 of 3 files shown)


src/child/containers/sidebars/search/Search.js
```diff
@@ -38,6 +38,15 @@ class Search extends Component {
 
     onIconClick(stockCode) {
         this.props.dispatch(toggleFavourite(stockCode));
+        if (this.props.selection.code === stockCode) {
```
jleft: This looks identical to the implementation in the Favourites component - can the duplication be removed/reduced?

src/child/containers/sidebars/favourites/Favourites.js
```diff
@@ -109,15 +109,7 @@ class Favourites extends Component {
 
     toggleFavourite(stockCode) {
```
jleft: I know this is existing code, but I find it quite confusing that `Favourites` has `toggleFavourite()` and `Search` has `onIconClick()` which do (essentially) the same thing - perhaps they can be renamed to be consistent? `onIconClick` seems the most consistent to me.

src/child/containers/sidebars/favourites/Favourites.js
```diff
@@ -109,15 +109,7 @@ class Favourites extends Component {
...
         this.props.dispatch(toggleFavourite(stockCode));
```
jleft: Does it make sense to also pull this up to `selectNewFavourite` in the Sidebar container? (If so, I'd probably also rename `selectNewFavourite`, maybe to `toggleFavourite`?)

jleft:
Thanks @rjmcneill LGTM :+1:

PR #837, opened 2016-06-07
Spawn new window on stock drag out

rjmcneill:
Blocked by #833 (built on top of that PR)


jdhodges:
We could do with the favourites tab being open when a new window is spawned. Also, it doesn't seem to work when pulling the tile outside of the window, would this be easy enough to do?

rjmcneill:
Favourite tab now open when a new window is spawned

rjmcneill:

PR #838, opened 2016-06-08
refactor: update Quandl service to return promises and export functions

jleft:
and add some initial tests


src/child/components/Favourite.js
```diff
@@ -18,18 +17,19 @@ class Favourite extends Component {
...
-            };
-            const chartData = response;
-            this.props.bindings.onQuandlResponse(stockCode, response.dataset.name);
-            this.setState({ stockData, chartData });
-        });
+        quandlServiceGetStockData(stockCode)
+            .then(response => {
+                const data = response.stockData.data[0];
+                const stockData = {
+                    name: response.dataset.name,
```
rjmcneill: Not sure if its worth it, but could pull this out:
```JavaScript
const stockName = response.dataset.name
```

It would just keep it consistent with how we're passing stockCodes and names around below as well

```JavaScript
this.props.bindings.onQuandlResponse(stockCode, stockName);
```

jleft: That looks neater to me - done :)

PR #841, opened 2016-06-08
feat: add confirmation box when unfavouriting

jdhodges:


src/child/components/Favourite.js
```diff
@@ -86,18 +84,39 @@ class Favourite extends Component {
...
               onClick={() => bindings.onClick(stockCode, name)}
               onDoubleClick={() => bindings.onDoubleClick(stockCode, name)}
             >
                 <div className="drop-target">
                     <div className={`darkens favourite tearable ${cls}`} draggable="false">
                         <div className="top">
-                            <div className="button-icon star active" onClick={this.onIconClick}>&nbsp;</div>
+                            <div ref="star" className="button-icon star active" onClick={bindings.onIconClick(stockCode)}>&nbsp;</div>
+                            {isUnfavouriting ?
+                                <Confirmation
```
rjmcneill: Is it possible to tidy up this section by passing functions down through bindings just?
I like that the `Confirmation` component is kept dumb, but it makes the `Favourite` looks pretty messy in my opinion

src/child/components/Favourite.js
```diff
@@ -71,7 +61,15 @@ class Favourite extends Component {
     }
 
     render() {
-        const { stockCode, selected, bindings } = this.props;
+        const { stockCode, selected, bindings, isUnfavouriting } = this.props;
+
+        const { star } = this.refs;
+
+        const modalHeight = 84;
```
rjmcneill: If these need to be declared here, might be worth sticking them at the top of the file?

src/child/components/Confirmation.js
```diff
@@ -0,0 +1,27 @@
+import React, { PropTypes } from 'react';
+
+const Confirmation = props => (
+    <div className="confirmation-backdrop" onClick={props.onBackdropClick}>
+        <div className="bubble-head" style={{ top: props.bubbleHeadTop() }}></div>
+        <div className={`confirmation-modal${props.isFlipped ? ' flipped' : ''}`} style={{ top: props.modalTop() }}>
+            <div className="confirmation-content" onClick={e => e.stopPropagation()}>
```
rjmcneill: I'd probably add this to the bindings if you're taking that approach so we're not declaring the functions in line

src/child/containers/sidebars/star/star.less
```diff
@@ -0,0 +1,54 @@
+.star {
```
jleft: There is already a `star` class in `favourites.less` for this, can this style (in `star.less`) be removed (and the class in `favourites.less` updated, if required)?

src/child/containers/sidebars/star/star-compact.less
```diff
@@ -0,0 +1,9 @@
+.compact {
```
jleft: I don't think `star` makes sense much now, I understand it was there in the Angular version, but as we don't have a star component/container, I think it makes sense to move this to be with the relevant component.

How about rename `star-compact.less` => `unfavourite-confirmation-compact.less` and `star.less` => `unfavourite-confirmation.less`?

They could all be in the same dir in components, i.e.:

```
child
    - components
        - favourite
             - Favourite.js
             - UnfavouriteConfirmation.js
             - unfavourite-confirmation-compact.less
             - unfavourite-confirmation.less
```

src/child/components/Confirmation.js
```diff
@@ -0,0 +1,24 @@
+import React, { PropTypes } from 'react';
+
+const Confirmation = props => (
```
jleft: Maybe `UnfavouriteConfirmation`?

src/child/containers/sidebars/favourites/Favourites.js
```diff
@@ -33,14 +33,18 @@ class Favourites extends Component {
...
         this.onQuandlResponse = this.onQuandlResponse.bind(this);
+        this.onConfirmClick = this.onConfirmClick.bind(this);
+        this.onBackdropClick = this.onBackdropClick.bind(this);
+        this.onIconClick = this.onIconClick.bind(this);
         this.onDoubleClick = this.onDoubleClick.bind(this);
         this.onDrop = this.onDrop.bind(this);
         this.onDragOver = this.onDragOver.bind(this);
         this.onDragLeave = this.onDragLeave.bind(this);
+
+        this.state = { unfavouriting: null };
```
jleft: `unfavouriting` sound like a boolean to me, how about `unfavouritingStockCode`?

src/child/components/Favourite.js
```diff
@@ -76,11 +91,33 @@ class Favourite extends Component {
...
+    onModalClick(e) {
+        e.stopPropagation();
+    }
+
+    onBackdropClick(e) {
+        this.setState({ isHovered: false });
+        this.props.bindings.onBackdropClick(e);
+    }
+
+    flipModal(starTopPos) {
```
jleft: I'm probably just being slow, but it took me a while to work out what this means... does `shouldPositionAboveStar` make sense (and I think it's more descriptive than `flipmodal`)?

src/child/components/Favourite.js
```diff
@@ -76,11 +91,33 @@ class Favourite extends Component {
...
+        const modalBottom = starTopPos + modalOffset + modalHeight;
+        return modalBottom >= window.innerHeight;
+    }
+
+    bubbleHeadTop() {
```
jleft: `bubbleHeadTop` => `bubbleHeadTopPosition`

src/child/components/Favourite.js
```diff
@@ -76,11 +91,33 @@ class Favourite extends Component {
...
+        return this.state.starTop + (this.flipModal(this.state.starTop) ? bubbleHeadFlippedOffset : bubbleHeadOffset);
+    }
+
+    modalTop() {
```
jleft: `modalTop` => `modalTopPosition`

src/child/components/Favourite.js
```diff
@@ -76,11 +91,33 @@ class Favourite extends Component {
         }
     }
 
+    onModalClick(e) {
+        e.stopPropagation();
+    }
+
+    onBackdropClick(e) {
```
jleft: `onBackdropClick` => `onModalBackdropClick`

src/child/containers/sidebars/favourites/Favourites.js
```diff
@@ -99,8 +109,20 @@ class Favourites extends Component {
...
     }
 
+    onConfirmClick(stockCode) {
+        return () => {
+            this.setState({ unfavouriting: null });
+            this.props.bindings.toggleFavourite(stockCode);
+        };
+    }
+
+    onBackdropClick(e) {
```
jleft: `onModalBackdropClick`

src/child/containers/sidebars/favourites/Favourites.js
```diff
@@ -99,8 +109,20 @@ class Favourites extends Component {
         this.props.dispatch(quandlResponse(stockCode, stockName));
     }
 
+    onConfirmClick(stockCode) {
```
jleft: `onModalConfirmClick`

src/child/components/Confirmation.js
```diff
@@ -0,0 +1,24 @@
...
+                <div className="buttons">
+                    <button className="btn btn-primary" onClick={props.bindings.onConfirmClick}>Confirm</button>
+                </div>
+            </div>
+        </div>
+    </div>
+);
+
+Confirmation.propTypes = {
+    bindings: PropTypes.object.isRequired,
```
jleft: Could use `PropTypes.shape`?

src/child/components/Favourite.js
```diff
@@ -76,11 +91,33 @@ class Favourite extends Component {
         }
     }
 
+    onModalClick(e) {
```
jleft: I'm not sure I'd put this in the favourite component and make it a prop for the confirmation component, if all it's doing is stopping the event propagating.

src/child/components/Favourite.js
```diff
@@ -76,11 +91,33 @@ class Favourite extends Component {
...
+        e.stopPropagation();
```
jleft: `stopPropagation` is best avoided (and in the other instances that it's used in this PR), would it be easy to use an alternative (`preventDefault`)?

jleft:
Thanks @jdhodges - this looks good to me. I think the coupling between the favourite and modal is fairly high but this also seems to be similar in the Angular version, so I've made a few comments above largely to do with smaller/style issues, once these are resolved then I think this should be good to go,

PR #842, opened 2016-06-09
Remove favourite from window X when dragging to window Y

rjmcneill:
Blocked by #837 (built on top of that PR)


rjmcneill:
I've noticed a problem with this, currently looking into fixing it

rjmcneill:

PR #844, opened 2016-06-10
Remove dev tools for production

rjmcneill:


bjedrzejewski:
I always found dev tools useful in production- you can show during the demo how this is made (to a potential client/prospect).

jleft:
Thanks @rjmcneill, this looks OK, but personally, I prefer a approach similar to #834 where there are `whatever.dev.js` and `whatever.prod.js` versions of the file, for a few reasons:

+ It avoids unnecessary import of DevTools in production (in this PR, it will still be imported and included in the bundle, even though it's unused).
+ It would be consistent with our approach for the reducers
+ I find it a bit easier to see the differences, rather than having conditionals scattered throughout
+ It's a bit easier to test without having to mess about with environment settings

You'll also not want to call [DevTools.instrument()](https://github.com/ScottLogic/StockFlux/blob/72e040fafec7b38d3dda1bd93824107118f4517b/src/child/store/configureStore.js#L15) in production - I think it would also make sense to have a dev and prod version of configureStore for this.

jleft:
> I always found dev tools useful in production- you can show during the demo how this is made (to a potential client/prospect).

That's a good point, I think it would be very cool to demonstrate how Redux works, I think the main downside would be the performance impact though.

jleft:
For now, to be consistent with the latest release/Angular version, dev tools will be disabled in production. It's a one line change, so easy to change this behaviour later :smiley: 

Closing as [this change](https://github.com/ScottLogic/StockFlux/blob/0173789760fa1484a45d28b8cf167fd831472615/src/shared/ConfigService.js#L12) was included in #847.

PR #853, opened 2016-06-30
fix: set bindings prop on compact view Search container

jleft:
bindings is a required property for the Search container


jleft:
Duplicated by #855

PR #855, opened 2016-06-30
fix: Added bindings prop to search in compact view

rjmcneill:
Removes console error.
Bug shows when toggling favourite in compact view.


jleft:
Thanks @rjmcneill :+1:

PR #857, opened 2016-07-04
Added splash screen to production

rjmcneill:
Need to make sure the new deploy script includes the splash screen image


jleft:
I think it's worth doing this for development: using the local server and [image](https://github.com/ScottLogic/StockFlux/blob/278a65138c4404fbf27935a73d9b10c3fe3cfaed/src/child/assets/png/splashscreen.png).

N.B. The [path was dynamic](https://github.com/ScottLogic/StockFlux/blob/d686e7fc45ee7dd4bdc47ae4cb75574f994501b9/gruntfile.js#L39) in the Angular version (but I don't think we want to do that here).

public/app.dev.json
```diff
@@ -20,5 +20,6 @@
         "company": "ScottLogic",
         "description": "Hosts BitFlux as an OpenFin application.",
         "name": "StockFlux"
-    }
+    },
+    "splashScreenImage": "splashscreen.png"
```
jleft: Do relative paths work, or should we use absolute URIs? (I don't have anything against relative paths, just noticed the Angular version uses absolute paths.)

public/app.prod.json
```diff
@@ -20,5 +20,6 @@
         "company": "ScottLogic",
         "description": "Hosts BitFlux as an OpenFin application.",
         "name": "StockFlux"
-    }
+    },
+    "splashScreenImage": "http://scottlogic.github.io/StockFlux/master/assets/png/splashscreen.png"
```
jleft: Thinking about this more... I think we should be using the same URL as above, for consistency (i.e. `http://localhost:5001`). I'm not sure using the deployed version makes much sense here.

Obviously, we'll have to make sure we have appropriate configs for the deployed version(s) too, but let's deal with that separately.

jleft:
Now that the splash screen image has been added to the public directory, I think the other copy can be removed from _src/child/assets/png/splashscreen.png_?

PR #858, opened 2016-07-04
Added icon to production

rjmcneill:
Need to make sure the new deploy script includes the icon image


jleft:
Similar comment to https://github.com/ScottLogic/StockFlux/pull/857#issuecomment-230296920; I think this should be done for development too, and the favicon should be included in this PR (assuming it's not already committed, which I don't think it is).

jleft:
I don't think we should be using the deployed version URLs here; `http://localhost:500X` is preferable.

jleft:
:+1: LGTM, thanks @rjmcneill - will merge when conflicts are resolved.

PR #859, opened 2016-07-04
Selection no longer possible through compact view

rjmcneill:


jleft:
Looks good, but is it worth pulling the event handler callback up a level to the _Sidebar_ container, in a similar way to `toggleFavourite`?

jleft:
LGTM - thanks :+1:

PR #860, opened 2016-07-04
Compact view no longer resizable

rjmcneill:


src/child/actions/window.js
```diff
@@ -61,6 +61,14 @@ export function resizeToCompact() {
     return dispatch => {
         dispatch(resizing());
         const compactWindowDimensions = configService.getCompactWindowDimensions();
+
+        fin.desktop.Window.getCurrent().updateOptions({
```
jleft: These options will still be set if the user goes from compact view to expanded view, preventing them from changing the window size in expanded view. The user should be still be able to resize and maximise in expanded view.

jleft:
Thanks @rjmcneill - this is looking good.

I've just had a look at the OpenFin API and `updateOptions()` is async (like `resizeTo`) - I think it would be quite neat if it was treated asynchronously with an async action creator.

I'm thinking something like this - let me know what you think!

+ Split update options and resize to into their own async action creators
+ Have both of the above return promises (essentially converting the existing callback API to a promise API)
+ Add an additional method which combines both of these, making use of the promises

Rough idea below for resizing to compact (would also need to update for resizing to expanded).

```js
// updateOptions async action creator
// N.B. This follows the common pattern where it may dispatch three actions
// - (starting, success, error) 
function updateOptionsToCompact() {
    return dispatch => {
        dispatch(updatingOptions());
        const [compactWindowWidth, compactWindowHeight] = configService.getCompactWindowDimensions();

        // Wrap callback API and return promise
        return new Promise((resolve, reject) => {
            fin.desktop.Window.getCurrent().updateOptions({
                resizable: false,
                maximizable: false,
                minWidth: compactWindowWidth,
                minHeight: compactWindowHeight
            },
            () => resolve(dispatch(updatingOptionsSuccess())),
            () => reject(dispatch(updatingOptionsError())));
        });
    };
}

// resizeTo compact async action creator
function resizeCompact() {
    return dispatch => {
        dispatch(resizing());
        const [compactWindowWidth, compactWindowHeight] = configService.getCompactWindowDimensions();

        // Wrap callback API and return promise
        return new Promise((resolve, reject) => {
            fin.desktop.Window.getCurrent().resizeTo(
                compactWindowWidth,
                compactWindowHeight,
                'top-right',
                () => resolve(dispatch(compact())),
                () => reject(dispatch(resizeError()))
            );
        });
    };
}

// composing the above into a single exposed function
// (I'm assuming the order of the calls doesn't matter?)
export function resizeToCompact() {
    return dispatch => Promise.all([
        dispatch(updateOptionsToCompact()),
        dispatch(resizeCompact())
    ]);
}
```

src/child/actions/window.js
```diff
@@ -57,30 +57,104 @@ export function resizeError() {
     };
 }
 
-export function resizeToCompact() {
+export function updatingOptions() {
```
jleft: I don't think `updatingOptions()`, `updatingOptionsSuccess()` or `updatingOptionsError()` should be exported. (Could potentially do this if they were being tested directly, but as they're not, lets not expose them.)

src/child/actions/window.js
```diff
@@ -57,30 +57,104 @@ export function resizeError() {
...
+export function updatingOptionsError() {
+    return {
+        type: ACTION_TYPES.OPTIONS_ERROR
+    };
+}
+
+function updateOptionsToCompact() {
+    return dispatch => {
+        dispatch(updatingOptions());
+        const [compactWindowWidth, compactWindowHeight] = configService.getCompactWindowDimensions();
```
jleft: Could name these `minWidth` and `minHeight` to use property value shorthand in the object passed to `updateOptions()` below - both in `updateOptionsToCompact()` and `updateOptionsToDefault()`.

(Sorry - my fault for messing with the code example in the comment I made!)

src/child/constants/actionTypes.js
```diff
@@ -22,5 +22,8 @@ export const WINDOW = {
     STATE_EXPANDED: 'STATE_EXPANDED',
     STATE_FULL_VIEW: 'STATE_FULL_VIEW',
     RESIZING: 'RESIZING',
-    RESIZE_ERROR: 'RESIZE_ERROR'
+    RESIZE_ERROR: 'RESIZE_ERROR',
+    UPDATING_OPTIONS: 'UPDATING_OPTIONS',
+    OPTIONS_SUCCESS: 'OPTIONS_SUCCESS',
+    OPTIONS_ERROR: 'OPTIONS_ERROR'
```
jleft: I'd add the verb to these actions - `UPDATING_OPTIONS_SUCCESS`, `UPDATING_OPTIONS_ERROR`

PR #861, opened 2016-07-05
feat: compact view expands to previous size

jdhodges:


src/child/actions/window.js
```diff
@@ -71,15 +78,22 @@ export function resizeToCompact() {
...
         fin.desktop.Window.getCurrent().resizeTo(
-            defaultWindowDimensions[0],
-            defaultWindowDimensions[1],
+            dimensions[0],
+            dimensions[1],
             'top-right',
-            () => dispatch(expand()),
+            () => {
+                dispatch(expand());
+                if (isMaximised) {
```
jleft: If the window is going to be maximised, does `resizeTo()` still need to be called, or is `maximize()` enough?

I think this raises an interesting point too - should `isMaximised` be set to `false` when `isCompact` is `true` - it seems contradictory that `isCompact` and `isMaximised` can both be `true` simultaneously?

src/child/index.js
```diff
@@ -20,6 +21,15 @@ require('script!../../node_modules/BitFlux/dist/bitflux.js');
 
 const store = configureStore();
 
+fin.desktop.main(() => {
+    fin.desktop.Window.getCurrent().addEventListener('bounds-changed', (e) => {
```
jleft: My initial feeling is that this is a bit excessive - just a feeling though.

Rather than storing the dimensions whenever the window changes size, can this just be done when the user selects compact view (and store them as `previousExpandedDimensions`, or something similar)? Would this be sufficient (or is there another use case which I'm not considering)?

src/child/actions/window.js
jdhodges: Without the `resizeTo()` first, a subsequent restore from maximize will cause the window to still be max sized but just in a different position. Using `resizeTo()` before `maximize()` ensured the restore would work correctly. Is there a better way to do this?

jleft: Ah, good point!

Just to check my understanding: if `resizeTo()` isn't called and the user was to do the following:

1. Open app
2. Maximise window
3. Compact view
4. Expand (back to maximised)
5. Restore

Then step 5 would be wrong as it would restore to compact view dimensions, and not expanded dimensions?

If that's the case, then it looks like performing the `resizeTo` would be necessary :+1:

src/child/index.js
jleft: I think it was mentioned offline that there are a few reasons for using `bounds-changed` over the alternative mentioned above. Please could the reason(s) be added here, so we have a record of this?

src/child/actions/window.js
```diff
@@ -75,6 +82,54 @@ function updatingOptionsError() {
     };
 }
 
+export function resizeToMin() {
```
rjmcneill: Can we combine this with the `minimise` function?
If the existing minimise function is used elsewhere, its probably worth just using the actions that is dispatched from the success of this promise, since we're not minimised unless this promise succeeds.
I'd rename this to `minimising` as well, then `minimiseSuccess` and `minimiseError`, or resizing, whatever you think.

Same applies to the maximise I think?

jdhodges: `minimise`, `maximise`, and `restore` are dispatched by [these event listeners](https://github.com/jdhodges/StockFlux/blob/fea2287d2244886634e6d03fffd9f9a43a16fcd8/src/child/containers/toolbar/Toolbar.js#L33-L35). This means they are dispatched when minimising, maximising, or restoring from either the `Toolbar` or the OS. I couldn't think of another way to cover both those cases so hopefully this is ok.

src/child/actions/window.js
```diff
@@ -4,17 +4,18 @@ import createActionCreator from '../utils/createActionCreator';
 import currentWindowService from '../services/currentWindowService';
 
 export const minimise = createActionCreator(() => ({
-    type: ACTION_TYPES.MINIMIZE
+    type: ACTION_TYPES.MINIMISE
```
jleft: Renaming for consistency is a great idea 👍  however, I think I'd rename to be consistent with the APIs which we're using (i.e. using the American spelling which OpenFin and the `Window` api use) to reduce any confusion about which spelling to use.

src/child/actions/window.js
```diff
@@ -53,6 +58,59 @@ const updatingOptionsError = createActionCreator(() => ({
     type: ACTION_TYPES.UPDATING_OPTIONS_ERROR
 }));
 
+export const resizeWindow = createActionCreator((dimensions) => ({
+    type: ACTION_TYPES.RESIZE_WINDOW,
+    previousExpandedDimensions: dimensions
+}));
+
+export function resizeToMin() {
```
jleft: Why `resizeTo[X]` (for `Min`, `Max` and `Restore`)? They're no calling `resizeTo` (as far as I'm aware?), so would `minimize`, `maximize` and `restore` be more appropriate?

src/child/actions/window.js
```diff
@@ -99,40 +157,40 @@ function resizeCompact() {
...
-                () => resolve(dispatch(expand())),
+                () => resolve(dispatch(resizePreviousSuccess(previousMaximisedState))),
                 () => reject(dispatch(resizeError()))
             );
         });
     };
 }
 
-export function resizeToCompact() {
+export function resizeToCompact(isMaximised) {
```
jleft: I'm not sure I'd add the param here - if the value is just coming from the state, then why not get the state in the thunk, using `getState()`?

> The inner function receives the store methods `dispatch` and `getState` as parameters.

(https://github.com/gaearon/redux-thunk#motivation)

src/child/actions/window.js
```diff
@@ -99,40 +157,40 @@ function resizeCompact() {
...
     return dispatch => Promise.all([
         dispatch(updateOptionsToCompact()),
-        dispatch(resizeCompact())
+        dispatch(resizeCompact(isMaximised))
     ]);
 }
 
-export function resizeToDefault() {
+export function resizeToPrevious(previousExpandedDimensions, previousMaximisedState) {
```
jleft: I'm not sure I'd add the params here - if the values are just coming from the state, then why not get the state in the thunk, using `getState()`?

> The inner function receives the store methods `dispatch` and `getState` as parameters.

(https://github.com/gaearon/redux-thunk#motivation)

src/child/index.js
```diff
@@ -30,6 +30,12 @@ currentWindowService.ready(() => {
     const windowStateService = new WindowStateService(currentWindow, store);
     windowStateService.start();
 
+    currentWindow.addEventListener('bounds-changed', (e) => {
```
jleft: https://github.com/ScottLogic/StockFlux/pull/861#discussion_r70758268

src/child/index.js
jdhodges: The `previousExpandedDimensions` would also have to be saved when maximising so we can support the use case:
```
resize normal view -> maximise -> compact -> full view (to maximised) -> restore (back to resized normal view)
```
This is fine if the maximise is performed via the `Toolbar` since we can get the current window size before we perform the actual window resize. The problem is when the maximise is performed via the OS/windows taskbar - in this case the window resize operation has already happened before we get the chance to save the original window dimensions.

jleft: Thanks for the explanation :smiley:

src/child/actions/window.js
```diff
@@ -3,18 +3,19 @@ import configService from '../../shared/ConfigService';
...
+export const minimize = createActionCreator(() => ({
     type: ACTION_TYPES.MINIMIZE
 }));
 
-export const compact = createActionCreator(() => ({
+export const compact = createActionCreator((isMaximised) => ({
     type: ACTION_TYPES.TOGGLE_COMPACT,
-    state: true
+    isCompact: true,
+    previousMaximisedState: isMaximised
```
jleft: Still need to change to American spelling of maximised/maximized. Hopefully a quick find and replace will sort it :smile:

src/child/index.js
jleft: I think this should be added to the `WindowStateService`, along with the other window event listeners,

src/child/index.js
```diff
@@ -30,6 +30,12 @@ currentWindowService.ready(() => {
...
+        if (e.changeType && !store.getState().childWindows[e.name].windowState.isCompact) {
```
jleft: I think `e.changeType !== 0` would show the intent more clearly, maybe with a comment explaining that this is a change in position, not size. Or even `const positionChangeType = 0;` and `e.changeType !== positionChangeType`

test/specs/child/actions/windowSpec.js
```diff
@@ -151,7 +168,7 @@ describe('child/actions/window', () => {
                     { windowName: 'window0002', type: ACTION_TYPES.UPDATING_OPTIONS },
                     { windowName: 'window0002', type: ACTION_TYPES.UPDATING_OPTIONS_SUCCESS },
                     { windowName: 'window0002', type: ACTION_TYPES.RESIZING },
-                    { windowName: 'window0002', type: ACTION_TYPES.RESIZE_ERROR }
+                    { windowName: 'window0002', type: ACTION_TYPES.RESIZE_ERROR },
```
jleft: Remove trailing comma

src/child/actions/window.js
jleft: Thanks for updating.

Just a follow up question - why are these dispatching `resize` actions internally? Same reason as before, I don't think they're actually resizing the window?

jleft:
I think one of my questions was lost in the diff (https://github.com/ScottLogic/StockFlux/pull/861#discussion_r75072785), but I think this is any easy change to make later - so happy to merge now and discuss later 👍

PR #863, opened 2016-07-07
[Single parent store] Dragging out favourite opens new window

rjmcneill:


src/child/components/Favourite.js
```diff
@@ -59,10 +59,11 @@ class Favourite extends Component {
         e.currentTarget.classList.remove('dragOver');
     }
 
-    onDragEnd(e) {
+    onDragEnd(e, stockCode) {
```
jdhodges: We can make this look a bit nicer if we use currying:
```
onDragEnd(stockCode) {
  return e => {
    ...
  };
}
```
Then on line 97 we can do:
```
onDragEnd={this.onDragEnd(stockCode)}
```

src/child/containers/sidebars/favourites/Favourites.js
```diff
@@ -69,6 +73,20 @@ class Favourites extends Component {
         }
     }
 
+    onDropOutside(e, stockCode) {
+        this.props.dispatch(dragOut(stockCode, this.props.favourites.names[stockCode]));
+
+        // Unfavourite the stock from this window
+        this.props.bindings.toggleFavourite(stockCode);
+
+        const childWindow = new fin.desktop.Window(
```
jleft: This feels quite odd - creating a new child window from the favourites container in a child window.

Feels like it should be the responsibility of the parent. Perhaps an action could be dispatched which will open the new window?

src/child/index.js
```diff
@@ -20,14 +21,21 @@ require('script!../../node_modules/BitFlux/dist/bitflux.js');
 /* eslint-enable import/no-unresolved */
 
 fin.desktop.main(() => {
-    fin.desktop.Window.getCurrent().contentWindow.addEventListener('beforeunload', () => {
-        fin.desktop.Window.getCurrent().contentWindow.opener.store.dispatch(close());
+    const store = fin.desktop.Window.getCurrent().contentWindow.opener.store;
+    fin.desktop.Window.getCurrent().contentWindow.store = fin.desktop.Window.getCurrent().contentWindow.opener.store;
```
jleft:
```js
fin.desktop.Window.getCurrent().contentWindow.store = store;
```

src/child/index.js
```diff
@@ -20,14 +21,21 @@ require('script!../../node_modules/BitFlux/dist/bitflux.js');
...
+
+    $(window).unload(() => {        // eslint-disable-line no-undef
```
jleft: What's the reason for using jQuery and changing from `beforeunload` to `unload`?

src/child/index.js
```diff
@@ -20,14 +21,21 @@ require('script!../../node_modules/BitFlux/dist/bitflux.js');
...
         unmountComponentAtNode(document.getElementById('app'));
+        store.dispatch(close());
     });
 
-    const store = fin.desktop.Window.getCurrent().contentWindow.opener.store;
     store.dispatch(open());
 
+    console.log(store.getState().dragOut);
```
jleft: Remove

src/parent/reducers/reducers.js
```diff
@@ -8,6 +8,43 @@ function parentReducer(state = {}, action) {
     case ACTION_TYPES.CLOSE: {
         const newState = Object.assign({}, state);
         delete newState[action.windowName];
+
+        if (!Object.keys(newState).length) {
+            fin.desktop.Window.getCurrent().contentWindow.close();
```
jleft: Side effect in the reducer - seems like this should be avoided.

src/parent/reducers/reducers.js
```diff
@@ -8,6 +8,43 @@ function parentReducer(state = {}, action) {
...
+        return newState;
+    }
+
+    case ACTION_TYPES.DRAG_OUT: {
+        const newState = Object.assign({}, state);
+        newState.dragOut = {
+            stockCode: action.code,
+            stockName: action.name
+        };
+        console.log(newState);
```
jleft: Remove

src/parent/reducers/reducers.js
```diff
@@ -8,6 +8,43 @@ function parentReducer(state = {}, action) {
...
+        return newState;
+    }
+
+    case ACTION_TYPES.DRAG_ACCEPT: {
```
jleft: I think making use of reducer composition would clean this up - there could be a `dragAndDropReducer` to handle these actions. I think changing the state shape would help with this too. I think something like the example would work well - so reducer composition can be used and hopefully would solve some of the more awkward operations in these reducers - such as deleting the dragOut state and counting the number of child windows based on the number of keys in the state.

Just ideas though - might be easier to discuss :smiley: 

For example:

**Current**
```js
{
    window0001: childReducer,
    window0002: childReducer,
    dragOut: {}  //parent reducer handles this too
}
```

**Proposed**
```js
{
    childWindows: {
        window0001: childReducer,
        window0002: childReducer
    },
    dragOut: dragAndDropReducer
}
```

rjmcneill:
Currently blocked by some WIP PRs and #878 
Will rebase, resolve, and update after these are merged

jleft:
Replaced by #893

PR #866, opened 2016-07-11
feat: Added hover styles to favourites

rjmcneill:


src/child/containers/sidebars/favourites/Favourites.js
```diff
@@ -66,6 +68,24 @@ class Favourites extends Component {
         }
     }
 
+    onMouseOver(stockCode) {
```
jdhodges: Might be easier just to use the event to get a reference to the favourite element like this:

```
    onMouseOver(e) {
        e.currentTarget.getElementsByClassName('favourite')[0].classList.add('hovered');
    }

    onMouseOut(e) {
        e.currentTarget.getElementsByClassName('favourite')[0].classList.remove('hovered');
    }
```

jleft:
Could this be achieved with just CSS [`:hover`](https://developer.mozilla.org/en-US/docs/Web/CSS/:hover)? Would seem simpler to me, unless there's a reason why this wouldn't be suitable?

rjmcneill:
@jleft the pointer events are turned off to make sure drag and drop works properly, so that's why I had to use this method.
If we go with this and you think @jdhodges suggestion is nicer, I'm happy to implement and push up again

src/child/containers/sidebars/favourites/Favourites.js
jleft: If we are to go with this approach, then I think this suggestion is nicer 👍 , however...

I'm not particularly keen on this approach for a few reasons:
+ It's imperative and manipulates the DOM directly (please correct me if I'm wrong and there's some React magic happening!)
  + Feels like a declarative approach would be nicer and more inline with the React way of thinking - perhaps using a prop could be an option (e.g. `isHovered` could be a prop used by the `Favourite` component to set the correct style, rather than adding a class directly)
+ It adds too much coupling (IMO) between the `Favourite` component and the `Favourites` container
  + The `Favourites` container has knowledge of the child `Favourite` component's DOM structure and directly manipulates it - this seems particularly brittle
  + The `Favourite` component relies on a parent component to ensure it's hover style works correctly - seems a bit excessive for simple UI state which could be handled with CSS if there weren't the DnD issues 

Couple of alternative ideas:
+ Make use of `props` (e.g. something like the `isHovered` suggestion above)
+ Make use of `state` locally within the `Favourite` component - I think this seems a bit more appropriate in this case, as it's only simple UI state and it removes the coupling; although, obvious downside is that being stateless would be simpler than stateful

jleft:
> @jleft the pointer events are turned off to make sure drag and drop works properly, so that's why I had to use this method.
If we go with this and you think @jdhodges suggestion is nicer, I'm happy to implement and push up again

Thanks @rjmcneill - makes sense, I've added a comment above

src/child/components/Favourite.js
```diff
@@ -95,9 +106,11 @@ class Favourite extends Component {
               onDrop={e => bindings.dnd.onDrop(e, stockCode)}
               onDragEnd={this.onDragEnd}
               onDoubleClick={() => bindings.onDoubleClick(stockCode, name)}
+              onMouseOver={this.onMouseOver}
+              onMouseOut={this.onMouseOut}
             >
                 <div className="drop-target">
-                    <div className={`darkens favourite tearable ${cls}`} draggable="false">
+                    <div className={`darkens favourite tearable ${cls} ${isHovered ? 'hovered' : ''}`} draggable="false">
```
jleft: I think it would be a bit cleaner and more consistent to make use of the `cls` const above and revert the change on this line.

```js
const cls = classNames({
    selected,
    dark: selected,
    hovered: isHovered
});
```

PR #867, opened 2016-07-13
fix: fix drag and drop

jdhodges:
Does not include the removal of a favourite from window X when dragged from window Y as this is covered by #864.

- Dragged tile will displace another tile when it is half intersecting it (vertically)
- Dragged tile can be dropped after the final tile (rather than in the space below the final tile)
- Cleaned up the transitions
- When dragging into another window, the tile can be dropped in the sidebar regardless of whether the favourites are displayed or not

### DIFF
--- a/src/child/containers/sidebars/Sidebar.js
+++ b/src/child/containers/sidebars/Sidebar.js
@@ -6,7 +6,7 @@ import { sidebarSelector as mapStateToProps } from '../../selectors/selectors';
 
 import { selectFavourites, selectSearch } from '../../actions/sidebar';
 import { selectStock, unselectStock } from '../../actions/selection';
-import { toggleFavourite } from '../../actions/favourites';
+import { toggleFavourite, insertFavouriteAt } from '../../actions/favourites';
 
 import windowStateShape from '../../propTypeShapes/windowState';
 import selectionShape from '../../propTypeShapes/selection';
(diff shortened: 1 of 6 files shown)


src/child/containers/sidebars/favourites/Favourites.js
```diff
@@ -96,50 +61,112 @@ class Favourites extends Component {
...
+        const { favourites, bindings } = props;
 
-        dropTarget.addEventListener('drop', e => {
-            const codes = this.props.favourites.codes;
-            const code = e.dataTransfer.getData('text/plain');
-            this.props.dispatch(insertFavouriteAt(codes.length, code));
-            dropTarget.classList.remove('dragOver');
-        }, false);
+        if (dragStartClientY) {
+            const favouriteHeight = 110;
```
rjmcneill: Could we get this from the HTML element itself?

src/child/containers/sidebars/Sidebar.js
```diff
@@ -15,12 +15,49 @@ import favouritesShape from '../../propTypeShapes/favourites';
...
+        for (let i = 0; i < types.length; i++) {
+            if (types[i] !== 'text/plain') {
+                return (types[i]).toUpperCase();
+            }
+        }
+        return undefined;
+    }
+
+    addActive() {
+        this.favourites.classList.add('active');
```
jleft: I'm still not keen on manipulating the DOM directly like this (similar to [the comment in the hover styles PR](https://github.com/ScottLogic/StockFlux/pull/866#discussion_r71994300)) - I think using `props` or `state` would be more appropriate (for all instance of `classList.add` and `classList.remove` in this PR). Unless there's a reason to this approach?

src/child/containers/sidebars/favourites/Favourites.js
```diff
@@ -11,46 +11,30 @@ import selectionShape from '../../../propTypeShapes/selection';
 import favouritesShape from '../../../propTypeShapes/favourites';
 import windowStateShape from '../../../propTypeShapes/windowState';
 
-/*
- *  dataTransfer.getData is only available in dragstart, drop and dragEnd
```
jleft: I think I'd keep this comment with the function, as I think it makes it easier to understand what's happening.

src/child/containers/sidebars/favourites/Favourites.js
```diff
@@ -121,36 +94,93 @@ class Favourites extends Component {
...
+        });
     }
 
     onDragOver(e) {
-        if (!e.currentTarget.classList.contains('dragOver')) {
-            e.currentTarget.classList.add('dragOver');
+        const { removeDragOver, props, state } = this;
+        const { favourites, bindings } = props;
+        const { dragOverIndex, favouriteHeight, dragStartClientY } = state;
+        const favouriteElements = this.sidetab.childNodes;
```
jleft: Again, similar comment to above - I'm really not keen on a parent element selecting elements in a child component and adding/removing classes imperatively, outside of React. If a class needs to be toggled on a child component, then could a prop be passed down to that child component, where it can set the class?

src/child/containers/sidebars/favourites/favourites.less
```diff
@@ -1,4 +1,4 @@
-@favourite-height: 110px;
+@favourite-height: 109px;
```
jleft: Why?

PR #868, opened 2016-07-18
Moved PropTypes.object to PropTypes.shape where possible

rjmcneill:


jleft:
- Commit should start with `refactor`, not `chore`

> refactor: A code change that neither fixes a bug nor adds a feature
> chore: Changes to the build process or auxiliary tools and libraries such as documentation generation

- There seems to be a fair bit of repetition with some of the shapes (e.g. `favourites`, `windowState`, `selection`) - could these be extracted to a separate modules, exported and reused?

src/child/containers/sidebars/Sidebar.js
```diff
@@ -78,9 +81,9 @@ class Sidebar extends Component {
 Sidebar.propTypes = {
     dispatch: PropTypes.func.isRequired,
     sidebar: PropTypes.object.isRequired,
-    windowState: PropTypes.object.isRequired,
-    selection: PropTypes.object.isRequired,
-    favourites: PropTypes.object.isRequired
+    windowState: windowStateShape,
+    selection: selectionShape,
+    favourites: favouritesShape,
```
jleft: Remove trailing comma

src/child/proptypes/favourites.js
```diff
@@ -0,0 +1,8 @@
+import { PropTypes } from 'react';
+
+const favouritesShape = PropTypes.shape({
+    codes: PropTypes.array.isRequired,
+    names: PropTypes.object.isRequired
+}).isRequired;
```
jleft: Minor detail and just a thought - it might be a bit more flexible and I'm not sure how clear it is when using these shapes that they are required - but the `.isRequired` could be omitted here.

E.g.

**Before**

```js
const favouritesShape = PropTypes.shape({
    codes: PropTypes.array.isRequired,
    names: PropTypes.object.isRequired
}).isRequired;

Whatever.propTypes = {
    favourites: favouritesShape
};
```
**After**

```js
const favouritesShape = PropTypes.shape({
    codes: PropTypes.array.isRequired,
    names: PropTypes.object.isRequired
});

Whatever.propTypes = {
    favourites: favouritesShape.isRequired
};
```

src/child/proptypes/favourites.js
```diff
@@ -0,0 +1,8 @@
+import { PropTypes } from 'react';
```
jleft: Could rename the `proptypes` directory to `propTypesShapes` - I think camelCase will be more consistent and I don't think I'd want to encourage any other propTypes going in here, only shapes which are reused.

PR #871, opened 2016-07-22
feat: add closed windows list

jdhodges:


src/child/containers/sidebars/Sidebar.js
```diff
@@ -130,13 +132,18 @@ class Sidebar extends Component {
         }
     }
 
+    openWindow(windowName) {
+        this.props.dispatch(reopenWindow(windowName));
+    }
+
     render() {
-        const { sidebar } = this.props;
+        const { sidebar, unopenedWindows } = this.props;
```
rjmcneill: `unopenedWindows` seems a little odd to me, would `closedWindows` be a better name?
Unless there's a reason otherwise?

src/parent/reducers/reducers.js
```diff
@@ -1,12 +1,61 @@
...
+        switch (action.type) {
+        case SIDEBAR.REOPEN_WINDOW: {
+            const newClosedWindow = Object.assign({}, state.closedWindows[action.reopenWindowName], {});
+
+            delete newClosedWindow.date;
+
+            const newClosedWindows = Object.assign({}, state.closedWindows);
+            delete newClosedWindows[action.reopenWindowName];
+
+            return {
```
jleft: Shouldn't this also handle the `dragOut` part of the state? Otherwise won't the state be returned without the `dragOut` slice of state?

src/child/containers/sidebars/Sidebar.js
```diff
@@ -166,7 +172,8 @@ Sidebar.propTypes = {
     sidebar: PropTypes.object.isRequired,
     windowState: windowStateShape.isRequired,
     selection: selectionShape.isRequired,
-    favourites: favouritesShape.isRequired
+    favourites: favouritesShape.isRequired,
+    numberOfClosedWindows: PropTypes.number.isRequired
```
jleft: I'd be tempted to name as `closedWindowsCount` as it's more concise but just as descriptive

src/child/selectors/selectors.js
```diff
@@ -27,10 +29,13 @@ export const searchSelector = createCurrentWindowStateSelector(
...
 );
 
-export const sidebarSelector = createCurrentWindowStateSelector(
-    (state) => {
-        const { sidebar, selection, favourites, windowState } = state;
-        return { sidebar, selection, favourites, windowState };
+export const sidebarSelector = createSelector(
+    [getClosedWindows, getCurrentWindowState],
+    (closedWindows, currentWindowState) => {
+        const closedWindowNames = closedWindows ? Object.keys(closedWindows) : [];
```
jleft: When would `closedWindows` be falsy?

src/child/selectors/selectors.js
```diff
@@ -40,3 +45,16 @@ export const toolbarSelector = createCurrentWindowStateSelector(
         return { windowState };
     }
 );
+
+export const closedWindowsSelector = createSelector(
+    getClosedWindows,
+    (state) => {
+        const closedWindows = state || {};
```
jleft: When would `closedWindows` be falsy?

src/parent/parent.js
```diff
@@ -7,3 +7,5 @@ fin.desktop.main(() => {
     const parentService = new ParentService(store);
     parentService.start();
 });
+
+window.parent = window;
```
jleft: Is this needed/used?

src/parent/reducers/reducers.js
```diff
@@ -1,12 +1,61 @@
...
 
 import childWindows from './childWindows';
+import closedWindows from './closedWindows';
 import dragOut from './dragOut';
 
-const rootReducer = combineReducers({
-    childWindows,
-    dragOut
-});
+const rootReducer = reduceReducers(
```
jleft: Another idea could be to structure the state shape like this:

```js
{
    childWindows: {
        window0001: { ... },
        window0002: { ... }
    },
    openWindows: ['window0001'], // (window IDs)
    closedWindow: ['window0002'], // (window IDs)
    ...
}
```

Might not be worth doing this now though, as I expect it would be a fairly large change.

src/child/actions/sidebar.js
```diff
@@ -8,3 +9,26 @@ export const selectSearch = createActionCreator(() => ({
 export const selectFavourites = createActionCreator(() => ({
     type: ACTION_TYPES.FAV_CLICKED
 }));
+
+const windowReopened = createActionCreator((reopenWindowName) => ({
+    type: ACTION_TYPES.REOPEN_WINDOW,
+    reopenWindowName
+}));
+
+const openWindowError = createActionCreator(() => ({
```
jleft: Worth passing the `windowName` too? Although looks like this action isn't handled at the moment?

src/child/index.js
```diff
@@ -25,14 +25,18 @@ require('script!../../node_modules/d3fc/dist/d3fc.bundle.min.js');
...
     const currentWindow = currentWindowService.getCurrentWindow();
-    const store = currentWindow.contentWindow.opener.store;
     const rootElement = document.getElementById('app');
 
     const windowStateService = new WindowStateService(currentWindow, store, rootElement);
     windowStateService.start();
 
+    window.store = store;
+    window.parent = parent;
+
```
jdhodges: `store` is used here to cover the case when the opener window is closed before the spawned window reaches the ready state. `parent` doesn't seem to be used, must've been left over after a rebase!

src/child/selectors/selectors.js
jdhodges: `closedWindows` is initialised as `{}` and `Object.keys({})` returns `[]` so this isn't needed.

src/child/actions/sidebar.js
```diff
@@ -16,3 +17,27 @@ export const selectFavourites = createActionCreator(() => ({
...
+}));
+
+const openWindowError = createActionCreator((reopenWindowName) => ({
+    type: ACTION_TYPES.OPEN_WINDOW_ERROR,
+    reopenWindowName
+}));
+
+export function reopenWindow(windowName) {
+    return dispatch => {
+        const childWindow = new fin.desktop.Window(
```
jleft: Might it be better to follow the pattern used when dragging out a window, to send a message to the parent service to open the window?

As it is, does this mean a reopened window would be child of a child (or is the window creation run in the parent)? It would be more consistent and I think it would avoid having to store a global reference to the store on the child windows.

src/child/actions/sidebar.js
```diff
@@ -16,3 +17,27 @@ export const selectFavourites = createActionCreator(() => ({
...
+            configService.getWindowConfig(windowName),
+            () => childWindow.show(
+                false,
+                () => dispatch(windowReopened(windowName)),
+                () => dispatch(openWindowError())
```
jleft: If a message isn't used to reopen the window, as suggested above, then I think this should be passed the `windowName`?

src/child/index.js
```diff
@@ -26,14 +26,17 @@ require('script!../../node_modules/d3fc/dist/d3fc.bundle.min.js');
...
 currentWindowService.ready(() => {
     const currentWindow = currentWindowService.getCurrentWindow();
-    const store = currentWindow.contentWindow.opener.store;
     const rootElement = document.getElementById('app');
 
     const windowStateService = new WindowStateService(currentWindow, store, rootElement);
     windowStateService.start();
 
+    window.store = store;
```
jleft: If we go with the following, then I don't think this is needed?

> Might it be better to follow the pattern used when dragging out a window, to send a message to the parent service to open the window?

src/parent/reducers/reducers.js
```diff
@@ -1,12 +1,72 @@
...
+
+            const newClosedWindows = Object.assign({}, state.closedWindows);
+            delete newClosedWindows[action.reopenWindowName];
+
+            return {
+                childWindows: Object.assign({}, state.childWindows, {
+                    [action.reopenWindowName]: newClosedWindow
+                }),
+                closedWindows: newClosedWindows,
+                dragOut: state.dragOut ? Object.assign({}, state.dragOut) : null
```
jleft: As the reducers are run left to right, isn't `dragOut: state.dragOut` enough?

src/parent/reducers/reducers.js
```diff
@@ -1,12 +1,72 @@
...
+                .sort((a, b) => newClosedWindows[a].date - newClosedWindows[b].date)
+                .slice(-5)
+                .reduce((prev, key) => Object.assign(prev, {
+                    [key]: newClosedWindows[key]
+                }), {});
+
+            return {
+                childWindows: newChildWindows,
+                closedWindows: limitedClosedWindows,
+                dragOut: state.dragOut ? Object.assign({}, state.dragOut) : null
```
jleft: As the reducers are run left to right, isn't `dragOut: state.dragOut` enough?

src/child/index.js
jleft: Okay, thanks. Related: https://github.com/ScottLogic/StockFlux/pull/871#discussion_r81630443

src/child/selectors/selectors.js
jleft: 👍

src/parent/reducers/reducers.js
```diff
@@ -1,12 +1,72 @@
...
+            const numberOfFavourites = state.childWindows[action.windowName].favourites.codes.length;
+
+            const newClosedWindows = Object.assign({}, state.closedWindows, numberOfFavourites ? {
+                [action.windowName]: newClosedWindow
+            } : {});
+
+            const limitedClosedWindows = Object.keys(newClosedWindows)
+                .sort((a, b) => newClosedWindows[a].date - newClosedWindows[b].date)
+                .slice(-5)
+                .reduce((prev, key) => Object.assign(prev, {
```
jleft: `key` => `windowName`

src/parent/reducers/reducers.js
```diff
@@ -1,12 +1,72 @@
...
+        case ACTION_TYPES.CLOSE: {
+            const newClosedWindow = Object.assign({}, state.childWindows[action.windowName], {
+                date: action.date
+            });
+
+            const newChildWindows = Object.assign({}, state.childWindows);
+
+            delete newChildWindows[action.windowName];
+
+            const numberOfFavourites = state.childWindows[action.windowName].favourites.codes.length;
```
jleft: I'd probably do the following:

`const closingWindowHasFavourites = state.childWindows[action.windowName].favourites.codes.length > 0`

As I think it's slightly more clear what's happening. Could even do this check earlier and return early - I think it's fine as-is though.

src/parent/actions/parent.js
```diff
@@ -24,22 +32,30 @@ export function dragAccept() {
...
-        } else {
-            const application = fin.desktop.Application.getCurrent();
-            application.getChildWindows(children => {
-                children.find(childWindow => childWindow.name === currentWindowName).close();
-            });
-        }
+        dispatch(toggleFavourite(code));
+        const application = fin.desktop.Application.getCurrent();
+        application.getChildWindows(children => {
+            children.find(childWindow => childWindow.name === currentWindowName).close();
```
jleft: I think the window should only be closed if it has <= 1 favourite.

PR #874, opened 2016-07-25
fix: compact/maximised works with state persisting

rjmcneill:


src/child/index.js
```diff
@@ -20,13 +21,22 @@ require('script!../../node_modules/BitFlux/dist/bitflux.js');
...
     fin.desktop.Window.getCurrent().contentWindow.addEventListener('beforeunload', () => {
         fin.desktop.Window.getCurrent().contentWindow.opener.store.dispatch(close());
         unmountComponentAtNode(document.getElementById('app'));
     });
 
-    const store = fin.desktop.Window.getCurrent().contentWindow.opener.store;
-    store.dispatch(open());
+    // Ensure that the window is started in the same
+    // state as the previous session, if available
+    if (windowState.isCompact) {
```
jleft: Did you consider doing this before the window is created and shown?

When creating the window, the previous window state (rather than the default) could be set in the [options object](http://cdn.openfin.co/jsdocs/stable/tutorial-window.constructor.html) passed to the window constructor; if anything can't be set in this option object and needs to be done (e.g. a method call), then I think this should be done before the window is shown.

jleft:
Replaced by #895

PR #879, opened 2016-07-31
refactor: deduplicate action types, split action creators and move close to parent

jleft:


jdhodges:
I seem to be getting an error when closing via windows:

image

This causes the child window to close but the parent does not.

src/parent/services/ParentService.js
```diff
@@ -0,0 +1,47 @@
...
+
+class ParentService {
+
+    constructor(store) {
+        this.store = store;
+        this.onChildClosed = this.onChildClosed.bind(this);
+    }
+
+    onChildClosed({ name }) {
+        this.store.dispatch(close(name));
```
jdhodges: Looks like we can fix the problem I mentioned in my last comment if we do this instead
```
if (Object.keys(this.store.getState()).length === 1) {
    fin.desktop.Window.getCurrent().contentWindow.close();
} else {
    this.store.dispatch(close(name));
}
```
With the persistence of state stuff being in, we don't want to dispatch the close action when the final child window is closed.

jleft: > With the persistence of state stuff being in, we don't want to dispatch the close action when the final child window is closed.

Good spot - didn't think of that.

> Looks like we can fix the problem I mentioned in my last comment if we do this instead

I'm not convinced this will solve the problem - it seems that closing a window using Alt+F4/Taskbar doesn't trigger the `beforeunload` event, so the root component is not unmounted which leads to the errors seen above.

jdhodges: Would using `unload` work?

PR #880, opened 2016-08-01
refactor: move window state handlers from toolbar container to service

jleft:


src/child/index.js
```diff
@@ -19,19 +21,26 @@ require('script!../../node_modules/d3fc/dist/d3fc.bundle.min.js');
...
+currentWindowService.ready(() => {
+    const currentWindow = currentWindowService.getCurrentWindow();
+    const store = currentWindow.contentWindow.opener.store;
+    const rootElement = document.getElementById('app');
+
+    const windowStateService = new WindowStateService(currentWindow, store);
+    windowStateService.start();
+
+    currentWindow.contentWindow.addEventListener('beforeunload', () => {
+        store.dispatch(close());
```
rjmcneill: I think we need to unmount before dispatching close and stopping the service.
I've fixed that in another PR, but just in case this gets merged first

PR #881, opened 2016-08-05
refactor: close parent in thunk

jdhodges:


rjmcneill:
LGTM :+1:

jleft:
Note: do not merge yet; need to consider how this would work with moving the close to parent (#879). Let's discuss.

PR #882, opened 2016-08-09
refactor: add child windows property to state

jdhodges:
Also moved state/action freezing in dev from child to parent.


src/parent/reducers/reducers.js
```diff
@@ -7,7 +7,7 @@ function parentReducer(state = {}, action) {
 
     case ACTION_TYPES.CLOSE: {
         const newState = Object.assign({}, state);
-        delete newState[action.windowName];
+        delete newState.childWindows[action.windowName];
```
jleft: I'd consider using `combineReducers`, so you don't have to keep repeating `.childWindows` in this reducer (so it will only receive the state it is responsible for, not the entire state).

src/parent/reducers/reducers.js
```diff
@@ -17,18 +17,11 @@ function parentReducer(state = {}, action) {
...
-            } else {
-                newState = Object.assign({}, state, {
-                    [action.windowName]: childReducers({}, action)
-                });
-            }
+            const childState = state.childWindows && state.childWindows[action.windowName] || {};
+
+            const childWindows = Object.assign({}, state.childWindows);
+            childWindows[action.windowName] = childReducers(childState, action);
+            const newState = Object.assign({}, state, { childWindows });
```
jleft: If we go with the combineReducers option above, then I think this could be simplified to something like this:

```js
return Object.assign({}, state, {
    [action.windowName]: childReducers(state[action.windowName], action)
});
```

jleft:
I think the child window count checks in the ParentService will also need to be updates, given the change in state shape?

PR #893, opened 2016-08-24
feat: drag out favourite to spawn new window

jdhodges:
Replaces #863

### DIFF
--- a/src/child/index.js
+++ b/src/child/index.js
@@ -8,6 +8,9 @@ import currentWindowService from './services/currentWindowService';
 import 'babel-polyfill';
 
 import { open } from './actions/window';
+import { toggleFavourite } from './actions/favourites';
+import { selectFavourites } from './actions/sidebar';
+import { dragAccept } from '../parent/actions/parent';
 
 import './assets/styles/style.less';
 import '../../node_modules/d3fc/dist/d3fc.min.css';
@@ -31,6 +34,12 @@ currentWindowService.ready(() => {
     windowStateService.start();
 
     store.dispatch(open());
+
+    if (store.getState().dragOut) {
+        store.dispatch(toggleFavourite(store.getState().dragOut));
+        store.dispatch(selectFavourites());
+        store.dispatch(dragAccept());
+    }
 
     render(
         <Provider store={store}>
--- a/src/parent/actions/parent.js
+++ b/src/parent/actions/parent.js
@@ -6,3 +6,27 @@ export function close(windowName) {
         type: ACTION_TYPES.CLOSE
     };
 }
+
+function dragOut(code) {
+    return {
+        type: ACTION_TYPES.DRAG_OUT,
+        code
+    };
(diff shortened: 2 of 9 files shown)


src/child/index.js
```diff
@@ -22,16 +25,26 @@ require('script!../../node_modules/d3fc/dist/d3fc.bundle.min.js');
...
+
 currentWindowService.ready(() => {
     const currentWindow = currentWindowService.getCurrentWindow();
-    const store = currentWindow.contentWindow.opener.store;
     const rootElement = document.getElementById('app');
 
     const windowStateService = new WindowStateService(currentWindow, store, rootElement);
     windowStateService.start();
 
+    window.store = store;
```
jleft: I'm really not keen on storing the `store` as a global on the child windows (I'm also not keen in the parent, but it seemed a reasonable compromise to give the child access to the store). Is this even used?

src/child/index.js
```diff
@@ -22,16 +25,26 @@ require('script!../../node_modules/d3fc/dist/d3fc.bundle.min.js');
...
+    window.parent = parent;
```
jleft: Similar comment to above. Is this used?

src/parent/actions/parent.js
```diff
@@ -6,3 +6,23 @@ export function close(windowName) {
...
+export function dragAccept() {
+    return {
+        type: ACTION_TYPES.DRAG_ACCEPT
+    };
+}
+
+export function favouriteDroppedOutside(code, position) {
+    return (dispatch) => {
+        dispatch(dragOut(code));
+        window.opener.parentService.createChildWindow(null, position);
```
jleft: This feels a bit weird to me and I'm not to keen on using `window` and the `parentService` like this. This might actually be a good use case for using events/the inter-application messaging bus. Let's discuss.

src/parent/parent.js
```diff
@@ -6,4 +6,7 @@ fin.desktop.main(() => {
     const store = parentStore();
     const parentService = new ParentService(store);
     parentService.start();
+
+    window.parentService = parentService;
+    window.parent = window;
```
jleft: Why?

src/parent/reducers/reducers.js
```diff
@@ -1,10 +1,33 @@
-import { combineReducers } from 'redux';
+import { PARENT as ACTION_TYPES } from '../../shared/constants/actionTypes';
 
 import childWindows from './childWindows';
 
-const rootReducer = combineReducers({
-    childWindows
-});
+function rootReducer(state = { childWindows: {}, dragOut: null }, action) {
```
jleft: Can we keep `combineReducers` and have a `dragOut` reducer? See https://github.com/ScottLogic/StockFlux/pull/863#discussion_r72661223

I'm not sure, but I think `draggedOutStockCode` might be a bit more descriptive than `dragOut`?

PR #894, opened 2016-08-25
feat: drag favourite from window x to window y

jdhodges:
replaces #864


src/child/actions/favourites.js
```diff
@@ -17,3 +18,25 @@ export const quandlResponse = createActionCreator((code, name) => ({
...
+            dispatch(removeFavourite(dragStartWindow, code));
+        } else {
+            const application = fin.desktop.Application.getCurrent();
+            application.getChildWindows(children => {
+                children.find(childWindow => childWindow.name === dragStartWindow).close();
+            });
+        }
+
+        const insertAtIndex = getState().childWindows[currentWindowService.getCurrentWindowName()].favourites.codes.length;
+        dispatch(insertFavouriteAt(insertAtIndex, code));
```
rjmcneill: Could we do this instead?
```JavaScript
dispatch(toggleFavourite(stockCode));
```

Would mean we could remove the index line.

src/child/actions/favourites.js
```diff
@@ -17,3 +17,24 @@ export const quandlResponse = createActionCreator((code, name) => ({
...
+const removeFavourite = createActionCreator((dragStartWindow, code) => ({
+    type: ACTION_TYPES.REMOVE_FAVOURITE,
+    dragStartWindow,
+    code
+}));
+
+export function moveFavouriteFromWindow(dragStartWindow, code) {
+    return (dispatch, getState) => {
+        if (getState().childWindows[dragStartWindow].favourites.codes.length > 1) {
+            dispatch(removeFavourite(dragStartWindow, code));
```
jleft: Maybe I'm oversimplifying this, but wouldn't simply toggling the favourite in the `dragStartWindow` window solve this and remove all of the logic in the childWindows reducer below?

We'd need to add a new action creator to specify the window to toggle the favourite in, but that should be pretty simple:

```js
function toggleFavouriteInWindow(windowName, code) {
    return {
        type: ACTION_TYPES.TOGGLE_FAVOURITE,
        windowName,
        code
    };
}
```

Then `dispatch(toggleFavouriteInWindow(dragStartWindow, code))` instead of `dispatch(removeFavourite(dragStartWindow, code))`.

The else should handle closing the window and we already have the `CLOSE` action in place to handle this and delete the corresponding window state.

src/child/containers/sidebars/favourites/Favourites.js
```diff
@@ -48,6 +48,15 @@ class Favourites extends Component {
         });
     }
 
+    componentWillReceiveProps(nextProps) {
+        // onDragEnd is not called when a tile is successfully dragged into
```
jleft: I think it _should_ be called - is it definitely not being called? And if so, any idea why?

src/child/actions/favourites.js
```diff
@@ -1,19 +1,58 @@
...
+export function toggleFavourite(code, windowName) {
+    return (dispatch, getState) => {
+        const windowToToggleFavouriteIn = windowName || currentWindowService.getCurrentWindowName();
+        const { favourites, selection } = getState().childWindows[windowToToggleFavouriteIn];
+
+        dispatch(toggleFavouriteInWindow(code, windowToToggleFavouriteIn));
+
+        const isFavourite = favourites.codes.some(favourite => favourite === code);
+        if (selection.code === code && isFavourite) {
+            if (favourites.codes.length) {
```
jleft: I think this should be `favourites.codes.length > 1` because if there is only 1 favourite and it is unfavourited, then there will not be any favourites left to select. So there need to be at least 2 favourites to make a new selection (1 to unfavourite and the 1 to select as the new selection).

jleft:
Tests fixed in #898

PR #895, opened 2016-08-26
fix: compact/maximised works with state persisting

jdhodges:
Rebase of #874

### DIFF
--- a/src/child/actions/window.js
+++ b/src/child/actions/window.js
@@ -164,7 +164,7 @@ function resizeCompact() {
             currentWindowService.resizeTo(
                 compactWindowWidth,
                 compactWindowHeight,
-                'top-right',
+                'top-left',
                 () => resolve(dispatch(resizeCompactSuccess())),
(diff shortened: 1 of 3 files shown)


src/child/index.js
```diff
@@ -32,6 +33,19 @@ currentWindowService.ready(() => {
 
     store.dispatch(open());
 
+    const windowState = indexSelector(store.getState());
+
+    // Ensure that the window is started in the same
+    // state as the previous session
+    if (windowState.isCompact) {
```
jleft: Comment from original PR: https://github.com/ScottLogic/StockFlux/pull/874#discussion_r72294384

> Did you consider doing this before the window is created and shown?

> When creating the window, the previous window state (rather than the default) could be set in the [options object](http://cdn.openfin.co/jsdocs/stable/tutorial-window.constructor.html) passed to the window constructor; if anything can't be set in this option object and needs to be done (e.g. a method call), then I think this should be done before the window is shown.

PR #897, opened 2016-08-31
fix: Added hover styles to compact view

rjmcneill:
The current implementation has a hover style on the non-selected favourite, but not the selected one.
This PR fixes that and ensures hover styles on all favourites in compact view.

### DIFF
--- a/src/child/containers/sidebars/favourites/favourite-compact.less
+++ b/src/child/containers/sidebars/favourites/favourite-compact.less
@@ -9,5 +9,9 @@
         background: @highlight-colour;
       }
     }
+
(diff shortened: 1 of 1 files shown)


jleft:
The current behaviour is consistent with the Angular version (probably because you can not select a stock when in compact view), so I'm tempted not to merge now, but there is discussion in #729 about potentially changing this:

>I can imagine the compact view looking like it doesn't react much. We could could introduce the hover states, as you mentioned, but prevent it from being selected (as that would produce no effect).

Let's get some feedback before deciding, the change looks good to me though :smile:

PR #922, opened 2016-10-26
feat: add some default stocks on first window open

jleft:
Supersedes #919


jleft:
Thanks for reviewing @rjmcneill 👍

PR #931, opened 2016-12-12
fix: limit concurrent Quandl API requests

ColinEberhardt:
see: https://blog.quandl.com/change-quandl-api-limits


ColinEberhardt:
Thanks - that's a good point about BitFlux, I hadn't realised it made its own API requests.

PR #959, opened 2018-11-03
fix: replace deprecated Quandl WIKI database with EOD database

### LINKED ISSUE #958: Data Quality
## Problem

StockFlux uses the [WIKI database](https://www.quandl.com/databases/WIKIP) from Quandl, which is no longer supported by Quandl:

> WIKI is a community-maintained data feed based on public sources. Unfortunately, one of the main sources of that data is no longer available, and our community volunteers have been unable to find a suitable alternative source.
>
> As a result, the WIKI data feed is likely to be a lot less reliable in the future, with potentially missing or incorrect data or delayed updates.
> 
> We will continue to host this data feed on Quandl, but we no longer recommend using it for investment or analysis.
>
> Unfortunately, we don't have any free stock price data feeds that can replace WIKI at this time. The closest alternative we have is EOD, which is a value-priced end-of-day US stock price data feed.

### Ticker search

**Behaviour**: no results are returned for any search queries.

**Cause**: the database's newest stock price data is older than four weeks old. StockFlux filters-out any search results whose most recent price is not within the last four weeks.

image

###  Mini Charts

**Behaviour**: insufficient/no data to render mini charts.

**Cause**: the database's newest stock price data is older than four weeks old. StockFlux's mini charts render the last four weeks prices from the current day, as there is no data in this range, the chart cannot render.

image

### Main chart (BitFlux)

**Behaviour**: chart renders, but with out-of-date data.

**Cause**: the main chart renders using the available data (until the newest price date, rather than today's today).

image

### Use a different database

The [EOD database](https://www.quandl.com/data/EOD-End-of-Day-US-Stock-Prices) looks promising. We have an API key that allows us to use this database. As this is a premium data feed, requiring an API key, we should consider if this would impact local development (how would the API key(s) work). Should also investigate the impact on BitFlux.

### Use older data from the current WIKI database

We could relax the ticker search filter to allow tickers to be selected with data older than four weeks and render the minicharts with four weeks of data until the ticker's newest price date, rather than today's price.

## Other Considerations

Might be worth considering adding some form of monitoring to automatically alert us to issues with the application.

jleft:
Fixes #958

### DIFF
--- a/src/child/components/favourite/Favourite.js
+++ b/src/child/components/favourite/Favourite.js
@@ -142,7 +142,7 @@ class Favourite extends Component {
 
         const price = !isNaN(+stockData.price) ? (+stockData.price).toFixed(2) : null;
         const delta = !isNaN(+stockData.delta) ? (+stockData.delta).toFixed(2) : null;
-        const percentage = !isNaN(+stockData.percentage) ? Math.abs((+stockData.percentage)).toFixed(2) : null;
+        const percentage = !isNaN(+stockData.percentage) ? (+stockData.percentage).toFixed(2) : null;
         const name = stockData.name ? truncate(stockData.name) : '';
 
         const confirmationBindings = {
@@ -180,9 +180,9 @@ class Favourite extends Component {
                             <div className="details">
                                 {price && <div className="price">{price}</div>}
                                 {delta && <div className="delta">{delta}</div>}
-                                {percentage && <div className="percentage">
+                                {percentage !== null && <div className="percentage">
                                     <img src={percentage > 0 ? arrowUp : arrowDown} className="stock-arrow" draggable="false" />
-                                    {percentage}%
+                                    {Math.abs(percentage)}%
                                 </div>}
                             </div>
                         </div>
--- a/src/child/containers/showcase/Showcase.js
+++ b/src/child/containers/showcase/Showcase.js
@@ -1,12 +1,37 @@
 import React, { Component, PropTypes } from 'react';
-import { apiKey as quandlServiceApiKey } from '../../services/QuandlService';
+import { apiKey as quandlServiceApiKey, dataset as quandlServiceDataset } from '../../services/QuandlService';
 import configService from '../../../shared/ConfigService';
+
+const columnNameMap = {
+    Open: 'unadjustedOpen',
+    High: 'unadjustedHigh',
+    Low: 'unadjustedLow',
+    Close: 'unadjustedClose',
+    Volume: 'unadjustedVolume',
+    Adj_Open: 'open',
+    Adj_High: 'high',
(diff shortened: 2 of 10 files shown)


src/child/components/favourite/Favourite.js
```diff
@@ -142,7 +142,7 @@ class Favourite extends Component {
 
         const price = !isNaN(+stockData.price) ? (+stockData.price).toFixed(2) : null;
         const delta = !isNaN(+stockData.delta) ? (+stockData.delta).toFixed(2) : null;
-        const percentage = !isNaN(+stockData.percentage) ? Math.abs((+stockData.percentage)).toFixed(2) : null;
+        const percentage = !isNaN(+stockData.percentage) ? (+stockData.percentage).toFixed(2) : null;
```
msuperina: See comments below: I generally find less prone to errors to rely on optionals
```
const percentage = {
    exists: !isNaN(+stockData.percentage),
    value: Math.abs((+stockData.percentage).toFixed(2))
};
```

src/child/components/favourite/Favourite.js
```diff
@@ -180,9 +180,9 @@ class Favourite extends Component {
                             <div className="details">
                                 {price && <div className="price">{price}</div>}
                                 {delta && <div className="delta">{delta}</div>}
-                                {percentage && <div className="percentage">
+                                {percentage !== null && <div className="percentage">
```
msuperina: Personally, I would not rely on null or any specific value, as it needs to be synchronized with the test you make when defining `percentage`

src/child/components/favourite/Favourite.js
```diff
@@ -180,9 +180,9 @@ class Favourite extends Component {
...
                                     <img src={percentage > 0 ? arrowUp : arrowDown} className="stock-arrow" draggable="false" />
-                                    {percentage}%
+                                    {Math.abs(percentage)}%
```
msuperina: I would keep percentage defining `percentage` with `Math.abs`

src/child/containers/showcase/Showcase.js
```diff
@@ -1,12 +1,37 @@
...
+    Adj_High: 'high',
+    Adj_Low: 'low',
+    Adj_Close: 'close',
+    Adj_Volume: 'volume'
+};
+
+function mapColumnNames(colName) {
+    let mappedName = columnNameMap[colName];
+    if (!mappedName) {
+        mappedName = colName[0].toLowerCase() + colName.substr(1);
```
msuperina: What happens if in the future someone adds an entry to `columnNameMap` in such a way that 
```
columnNameMap[colName] === undefined
```
but 
```
colName[0].toLowerCase() + colName.substr(1) !== undefined
```

PR #977, opened 2018-11-08
fix: upgrade d3fc to version ^14.0.21

### LINKED ISSUE #962: Update d3fc to v14
Upgrade d3 and d3fc
Stop using globals and use module imports 

Minichart is the only component affected by d3fc upgrade and needs code change to use the new APIs.
Showcase is affected by module import of bitflux.

msuperina:
Fixes #962


package.json
```diff
@@ -79,7 +79,12 @@
     "BitFlux": "github:scottlogic/BitFlux#1.6.2",
     "babel-polyfill": "^6.3.14",
     "classnames": "^2.2.3",
-    "d3fc": "^5.3.0",
+    "d3-scale": "^2.1.2",
+    "d3-selection": "^1.3.2",
+    "d3fc": "^14.0.21",
+    "d3fc-discontinuous-scale": "^2.0.5",
```
ColinEberhardt: you shouldn't have to add these `d3fc-*` packages manually, they are all bundled up into `d3fc`

PR #982, opened 2018-11-15
chore: update build tools

### LINKED ISSUE #981: Upgrade build tools
At the moment the code is built with babel 6 and webpack 1.
It would be good to upgrade the build tools to latest versions in order to support more easily latest JS features.
Upgrading react-hot-loader may be difficult.

msuperina:
Fixes #981 

This PR aims at updating all the dependencies related to the build process.
react-hot-loader is removed as it is not used and causes conflicts when importing code from openfin-layouts


webpack/webpack.config.prod.js
```diff
@@ -15,15 +16,19 @@ config.plugins.push(
...
-        }
-    }),
     new CopyWebpackPlugin([
         { from: 'src/static' },
         { from: 'openfin-config/app.prod.json', to: './app.json' }
     ])
 );
+config.optimization = {
+    minimizer: [
+        new UglifyJsPlugin({
```
owennw: I thought webpack 4 in `mode: production` automatically adds in uglification.

webpack/webpack.config.shared.js
```diff
@@ -10,22 +10,19 @@ module.exports = {
         filename: './[name]_bundle.js'
     },
     module: {
-        loaders: [
-            { test: /\.js$/, exclude: /node_modules/, loader: 'react-hot!babel' },
+        rules: [
+            { test: /\.js$/, exclude: /node_modules(?!(\/|\\)openfin-layouts)/, loader: 'babel-loader' },
             { test: /\.less$/, loader: 'style-loader!css-loader!less-loader' },
```
owennw: It might be nicer to use `use: [ { loader: 'style-loader' }, { loader: 'css-loader' }, ... ]` here

webpack/webpack.config.prod.js
owennw: In fact I don't see anywhere where `mode` is being specified -- does this come with a warning?

msuperina: This is the npm build command `npm run clean && webpack -p --config webpack/webpack.config.prod.js` where production mode is applied.
Agree, let's use webpack defaults for now.

owennw:
Slight typo in the commit name there! Maybe there are some github hooks that can check the commit messages? :D

webpack/webpack.config.shared.js
msuperina: Given we don't apply any option to those loaders, don't you think we would have `loader` written all over the place for little benefit ?

webpack/webpack.config.prod.js
owennw: I mean this: https://webpack.js.org/concepts/mode/ I don't see that anywhere, and it should be specified (it's what turns on the minification etc by default)

webpack/webpack.config.shared.js
owennw: I think it's an aesthetic opinion. For me I prefer that syntax to chaining with `!`, but you may feel differently.

webpack/webpack.config.prod.js
msuperina: Documentation says 
> Possible values for mode are: none, development or production(default).
```
Usage 
Just provide the mode option in the config:
module.exports = {
  mode: 'production'
};
or pass it as a CLI argument:
webpack --mode=production
```
The latter is what we do in the build command
```
npm run clean && webpack -p --config webpack/webpack.config.prod.js
```

owennw:
Looks fine to me

PR #983, opened 2018-11-26
Openfin layouts release

msuperina:


src/child/index.js
```diff
@@ -33,6 +34,8 @@ currentWindowService.ready(() => {
     const store = currentWindow.contentWindow.opener.store;
     const rootElement = document.getElementById('app');
 
+    startLayoutsService(store);
```
owennw: Given both of these layout services take the store, and look at actions, could they be expressed as middleware and folded into the store creation?

src/parent/services/LayoutsService.js
```diff
@@ -0,0 +1,70 @@
...
+
+const uuid = fin.desktop.Application.getCurrent().uuid;
+
+export function start(store) {
+    const layoutsStore = LayoutsStore();
+
+    fin.desktop.InterApplicationBus.subscribe(uuid, 'join-snap-group', updateGroupsAndNotify);
+    fin.desktop.InterApplicationBus.subscribe(uuid, 'leave-snap-group', updateGroupsAndNotify);
+
+    store.subscribe(updateGroupsAndNotify);
```
owennw: Took me a while to find this function -- would've been easier to locate if this were below the function declaration

src/parent/services/LayoutsService.js
```diff
@@ -0,0 +1,70 @@
...
+            {}
+        );
+
+        const nextCounts = sameGroupCounts();
+        return diff(previousCounts, nextCounts);
+    }
+
+    function sameGroupCounts() {
+        return Object.keys(sameGroup || {}).reduce(
+            (counts, key) => (counts[key] = sameGroup[key].length) && counts,
```
owennw: An assignment into an operator :o! Not my personal cup of tea.

src/parent/services/LayoutsService.js
```diff
@@ -0,0 +1,70 @@
...
+            {}
+        );
+    }
+
+    function diff(previousCounts, nextCounts) {
+        return {
+            joined: Object.keys(nextCounts).filter(key => nextCounts[key] !== previousCounts[key] && nextCounts[key] > 1),
+            left: Object.keys(nextCounts).filter(key => nextCounts[key] !== previousCounts[key] && nextCounts[key] === 1)
```
owennw: You could store `Object.keys(nextCounts)` above this and only do it once, if you felt the need.

src/child/services/LayoutsService.js
```diff
@@ -0,0 +1,32 @@
+import * as Layouts from 'openfin-layouts';
```
owennw: I don't see any unit tests for this file, can you add some?

src/parent/services/LayoutsService.js
```diff
@@ -0,0 +1,70 @@
+import * as Layouts from 'openfin-layouts';
```
owennw: Can you add unit tests for this function?

src/child/index.js
msuperina: I am applying the same pattern as the other services which all get injected the store.
However those service are special: they deal with window layouts, which should not be part of the store.

owennw: Fair enough!

PR #1002, opened 2018-12-06
Fix linter complaints about CRLF line endings

### LINKED ISSUE #999: Stop eslint complaining about CRLF line endings
We have a `.gitattributes` file in the repo specifying which files are text and should have line-endings _normalized_, i.e. modified from CRLF to LF, when committing.

However, we are extending the [airbnb rules](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/style.js#L176) which specify line-endings should always be LF.

We need to override this by adding `"linebreak-style": 0` to the rules in our project `.eslintrc`, to prevent eslint marking CRLF as an error.

chriswilty:
ESLint is complaining about CRLF line endings. This is because we are extending airbnb rules, which include a rule to error on non-unix line endings.

We already have a `.gitattributes` file to normalize CRLF to LF when creating a commit, and we are all developing on Windows OS, so it makes sense to override the airbnb rule.

fixes #999

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -22,6 +22,7 @@
     "no-console": 0,
     "indent": ["error", 4],
     "comma-dangle" : 0,
+    "linebreak-style": 0,
     "prefer-template": 1,
     "padded-blocks": 0,
     "max-len": 0,
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,6 @@
 # These files are text and should be normalized (convert crlf -> lf)
 *.js      text
+*.jsx     text
 *.md      text
 *.css     text
 *.html    text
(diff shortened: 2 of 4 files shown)


.eslintrc
```diff
@@ -22,6 +22,7 @@
     "no-console": 0,
     "indent": ["error", 4],
     "comma-dangle" : 0,
+    "linebreak-style": 0,
```
chriswilty: This is the only important bit!

msuperina:
> we are all developing on Windows OS
This is not true. Some people in Newcastle use mac, some use a linux box.

.eslintrc
AndySmithSL: So what does that do, does it disable the ESLint check? Would be good as I'm getting lint errors all over the place.

chriswilty: Yep, exactly that :)

PR #1008, opened 2019-01-03
Fix close button issues due to nonspecific css

chriswilty:
Some css issues fixed here, for the window close button.

In particular, the non-specific use of `.close { display: none; }` was causing havoc!


chriswilty:
@AndySmithSL Sorry, you'll need to approve this one again as I needed to rebase and force-push. The remaining two PRs I will open only after rebasing on dev.

Issue #1006, opened 2019-01-03
Closed event fires more than once for reopened windows


chriswilty:
**Background**
When a (child) application window is closed by the user, a `closed` event is fired which is intercepted by the parent, which in turn dispatches an action to update the state to reflect the change in currently open and closed windows. Up to five most recently closed windows can be viewed and re-opened by clicking the _closed windows_ icon.

**Bug description**
It seems that closed windows are not having their event listeners removed, because the number of times the `closed` event is fired equals the number of times a window has previously been closed.

**Solution**
An event listener is attached to each child openfin Window object in `src/parent/services/ParentService.js`. The listener function does not have access to the openfin window that triggered the event (which seems like an odd omission from the openfin API) but it does have the window name, so in order to clean up within the event listener we would likely need to store a mapping of name : window within the ParentService, add an entry whenever a window is created (see `createChildWindowSuccess`), then look up the window in the `closed` event listener, remove the listener from the window and remove the entry from the map.

PR #1009, opened 2019-01-04
Temporary workaround for issue #1006

chriswilty:
Step 1 of fixing the bug described in #1006, in which a closed event is fired more than once for re-opened windows. A future PR will fix this properly.

### DIFF
--- a/src/parent/reducers/reducers.js
+++ b/src/parent/reducers/reducers.js
@@ -7,6 +7,8 @@ import childWindows from './childWindows';
 import closedWindows from './closedWindows';
 import dragOut from './dragOut';
 
+const MAX_CLOSED_WINDOWS = 5;
+
 const rootReducer = reduceReducers(
     combineReducers({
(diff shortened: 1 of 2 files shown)


src/parent/reducers/reducers.js
```diff
@@ -33,30 +35,32 @@ const rootReducer = reduceReducers(
...
         case ACTION_TYPES.CLOSE: {
-            const newClosedWindow = Object.assign({}, state.childWindows[action.windowName], {
-                date: action.date
-            });
-
-            const newChildWindows = Object.assign({}, state.childWindows);
-
-            delete newChildWindows[action.windowName];
+            // FIXME Solve why close event is fired twice when re-opened window is closed! Event listener not removed?
+            if (!state.childWindows[action.windowName]) return state;
```
chriswilty: This is the workaround; the rest is code sanitising and refactoring into ES syntax.

PR #1018, opened 2019-04-10
refactor: use openfin-layouts dock events

mboyd-sl:


mboyd-sl:
Closing as additional handling would be required in the parent LayoutsService

PR #1020, opened 2019-05-01
Chart standalone

JackG188:
Initial version of the standalone chart app that uses random generated data

TODO

- remove BitFlux css that is unused or has been overridden 
- extract common css variables out for use across all 3 standalone apps


stockflux-chart/package.json
```diff
@@ -3,6 +3,9 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "bootstrap": "^3.3.6",
+    "BitFlux": "github:scottlogic/BitFlux#1.8.1",
+    "grunt-cli": "^1.2.0",
```
oriondean: this should probably be a dev dependency

stockflux-chart/src/index.js
```diff
@@ -1,6 +1,11 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import App from './App';
+import '../node_modules/malihu-custom-scrollbar-plugin/jquery.mCustomScrollbar.css';
```
oriondean: Do you need this custom scrollbar behaviour?

stockflux-chart/src/styles/BitFlux/notification.css
```diff
@@ -0,0 +1,25 @@
+#notifications {
```
oriondean: is this used anywhere?

stockflux-chart/package.json
markjose: What is this used for, can we remove it. Feels weird using grunt when we've moved to react scripts?

markjose: Also can we add a task to the shoet term backlog to moved styles in to the common components module which would includr our dep on bootstrap.

stockflux-chart/src/index.js
markjose: Bin it, speak to @oriondean about his approach in the watch list.

stockflux-chart/src/components/Chart.js
```diff
@@ -1,9 +1,23 @@
...
 
-export default () => (
-  <div className={styles.chart}>
-    Chart Not implemented
-  </div>
-);
+        chart.periodsOfDataToFetch(1200);
+        chart.proportionOfDataToDisplayByDefault(112/1200);
+        // using random generated data
+        chart.run(showcaseContainer);
```
markjose: This will run on every render, is that correct? Moving forward we'll want to stream the data from the backend service.

stockflux-chart/src/index.js
```diff
@@ -1,6 +1,11 @@
...
+
+import 'jquery/src/jquery';
+import 'malihu-custom-scrollbar-plugin/jquery.mCustomScrollbar.js';
```
markjose: See above, but also seems to imported twice, can we remove all of these excet the react imports?

stockflux-chart/package.json
JackG188: It's used to build BitFlux post install

PR #1021, opened 2019-05-02
WIP Search 

oriondean:
Felt like i should get this merged before a larger refactor of the search application begins. This will atleast allow people to try out the search application if they please, even if it isn't the finished deal.

### DIFF
--- a/stockflux-search/public/app.dev.json
+++ b/stockflux-search/public/app.dev.json
@@ -1,7 +1,7 @@
 {
     "runtime": {
         "arguments": "--enable-aggressive-domstorage-flushing",
-        "version": "9.61.38.40"
+        "version": "10.66.39.43"
     },
     "startup_app": {
         "name": "stockflux-search",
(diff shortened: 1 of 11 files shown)


stockflux-search/public/app.dev.json
```diff
@@ -1,7 +1,7 @@
 {
     "runtime": {
         "arguments": "--enable-aggressive-domstorage-flushing",
-        "version": "9.61.38.40"
+        "version": "10.66.39.43"
```
markjose: Think we can actually use "stable" in here, but that may be a risk if there're breaking changes. It'd be nice to think that this repo will be kept at the bleeding edge going forward though.

stockflux-search/src/components/Search.js
```diff
@@ -106,47 +90,41 @@ export default function Search(props) {
...
-              An error occurred while retrieving data. Please check your internet connection or wait for our data services to be re-established.
-                        </div>
+              <p>Use the search tab to add new stocks to the list</p>
+            </div>
           }
 
           {isSearching && <div className={styles.message}>Loading search results...</div>}
 
-          {!isSearching && !results && codes.map((stockCode) =>
+          {(results || []).map((stock) =>
```
markjose: no parentheses needed around stock param

PR #1023, opened 2019-05-06
Initial working version of watchlist standalone

AugusteBa:
Creating this PR to get this merged before a larger refactor of the watchlist application. This will allow others to use watchlist standalone if they need.


stockflux-watchlist/src/App.js
```diff
@@ -1,6 +1,7 @@
-import React from 'react';
-import Watchlist from './components/Watchlist';
-import Components from 'stockflux-components';
+import React from "react";
+import { Watchlist } from "./components/Watchlist/Watchlist";
```
jrhenderson1988: It seems that `WatchList` is the default export, so you could remove the `{` and `}` here

AugusteBa: Indeed, thanks for spotting! Fixed -> 13d11dfc1caba9c5e79b0182a6313153503d112e

stockflux-watchlist/src/services/formatters.js
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable import/prefer-default-export */
+
+export const truncate = input => {
```
jrhenderson1988: Just a note on this. I noticed that the current frontend apps were stripping down stock names themselves by removing everything from the first occurrence of `(`. This is because the ticker symbol was included in the name e.g. *Apple Inc. (AAPL)* in the data provided by Quandl. Going forward this should no longer be necessary, as the backend will take care of it, since other souces don't include the ticker in the name (like IEX Trading). For now I think it's alright to leave this in but when we start connecting the frontend apps to the new backend, this can be removed.

stockflux-watchlist/package.json
```diff
@@ -1,39 +1,43 @@
...
-  }
+    "name": "stockflux-watchlist",
+    "version": "0.1.0",
+    "private": true,
+    "dependencies": {
+        "classnames": "^2.2.6",
+        "d3-scale": "^2.1.2",
+        "d3-selection": "^1.3.2",
+        "d3fc": "^14.0.21",
+        "imports-loader": "^0.8.0",
```
oriondean: Is this a necessary dependency?

stockflux-watchlist/public/app.dev.json
```diff
@@ -1,37 +1,37 @@
...
-        {
-            "name": "layouts",
-            "config": {
-                "features": {
-                    "tab": false
-                }
-            }
+  "runtime": {
+    "arguments": "--enable-aggressive-domstorage-flushing",
+    "version": "9.61.38.40"
```
oriondean: Feel free to bump this up to latest stable (e.g. 10.66.39.43)

stockflux-watchlist/src/components/Minichart/Minichart.js
```diff
@@ -0,0 +1,156 @@
...
+  return seriesSvgPoint()
+    .crossValue(d => d.date)
+    .mainValue(d => d.close);
+};
+
+const getTrimmedData = chartData => {
+  return chartData.map(d => {
+    const datum = d;
+    const date = moment(d.date);
+    datum.date = date.toDate();
```
oriondean: This map looks like it's modifying the existing chartData rather than just purely mapping it into a new object

AugusteBa: Thanks for noticing, fixed -> a03e818b1a82eea4f6e3cba19b30cd030588ee23

stockflux-watchlist/public/app.dev.json
AugusteBa: Fixed -> a03e818b1a82eea4f6e3cba19b30cd030588ee23

stockflux-watchlist/package.json
AugusteBa: No, it isn't, thanks! -> a03e818b1a82eea4f6e3cba19b30cd030588ee23

PR #1028, opened 2019-05-14
Added chart intent listener, refactor chart

JackG188:


stockflux-chart/src/App.js
```diff
@@ -24,12 +25,32 @@ import './styles/BitFlux/notification.css';
 import './styles/app.css';
 import './styles/BitFlux/sprite.css';
 
+let isListening = false;
```
AugusteBa: Could we use React hooks for `isListening`?

stockflux-chart/src/styles/app.css
```diff
@@ -37,9 +37,9 @@ body {
     cursor: pointer;
 }
 
-.disable-transitions {
+/* .disable-transitions {
```
AugusteBa: Is this commented out intentionally?

stockflux-chart/src/App.js
JackG188: Leaving it out as we don't want it to trigger an additional render

PR #1040, opened 2019-05-16
Error handling when unable to fetch data + spawn charts on a watchlist card click

AugusteBa:


stockflux-core/src/intents/viewChart.js
```diff
@@ -0,0 +1,11 @@
+import * as fdc3 from 'openfin-fdc3';
+
+export const viewChart = async function(symbol, stockName) {
+  await fdc3.raiseIntent(fdc3.Intents.VIEW_CHART, {
```
oriondean: Can this also be used by the searchbar in `SearchResult.ts`?

PR #1047, opened 2019-05-21
Watchlist Notifications

AugusteBa:
Watchlist without a name:
image

Watchlist with a name:
image

Not sure if I love or hate the animation (might be kind of irritating having something moving at the bottom of the screen, but at the same time it is kind of cool since it attracts user attention. Any feedabck would be much appreciated.
 
notifications

Icons were created as a temporary placeholder and are subject to change if we get any UX designers from Edinburgh help (or anyone who can make them nicely)


stockflux-watchlist/public/notification.css
```diff
@@ -0,0 +1,81 @@
...
+.notification-icon {
+  margin: auto;
+  border-right: 1.5px solid #45d6bc;
+}
+
+span#message,
+span#symbol,
+p#title {
+  color: lightgray;
+  font-family: 'Roboto', sans-serif;
```
oriondean: Do we get Roboto without importing from googlefonts?

stockflux-watchlist/src/components/Watchlist/Watchlist.js
```diff
@@ -68,10 +69,35 @@ const Watchlist = () => {
...
       setWatchlist(getDistinctElementArray([newSymbol, ...watchlist]));
+      displayNotification(newSymbol);
     }
   });
   latestListener = currentListener;
 
+  const displayNotification = newSymbol => {
+    const alreadyInWatchlist = watchlist.includes(newSymbol);
+    const watchlistName = name.slice().toUpperCase();
+    if (alreadyInWatchlist) {
```
oriondean: This might be more succinct than the if-else:

```
showNotification({
        message: {
          symbol: newSymbol,
          watchlistName: name.toUpperCase(),
          alreadyInWatchlist,
          messageText: ` ${alreadyInWatchlist ? 'moved' : 'added'} to the top`
        }
      });
```

stockflux-watchlist/public/notification.css
AugusteBa: Nope, this is what happens when I remove it -> 
image

stockflux-watchlist/src/components/Watchlist/Watchlist.js
AugusteBa: Absolutely, thanks for noticing, fixed -> 27269452380451083d247e8139fb6ca741e28185

stockflux-watchlist/public/notification.css
AugusteBa: I misunderstood you initially, sorry, here's the fix and thanks -> f035838fc3d63a4f22f7bc2367bf2d96fe61d96c.

PR #1051, opened 2019-05-23
Reduce chart size, fix missing spinner

JackG188:


stockflux-chart/src/App.js
```diff
@@ -5,6 +5,7 @@ import {InterApplicationBusHooks} from 'openfin-react-hooks';
 
 import 'stockflux-bitflux/node_modules/d3fc/dist/d3fc.css';
 import 'bootstrap/dist/css/bootstrap.css';
+import 'stockflux-components/src/components/spinner/Spinner.css';
```
AugusteBa: Does it not see the spinner CSS, when you do `npm run build` in stockflux-components? Ideally, we never want to import any individual files from stockflux-components as we list it as a dependency in package.json.

PR #1052, opened 2019-05-24
Close the chart container when the last chart window is closed

JackG188:
This will prevent instances of the chart container sitting idle as the user is unable to close the container manually

### DIFF
--- a/stockflux-chart-container/src/App.js
+++ b/stockflux-chart-container/src/App.js
@@ -3,11 +3,12 @@ import * as fdc3 from 'openfin-fdc3';
 import {InterApplicationBusHooks} from 'openfin-react-hooks';
 
 let latestListener;
+let currentListener;
+let windows = [];
 
 function App() {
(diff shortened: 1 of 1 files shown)


stockflux-chart-container/src/App.js
```diff
@@ -27,23 +28,32 @@ function App() {
...
-          newWindows.splice(newWindows.indexOf(windowName));
-          setWindows(newWindows);
-        })
-      });
+      }
     }
-  });
+  }
+  
+  const wasFinalChartWindow = (windowName) => {
```
oriondean: This method seems to be doing two things (removing a window as a side effect)

stockflux-chart-container/src/App.js
```diff
@@ -27,23 +28,32 @@ function App() {
...
+    windows = windows.filter(name => name !== windowName);
+    return windows.length === 0 ? true : false;
```
oriondean: return `windows.length === 0`?

PR #1057, opened 2019-06-03
Created Circle CI config file 

JackG188:
Currently running on all branches, will add in filter to only run on dev when CRA-WIP is merged in

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,8 +6,8 @@ workflows:
     jobs:
       - install_and_build
       - deploy:
-        requires:
-          - install_and_build
+          requires:
+            - install_and_build
 
 jobs:
   install:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,10 +8,9 @@ workflows:
       - deploy:
         requires:
           - install_and_build
-        branches:
-          ignore:
-            - dev
-            - master
+        filters:
+          branches:
+            ignore: dev
 
 jobs:
   install:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,91 @@
...
+    docker:
+      - image: circleci/node:10.15.1
+    steps:
+      - checkout:
+          path: ~/stockflux
+      - run:
+          name: Build Components
+          command: |
+            cd ~/stockflux/stockflux-components
+            npm install
```
jrhenderson1988: Did you mean to do `npm ci` here?

.circleci/config.yml
```diff
@@ -0,0 +1,91 @@
...
+            npm run build
+      - run:
+          name: Build BitFlux
+          command: |
+            cd ~/stockflux/stockflux-bitflux
+            sudo npm install -g grunt-cli
+            npm install
```
jrhenderson1988: and here

.circleci/config.yml
```diff
@@ -0,0 +1,91 @@
...
+      - image: circleci/python:2.7-jessie
+    steps:
+      - attach_workspace:
+          at: ~/stockflux
+      - run:
+          name: Install AWS CLI
+          command: pip install awscli --user --upgrade
+      - run:
+          name: Upload chart
+          command: ~/.local/bin/aws s3 sync stockflux-chart/build s3://stockflux-public/artifacts/stockflux-chart/v0.0.1/ --delete
```
jrhenderson1988: I ran into problems when I didn't specify the region on the command line when running these commands. Could be worth adding `--region eu-west-2` to the end of these lines to make sure you don't get the same issue.

PR #1149, opened 2019-06-26
First draft of readme

JackG188:


AugusteBa:
> Maybe we should include the ports on which the apps run as well in this readme?

Nvm, just noticed they are in.

PR #1219, opened 2019-07-16
Add securities and edit form created along with routing from table

MaxOvenden:
A router has been set up from the securities table to the newly created form component for editing or adding components.

### DIFF
--- a/stockflux-securities-master/src/components/InputForm.js
+++ b/stockflux-securities-master/src/components/InputForm.js
@@ -1,21 +1,21 @@
-import React, { useRef } from "react";
+import React, { useState } from "react";
 import { Link } from "react-router-dom";
 import "./InputForm.css";
 
 const InputForm = () => {
-  const nameRef = useRef(null);
-  const exchangeRef = useRef(null);
-  const symbolRef = useRef(null);
-  const isShownRef = useRef(true);
-  const isVisibleRef = useRef(true);
+  const [name, setName] = useState('');
+  const [exchange, setExchange] = useState('');
+  const [symbol, setSymbol] = useState('');
+  const [isShown, setIsShown] = useState(false);
+  const [isEnabled, setIsEnabled] = useState(false);
 
   const submitForm = () => {
     const securityObject = {
-      name: nameRef.current,
-      exchange: exchangeRef.current,
-      symbol: symbolRef.current,
-      isShown: isShownRef.current,
-      isVisible: isVisibleRef.current
+      name: name,
+      exchange: exchange,
+      symbol: symbol,
+      isShown: isShown,
+      isEnabled: isEnabled
     };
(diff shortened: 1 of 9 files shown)


stockflux-securities-master/src/components/InputForm.js
```diff
@@ -0,0 +1,101 @@
+import React, { useRef } from "react";
+import { Link } from "react-router-dom";
+import "./InputForm.css";
+
+const InputForm = () => {
+  const nameRef = useRef(null);
```
jrhenderson1988: The [official recommendation regarding refs](https://reactjs.org/docs/refs-and-the-dom.html) is to avoid them where possible. Instead of using refs, you should use state to manage the data flow between DOM elements and the local component state. See [here](https://reactjs.org/docs/forms.html) for examples on how to do this, and [here](https://reactjs.org/docs/hooks-state.html) for documentation on how to manage state using the state hook.

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -15,50 +17,54 @@ const SecuritiesTable = () => {
...
-    }).catch(() => {
-      setIsLoading(false);
-    });
+    getSecuritiesData()
+      .then(securities => {
+        setIsLoading(false);
+        setSecuritiesData(securities);
+      })
+      .catch(() => {
+        setIsLoading(false);
```
jrhenderson1988: It would be nice to handle errors here, maybe show some kind of error message somewhere if grabbing the securities data fails.

MaxOvenden:
Here's the UX for reference:

image

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -0,0 +1,101 @@
...
+          <label className="toggle-switch" for="is-enabled-toggle" />
+        </div>
+        <div className="input-submit-button-container">
+          <div className="input-submit-button">
+            <button type="submit">Create</button>
+          </div>
+        </div>
+      </form>
+
+      <Link to="/home">
```
jrhenderson1988: Can we change this to `/` instead of `/home` to match our defined routes?

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: Okay, I've added a message now: 

image

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: Yeah changed now

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: Okay I've made the change now

PR #1220, opened 2019-07-17
Adding services for securities table

MaxOvenden:
Created a securitiesService file and moved the getSecuritiesData API call from the component to this file.
The event prevent default has also been added to the form submit to prevent the page from refreshing each time.

### DIFF
--- a/stockflux-securities-master/src/components/InputForm.css
+++ b/stockflux-securities-master/src/components/InputForm.css
@@ -104,11 +104,19 @@
   width: 280px;
 }
 
-.input-submit-button > button:focus {
+.input-submit-button.in-progress > button {
   outline: none;
   color: var(--text-grey-darkest);
   border-color: var(--text-grey-darkest);
 }
+
+.input-submit-button > button:focus {
+  outline: none;
+}
+
+
+
+
 
 .back-button {
   position: fixed;
--- a/stockflux-securities-master/src/components/InputForm.css
+++ b/stockflux-securities-master/src/components/InputForm.css
@@ -114,10 +114,6 @@
   outline: none;
 }
 
-
-
-
-
 .back-button {
   position: fixed;
   top: 30px;
--- a/stockflux-securities-master/src/components/InputForm.js
+++ b/stockflux-securities-master/src/components/InputForm.js
@@ -3,21 +3,28 @@ import { Link } from "react-router-dom";
 import "./InputForm.css";
 
 const InputForm = () => {
-  const [name, setName] = useState('');
-  const [exchange, setExchange] = useState('');
-  const [symbol, setSymbol] = useState('');
+  const [name, setName] = useState("");
+  const [exchange, setExchange] = useState("");
+  const [symbol, setSymbol] = useState("");
   const [isShown, setIsShown] = useState(false);
   const [isEnabled, setIsEnabled] = useState(false);
-  
+  const [inProgress, setInProgress] = useState(false);
 
-  const submitForm = () => {
+  const submitForm = event => {
+    event.preventDefault();
+    setInProgress(true);
+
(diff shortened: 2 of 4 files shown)


stockflux-securities-master/src/services/SecuritiesService.js
```diff
@@ -0,0 +1,8 @@
+export async function getSecuritiesData() {
+  //temporarily using local host, will need to change to the AWS location
+  const response = await fetch(
+    "https://g3amedujs2.execute-api.eu-west-2.amazonaws.com/dev/api/securities-v2"
```
jrhenderson1988: Can we look at using the custom data to get the base URL of the API, rather than hard-coding it? You might need to do a bit of configuration locally to get it to use the domain you want. Take a look at the `StockfluxService` that the other guys wrote for details on how to do this.

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -26,9 +20,10 @@ const SecuritiesTable = () => {
       .then(securities => {
         setIsLoading(false);
         setSecuritiesData(securities);
-        setErrorMessage(null)
+        setErrorMessage(null);
       })
-      .catch(() => {
+      .catch((e) => {
+        console.log(e);
```
jrhenderson1988: You don't need this `console.log` any more. It was just to illustrate what would come through in the event of an error.

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -85,10 +89,14 @@ const InputForm = () => {
             type="checkbox"
             onChange={() => setIsEnabled(!isEnabled)}
           />
-          <label className="toggle-switch" for="is-enabled-toggle" />
+          <label className="toggle-switch" htmlFor="is-enabled-toggle" />
         </div>
         <div className="input-submit-button-container">
-          <div className="input-submit-button">
+          <div
+            className={
```
jrhenderson1988: I think this might be down to prettier, but can we keep the same style of formatting used in the other apps?

    <div
      className={"input-submit-button" + (inProgress ? " in-progress" : "")}
    >

It could be worth using [classnames](https://www.npmjs.com/package/classnames) to simplify this. It seems to be used in the launcher and watchlist apps.

stockflux-securities-master/src/components/InputForm.css
```diff
@@ -104,12 +104,20 @@
...
   outline: none;
   color: var(--text-grey-darkest);
   border-color: var(--text-grey-darkest);
 }
 
+.input-submit-button > button:focus {
+  outline: none;
+}
+
+
```
jrhenderson1988: Can we remove this excess space, please?

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: Okay I'll bear this in mind for the future. It might be worth turning prettier off as it sometimes makes some wild choices!

Regarding classnames, Jack told me to use them when I started but Mark said he preferred we didn't.

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: Yep understood

stockflux-securities-master/src/components/InputForm.css
MaxOvenden: done

stockflux-securities-master/src/services/SecuritiesService.js
MaxOvenden: Okay this has been changed and I didn't commit the app.dev.json file change which will correspond to this.

PR #1228, opened 2019-07-19
Make Get Security Service to Fill Edit Form

MaxOvenden:
The service for getting a single security has been added and is used in the edit form to fill the form with the security data for the selected security. As discussed the exchange and symbol are read-only in the edit form.

### DIFF
--- a/stockflux-securities-master/src/components/InputForm.css
+++ b/stockflux-securities-master/src/components/InputForm.css
@@ -32,17 +32,26 @@
   font-weight: var(--font-weight-bolder);
   margin: var(--margin-small);
   flex-basis: 80px;
-  }
+}
 
+.input-form-input-read-only,
 .input-form-input {
   font-size: var(--font-large);
   font-weight: var(--font-weight-normal);
-  color: var(--text-teal-light);
   background-color: var(--dark-background);
   border-radius: 25px;
-  border: solid 1px var(--text-teal);
-  padding: var(--padding-small) var(--padding-medium) ;
+  padding: var(--padding-small) var(--padding-medium);
   margin: var(--margin-small);
+}
+
+.input-form-input {
+  color: var(--text-teal-light);
+  border: solid 1px var(--text-teal);
+}
+
+.input-form-input-read-only {
+  color: var(--text-grey);
+  border: solid 1px var(--text-grey-darkest);
 }
 
 .input-form-input:focus {
@@ -110,6 +119,10 @@
   border-color: var(--text-grey-darkest);
 }
 
+.input-submit-button > button:hover {
+  color: var(--text-teal);
+}
+
 .input-submit-button > button:focus {
   outline: none;
 }
@@ -132,6 +145,10 @@
   border-radius: 20px;
   height: 30px;
   width: 70px;
+}
+
+.back-button > button:hover {
+  color: var(--text-teal);
 }
 
 .back-button > button:focus {
--- a/stockflux-securities-master/src/components/InputForm.css
+++ b/stockflux-securities-master/src/components/InputForm.css
@@ -34,8 +34,7 @@
   flex-basis: 80px;
 }
 
-.input-form-input-read-only,
-.input-form-input {
+.input-form {
   font-size: var(--font-large);
   font-weight: var(--font-weight-normal);
   background-color: var(--dark-background);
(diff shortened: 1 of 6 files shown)


stockflux-securities-master/src/components/InputForm.js
```diff
@@ -30,75 +59,90 @@ const InputForm = ({ match }) => {
...
+            />
           </div>
-        </div>
-      </form>
+          <div className="input-row">
+            <label className="input-label" htmlFor="exchange-input">
+              Exchange
+            </label>
+            <input
+              className="input-form-input"
```
jrhenderson1988: As discussed, can you add a class to these fields to make them appear disabled rather than editable, when we're editing?

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -30,75 +59,90 @@ const InputForm = ({ match }) => {
...
+              id="exchange-input"
+              value={exchange}
+              readOnly={match.params.securityId}
```
jrhenderson1988: Similar to our discussion yesterday with the `{condition && <Component />}` thing, I'm not sure this looks right. It will probably work fine since if there's a `securityId` then the value will be *truthy* and if there isn't one it will be `undefined` or `null` and therefore *falsy*. Is this something that you have been advised to do before? Personally, I would feel more comfortable casting it to boolean by wrapping it in `Boolean(match.params.securityId)` or `!!match.params.securityId`. I'd be interested to know what the most common approach is though.

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: Done this now

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: Following our discussion with Matt, I've changed them both to `!!match.params.securityId` .

stockflux-securities-master/src/components/InputForm.css
```diff
@@ -32,19 +32,28 @@
   font-weight: var(--font-weight-bolder);
   margin: var(--margin-small);
   flex-basis: 80px;
-  }
+}
 
+.input-form-input-read-only,
```
jrhenderson1988: I think the better approach here, which I think is more common, would be to have a base set of styling on the `.input-form-input` class which would dictate the style of the input in its default state and then you would *add* the `.input-form-input-read-only` class to it as a modifier which would provide some additional styling, overriding colours etc. In your HTML/JSX you should have the base class on all inputs and only add the extra class when you need to (in this case, when something is read only)

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -1,25 +1,54 @@
...
+          setErrorMessage("Error, cannot get security");
+        });
+    }
+  }, []);
 
   const submitForm = event => {
     event.preventDefault();
     setInProgress(true);
 
     const securityObject = {
```
jrhenderson1988: Just a tip regarding ES6 here. If the name of the property in an object matches the name of the variable, you can simply omit the value and it will have the same effect. For example, this object could become:

    const securityObject = { name, exchange, symbol, visible, enabled };

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -30,75 +59,94 @@ const InputForm = ({ match }) => {
...
+        <ErrorMessage errorMessage={errorMessage} />
+      ) : (
+        <form className="input-form-body" onSubmit={submitForm}>
+          <div className="input-row">
+            <label className="input-label" htmlFor="exchange-input">
+              Exchange
+            </label>
+            <input
+              className={`input-form-input${
+                match.params.securityId ? "-read-only" : ""
```
jrhenderson1988: See my comment above when I suggested having the `input-form-input` class always present on inputs and adding the `*-read-only` modifier class when appropriate.

MaxOvenden: yep done

stockflux-securities-master/src/components/InputForm.css
MaxOvenden: yep done

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: yeah good spot there, changed now

PR #1230, opened 2019-07-19
Created ToggleSwitch and TextField reusable components for use in form

MaxOvenden:
The toggle switches and the text input fields have been moved into reusable components to help with scalability as the project progresses


stockflux-securities-master/src/components/reusableComponents/ToggleSwitch.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import "./ToggleSwitch.css";
+
+const TextField = ({ value, id, checked, onChange }) => {
```
jrhenderson1988: Shouldn't this be `ToggleSwitch`?

stockflux-securities-master/src/components/reusableComponents/ToggleSwitch.js
```diff
@@ -0,0 +1,20 @@
...
+    <div>
+      <input
+        value={value}
+        id={id}
+        checked={checked}
+        type="checkbox"
+        className="input-form-toggle"
+        onChange={onChange}
+      />
+      <label className="toggle-switch" htmlFor={id} />
```
jrhenderson1988: This label is empty, should it have a value or is it not meant to be here?

stockflux-securities-master/src/components/reusableComponents/ToggleSwitch.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import "./ToggleSwitch.css";
+
+const TextField = ({ value, id, checked, onChange }) => {
+  return (
+    <div>
```
jrhenderson1988: If the label is removed then you will have a single component (`<input />`) to return so you won't need a `div`. As a tip, if you want to return elements in your components, it's possible to wrap them with `<React.Fragment>...</React.Fragment>`, or better use its shorthand syntax `<>...</>`

stockflux-securities-master/src/components/reusableComponents/TextField.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import "./TextField.css";
+
+const TextField = ({ value, id, onChange, readOnly }) => {
```
jrhenderson1988: It can be worth using the spread operator (`...`) to grab any other props that may be passed through and use the spread operator again to pass them as props to your `<input>`. It might also be worth adding support for passing through a custom `className` when your component is used, which would be appended to your existing `className`

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -4,6 +4,8 @@ import { Link } from "react-router-dom";
 import "./InputForm.css";
 import { getSecurity } from "../services/SecuritiesService";
 import ErrorMessage from "./ErrorMessage";
+import TextField from "./reusableComponents/TextField";
```
jrhenderson1988: You can just stick these components in the same place as the others. No need for another folder. The name `reusableComponents` doesn't sound quite right to me

stockflux-securities-master/src/components/reusableComponents/ToggleSwitch.js
MaxOvenden: Oh dear, such a bad mistake. Yes it should...

stockflux-securities-master/src/components/reusableComponents/ToggleSwitch.js
MaxOvenden: This label is empty because it's simply there for CSS styling purposes. It does it this way in a tutorial I found on how to make a toggle switch.

stockflux-securities-master/src/components/reusableComponents/ToggleSwitch.js
MaxOvenden: changed now

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: Okay, sure. I've moved them.

stockflux-securities-master/src/components/reusableComponents/ToggleSwitch.js
MaxOvenden: Okay I've changed the <div>s to be <> fragment tags.

stockflux-securities-master/src/components/reusableComponents/TextField.js
MaxOvenden: okay done this now

stockflux-securities-master/src/components/reusableComponents/ToggleSwitch.js
jrhenderson1988: Just spoke to you about this and had a quick demo so this is fine

stockflux-securities-master/src/components/TextField.js
```diff
@@ -0,0 +1,17 @@
...
+import "./TextField.css";
+
+const TextField = props => {
+  return (
+    <input
+      {...props}
+      className={
+        "input-form-input" +
+        (props.readOnly ? " read-only" : "") +
+        ` ${props.className}`
```
jrhenderson1988: Since you're using backticks here which allows for string interpolation, you can simplify the whole `className`:

    className={`input-form-input ${readOnly && "read-only"} ${className}`}

stockflux-securities-master/src/components/TextField.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import "./TextField.css";
+
+const TextField = props => {
```
jrhenderson1988: I think it would be better here to specify the arguments you want to receive in the parameters to the component and use the spread operator to grab the rest of the props:

    const TextField = ({ readOnly, className, ...props }) =>

stockflux-securities-master/src/components/ToggleSwitch.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import "./ToggleSwitch.css";
+
+const ToggleSwitch = props => {
```
jrhenderson1988: As discussed in person and as mentioned for the `TextField` component, I think it would be better to specify the values you want to receive in your arguments and use the spread operator to receive the rest:

    const ToggleSwitch = ({ id, onChange, className, ...props }) => {

This effectively extracts `id`, `className` and `onChange` from the `props` and leaves the rest of them intact in the `props` variable so when you spread `...props` into the component in your JSX, those props won't be included and you can manually add them in with your custom logic

stockflux-securities-master/src/components/TextField.js
MaxOvenden: I think that will result in an unwanted space at the end if the props are not defined. I'll do a version of that but fix that issue

stockflux-securities-master/src/components/ToggleSwitch.js
```diff
@@ -0,0 +1,18 @@
...
+  return (
+    <>
+      <input
+        {...props}
+        type="checkbox"
+        className={"input-form-toggle" + ` ${props.className}`}
```
jrhenderson1988: Similar to in the `TextField` component here, if you expect these properties in the component arguments directly, you can remove the `props.` prefix since they will exist as their own variables. You can also use the backtick, string interpolation feature to simply the `className` expression:

    className={`input-form-toggle ${className}`}

stockflux-securities-master/src/components/TextField.js
MaxOvenden: it looks like it doesn't like me using the `condition && value`  method in the className.  I've asked Matt if he knows why but he said he didn't know. 

` readOnly ? " read-only"  : ""`
does work though.

stockflux-securities-master/src/components/TextField.js
MaxOvenden: yep added these back in

stockflux-securities-master/src/components/ToggleSwitch.js
MaxOvenden: okay , understood. Made the change now

stockflux-securities-master/src/components/ToggleSwitch.js
MaxOvenden: Yep I've done this.

Worth noting:  I've moved the extra className from the input to the label as the input is hidden so extra css won'yt do anything to it. The label is the one which is actually displayed

stockflux-securities-master/src/components/TextField.js
jrhenderson1988: > I think that will result in an unwanted space at the end if the props are not defined. I'll do a version of that but fix that issue

It will, yes, but you have that anyway in your current version with ` ${props.className}`. If `className` is not explicitly defined when using your component, a trailing space will be left too. I'd be interested to have a look at why the `condition && value` approach isn't working

stockflux-securities-master/src/components/ToggleSwitch.js
jrhenderson1988: Cool. Yeah, that's the right approach since the input is hidden anyway.

MaxOvenden:
> Just a few minor things. As discussed as well, it would be worth using the React PropTypes library to define the types of the props accepted by each component and whether or not they're mandatory or optional.

I've addressed these changes and added React PropTypes to both components. However, note my comment about not being able to use the && method in the classNames.

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -102,40 +96,34 @@ const InputForm = ({ match }) => {
...
-            <input
-              className="input-form-toggle"
-              id="visible-toggle"
+            <ToggleSwitch
               value={visible}
+              id="visible-toggle"
               checked={visible}
-              type="checkbox"
               onChange={() => setVisible(!visible)}
+              className="testClass"
```
jrhenderson1988: Did you mean to remove this?

MaxOvenden: woops, yes I did

MaxOvenden: gone now

stockflux-securities-master/src/components/TextField.js
```diff
@@ -0,0 +1,28 @@
...
+      readOnly={readOnly}
+      className={`input-form-input${readOnly ? " read-only" : ""}${
+        className ? ` ${className}` : ""
+      }`}
+    />
+  );
+};
+
+TextField.propTypes = {
+  onChange: PropTypes.func.isRequired,
```
jrhenderson1988: I'm not sure that any of these should be required. I think it's totally valid to be able to omit an `onChange` function and an `id`. A `value` is only really necessary if an existing value is actually present. In most cases they will be provided but I don't think they should be made mandatory. It might actually be worth removing these 3 fields from the destructured arguments to the component and just allowing the `{...props}` spread to apply them on the input element. The prop types you've defined should take care of ensuring that the correct types are given when the component is used.

stockflux-securities-master/src/components/ToggleSwitch.js
```diff
@@ -0,0 +1,28 @@
...
+      />
+      <label className={`toggle-switch ${className}`} htmlFor={id} />
+    </>
+  );
+};
+
+ToggleSwitch.propTypes = {
+  onChange: PropTypes.func.isRequired,
+  id: PropTypes.string.isRequired,
+  value: PropTypes.bool.isRequired,
```
jrhenderson1988: Value shouldn't have to be a boolean type. Checkbox input fields are often given string values.

MaxOvenden: I've made id required as without it the component doesn't function. I've made value and onChange optional though

stockflux-securities-master/src/components/TextField.js
MaxOvenden: yep done

PR #1232, opened 2019-07-23
Added post service to create a security and display success or error message to the user

MaxOvenden:
On submitting the form, if successful the form will clear and a success message will appear below the form. Otherwise the errors will display in the same place. If there are multiple errors they will render in an un-ordered list.


MaxOvenden:
Successful input:
image

MaxOvenden:
Multiple errors in user input:

image

stockflux-securities-master/src/components/Message.js
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import "./Message.css";
+
+const Message = ({ message, type }) => {
```
jrhenderson1988: Good choice to rename this component and make it re-usable based on its type since an error message and success message component would be virtually identical

stockflux-securities-master/src/components/Message.css
```diff
@@ -0,0 +1,9 @@
+.securities-success-message {
```
jrhenderson1988: Maybe you could have a `.securities-message` class as the based and then allow for adding `.error` or `.success` as modifiers which only change the colour. Just like we did before.

stockflux-securities-master/src/components/SecuritiesTable.css
```diff
@@ -76,7 +76,7 @@
   border-left: 1px solid var(--clr-divider-light);
 }
 
-.securities-error-message,
+.securities-error-message-container,
```
jrhenderson1988: I would rename this to something like `message-container` since its purpose is not *just* to contain error messages.

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -48,7 +48,9 @@ const SecuritiesTable = () => {
           <Components.LargeSpinner />
         </div>
       ) : errorMessage ? (
-        <ErrorMessage errorMessage={errorMessage} />
+        <div className="securities-error-message-container">
```
jrhenderson1988: See above regarding renaming this container

stockflux-securities-master/src/components/SecuritiesTable.css
MaxOvenden: In the securities table it is currently only being used to contain error messages. I renamed it because there was a css conflict in the message component I forgot to remove.

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -16,6 +16,8 @@ const InputForm = ({ match }) => {
   const [inProgress, setInProgress] = useState(false);
   const [isLoading, setIsLoading] = useState(false);
   const [errorMessage, setErrorMessage] = useState(null);
+  const [sendingMessages, setSendingMessages] = useState([]);
```
jrhenderson1988: The naming of `sendingMessages` and `sendingSuccess` doesn't seem right to me. You're also using an awful lot of state for what is effectively a bunch of messages and a type (error/success). I think it would be worth refactoring this to be something like `messages` which would store an array of messages and `type` which would dictate the type that you pass to your `Message` component.

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -133,6 +150,18 @@ const InputForm = ({ match }) => {
           </div>
         </form>
       )}
+      <ul>
+        {sendingMessages.map(message => {
```
jrhenderson1988: Here you're creating a list of messages and creating a `Message` component instance for each one. Consider making your `Message` component accept an array of messages and have it create a *single* message list, with an `<li>` for each message. Then you can just pass the messages into the component and let the component deal with rendering it.

stockflux-securities-master/src/services/SecuritiesService.js
```diff
@@ -20,3 +20,28 @@ export async function getSecurity(securityId) {
...
+    `${options.customData.apiBaseUrl}/securities-v2`,
+    {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json"
+      },
+      body: JSON.stringify(security)
+    }
+  );
+  if (response.status === 201) {
```
jrhenderson1988: When checking status codes it can be worth checking the group they fall in. To illustrate, all `2xx` responses are success type messages, `3xx` are *redirects*/*see other*, `4xx` are client errors and `5xx` are server errors. I haven't used the fetch API extensively but it may have some kind of helper methods to help you determine the type of response it received.

stockflux-securities-master/src/services/SecuritiesService.js
```diff
@@ -20,3 +20,28 @@ export async function getSecurity(securityId) {
...
+    const responseJSON = await response.json();
```
jrhenderson1988: You could just `return await response.json();` here. No need to assign a variable and then return it.

stockflux-securities-master/src/services/SecuritiesService.js
```diff
@@ -20,3 +20,28 @@ export async function getSecurity(securityId) {
...
+    return responseJSON;
+  } else if (response.status === 400) {
+    const responseJSON = await response.json();
+    throw new Error(responseJSON.messages);
```
jrhenderson1988: You should be careful here. Usually, you would pass a string message into the `Error` object but here you're passing an array of messages. When it comes to handling the error in the `.catch(...)`, then you will need to check the type of the errors and treat them differently. If you take a look at how the current `StockFluxService` is implemented, things seem to be done a little differently. I would try to stick to the existing convention where possible.

stockflux-securities-master/src/components/Message.css
MaxOvenden: yep, good point, I've changed it now

stockflux-securities-master/src/components/Message.js
MaxOvenden: Cheers

stockflux-securities-master/src/services/SecuritiesService.js
MaxOvenden: I've changed this so that it checks if it lies between the upper and lower bands ie `>=200 && <=299 `

stockflux-securities-master/src/services/SecuritiesService.js
MaxOvenden: k, changed

stockflux-securities-master/src/services/SecuritiesService.js
MaxOvenden: I've moved this logic into the form and made the service return an object with a status code and a message. What do you think about this approach?

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: Okay all the messages are now contained in the same object with a type: success or error and with a message: [array of messages, ...]

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: Done now

stockflux-securities-master/src/components/SecuritiesTable.css
MaxOvenden: Changed the name now

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: Yep done

stockflux-securities-master/src/components/Alert.css
```diff
@@ -0,0 +1,11 @@
+.securities-message {
+  font-size: var(--font-xxxlarge);
+}
+
+.error {
```
jrhenderson1988: As it stands, any element containing the class `.error` or `.success`, regardless of where in your application will receive this styling. You should make it so that this style only applies based on the existence of the `.alert`/`.securities-alert` class as well:

    .alert.error { ... }

stockflux-securities-master/src/components/Alert.css
```diff
@@ -0,0 +1,11 @@
...
+  color: red;
```
jrhenderson1988: Did you mean to change this colour? It should probably refer to a variable. You probably don't want the browser default `red` colour.

stockflux-securities-master/src/components/Alert.css
```diff
@@ -0,0 +1,11 @@
...
+}
+
+.success {
+  color: greenyellow;
```
jrhenderson1988: Did you mean to change this colour? It should probably refer to a variable. You probably don't want the browser default `greenyellow` colour.

stockflux-securities-master/src/components/Alert.js
```diff
@@ -0,0 +1,26 @@
+import React, { useEffect, useState } from "react";
+import "./Alert.css";
+
+const Alert = ({ messages, type }) => {
+  const [messagesArray, setMessagesArray] = useState([]);
```
jrhenderson1988: You don't need to use state here at all. You have the `messages` item, which contains the message(s) that you can iterate over when rendering your component with JSX.

stockflux-securities-master/src/components/Alert.js
```diff
@@ -0,0 +1,26 @@
...
+      ? setMessagesArray([messages])
+      : setMessagesArray(messages);
+  }, []);
+
+  return (
+    <ul>
+      {messagesArray.map((message, index) => {
+        return (
+          <li key={index}>
+            <div className={`securities-message ${type}`}>{message}</div>
```
jrhenderson1988: You don't really need this `div` here either. It would suffice to put the message directly in the `li` which can have a class applied to it as well

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -15,7 +19,7 @@ const InputForm = ({ match }) => {
   const [enabled, setEnabled] = useState(false);
   const [inProgress, setInProgress] = useState(false);
   const [isLoading, setIsLoading] = useState(false);
-  const [errorMessage, setErrorMessage] = useState(null);
+  const [messages, setMessages] = useState({ type: "success", content: null });
```
jrhenderson1988: Maybe it would be better to have two items of state in the form:
- One that describes the current *state*/*context* of the form, or the *type* of messages it has (errors, success etc.) which could be some form of enum type (It's generally not good practice to rely on strings as identifiers, they're very error prone)
- One that stores an array of messages (which could be empty if there are none)

You can then use that state to decide to show messages if there are any and give it the *context* which describes the type of messages they are. You could even use this *state*/*context* item to determine if the form is in the *loading* state, or in the *progress* state. This would help you remove quite a bit of the state and simplify the rendering of the form.

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -133,6 +156,11 @@ const InputForm = ({ match }) => {
           </div>
         </form>
       )}
+      <ul>
```
jrhenderson1988: The `Alert` component is now a list, but you're using it here as if it still represents a single item. You'll end up rendering a list, within a list.

stockflux-securities-master/src/components/Alert.css
MaxOvenden: I didn't change this color but yes I see that it should probaly be a variable. As it stands there is no css variable defined for error text but I can create one.

stockflux-securities-master/src/components/Alert.js
MaxOvenden: You said earlier that you thought it should check the type so that the user can enter a single message as a string. If you don't think this is needed then I'll change it.

stockflux-securities-master/src/components/Alert.js
```diff
@@ -0,0 +1,26 @@
...
+  const [messagesArray, setMessagesArray] = useState([]);
+
+  useEffect(() => {
+    typeof messages === "string"
+      ? setMessagesArray([messages])
+      : setMessagesArray(messages);
+  }, []);
+
+  return (
+    <ul>
```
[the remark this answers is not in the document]
MaxOvenden: Yep added now

stockflux-securities-master/src/components/Alert.css
```diff
@@ -0,0 +1,11 @@
+.securities-message {
```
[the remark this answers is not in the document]
MaxOvenden: Yep, done now. Now that I've gotten rid of the surrounding div there's no need to call the <li> in the css as I can just reference the className.

stockflux-securities-master/src/components/Alert.css
MaxOvenden: Yep good point, now changed

stockflux-securities-master/src/components/Alert.js
MaxOvenden: Yep removed the div now

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: Yep, forgot to remove that. Gone now.

stockflux-securities-master/src/components/Alert.css
MaxOvenden: Okay I've added the css variables error-red and success-green to the root of the securities master idex.css. 

I tried adding them to the main file where the rest are defined but it doesn't seem to respond. Do you know why? I think it's got something to do with the build.

MaxOvenden: Okay I've rebuild the components folder and it now works. I've given the variables hexcodes for colors very similar to how they were before.

stockflux-securities-master/src/components/Alert.js
MaxOvenden: I've removed the state and the if statement now

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: That sounds very similar to how I had it before you suggested I change it!

Before I had _success: true or false_ and sendingMessages = [Messages]

I like the idea of re-using the state/context to show loading though. I'll give that a go and see if it works

MaxOvenden: I've changed the state to have one value for loading, sending, error and success. I've created an object to be used as an enum to make it safer than using strings.

stockflux-securities-master/src/components/Alert.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import "./Alert.css";
+
+const Alert = ({ messages, type, className }) => {
```
jrhenderson1988: Remember, we said this should be able to accept a string or an array so you should first check if it is an array and if it isn't then convert it to one...

    messages = Array.isArray(messages) ? messages : [messages];

or you could just as easily do it when you call `.map(...)`

    {(Array.isArray(messages) ? messages : [messages]).map((message, index) => {

stockflux-securities-master/src/components/Alert.js
```diff
@@ -0,0 +1,18 @@
...
+  return (
+    <ul className={`securities-alert-container ${className}`}>
+      {messages.map((message, index) => {
```
jrhenderson1988: Since this closure/arrow function simply returns a single parameter, you can omit the braces `{`/`}` and the `return` so it becomes:

    {messages.map((message, index) => (
      <li className={`securities-alert ${type}`} key={index}>
        {message}
      </li>
    )}

stockflux-securities-master/src/components/Alert.js
```diff
@@ -0,0 +1,18 @@
...
+        return (
+          <li className={`securities-alert ${type}`} key={index}>
+            {message}
+          </li>
+        );
+      })}
+    </ul>
+  );
+};
+
```
jrhenderson1988: If you want to enforce that the `messages` parameter is an array so you don't have to do that check, you can use `propTypes` to ensure that it is an array of strings. You can even use `PropTypes.oneOf(...)` to specify that it should be either a string or an array of strings.

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -13,9 +17,15 @@ const InputForm = ({ match }) => {
   const [symbol, setSymbol] = useState("");
   const [visible, setVisible] = useState(false);
   const [enabled, setEnabled] = useState(false);
-  const [inProgress, setInProgress] = useState(false);
-  const [isLoading, setIsLoading] = useState(false);
-  const [errorMessage, setErrorMessage] = useState(null);
+  const [contextOrState, setContextOrState] = useState(null);
```
jrhenderson1988: When I said context or state, I meant one or the other. You should change this state item and its associated usage to be one or the other. Maybe just go with `formState` to avoid any confusion.

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -27,47 +37,60 @@ const InputForm = ({ match }) => {
 
   useEffect(() => {
     if (match.params.securityId) {
-      setIsLoading(true);
+      setContextOrState(stateEnum.loading);
       getSecurity(match.params.securityId)
         .then(security => {
-          setIsLoading(false);
-          setErrorMessage(null);
+          setContextOrState(null);
```
jrhenderson1988: Don't forget to empty the messages here if there are any:

    setMessages([])

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -27,47 +37,60 @@ const InputForm = ({ match }) => {
...
+          setMessages(["Error, cannot get security"]);
         });
     }
   }, []);
 
   const submitForm = event => {
     event.preventDefault();
-    setInProgress(true);
-
+    setContextOrState(stateEnum.sending);
```
jrhenderson1988: Don't forget to clear the messages here too.

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -27,47 +37,60 @@ const InputForm = ({ match }) => {
...
+        setMessages(["Security was successfully created"]);
+        setContextOrState(stateEnum.success);
+        setSecurityState({
+          exchange: "",
+          symbol: "",
+          name: "",
+          visible: false,
+          enabled: false
+        });
+        setContextOrState(stateEnum.success);
```
jrhenderson1988: You've already called this once on line 66.

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -27,47 +37,60 @@ const InputForm = ({ match }) => {
...
+      })
+      .catch(err => {
+        if (err instanceof ValidationError) {
+          setMessages(err.messages);
+        } else {
+          setMessages(err.message);
```
jrhenderson1988: If you're expecting to pass an array of messages into your alert component, then you need to convert this string to an array. Otherwise, you will get a TypeError when you try to run `.map(...)` on a string.

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -23,7 +23,7 @@ const SecuritiesTable = () => {
       })
       .catch(() => {
         setIsLoading(false);
-        setErrorMessage(timeoutMessage);
+        setErrorMessage([timeoutMessage]);
```
jrhenderson1988: If you're setting an array of messages here then shouldn't it be called `errorMessages`?

stockflux-securities-master/src/components/Alert.js
MaxOvenden: My mistake, I thought in your last review you told me to remove this check. I'll add it back in but without the useEffect and state

stockflux-securities-master/src/components/Alert.js
MaxOvenden: Alright, I think I'll go with adding the check in like above as it makes it more versatile

MaxOvenden: I'll also add the prop types in as you suggested

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: So do you still want the same state object to be used for loading, sending, error and success?

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: As it's being assigned to either an error message or a success message later in the post security function .then() or .catch() does it need to be cleared beforehand? 

There's no chance of it not being reassigned so I don't think clearing it beforehand will do anything

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: Good point. Changed now.

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: Yes good spot. I guess because if it throws a normal error it will be in the form of as string. 

Changed now.

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: Yeah, I need to be more careful to spot these obvious mistakes myself before I commit. Changed now.

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: Okay sure

stockflux-securities-master/src/components/Alert.js
MaxOvenden: Yep done

stockflux-securities-master/src/components/Alert.js
MaxOvenden: done

stockflux-securities-master/src/components/Alert.js
MaxOvenden: done now

stockflux-securities-master/src/components/InputForm.js
jrhenderson1988: I just mean rename `contextOrState` to `context` or `state`. My suggestion of using `formState` was just to try and clarify what I mean with a more descriptive name. You don't need to change any logic, just change the name of the variable and its associated setter function.

stockflux-securities-master/src/components/InputForm.js
jrhenderson1988: At the point where the form has been submitted, it should be neither in a success nor error state and therefore shouldn't have any messages. Clearing the messages here will have the effect of removing the success/error messages from view until the submit/save completes. It's very minor but it goes to keeping the state more correct at each point in time.

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: Okay thanks for the clarification, I've now made the change.

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: Ah cool, I understand. So if the submitting takes a while there will be no message while it'a waiting. I've now made the change.

stockflux-securities-master/src/components/Alert.js
```diff
@@ -0,0 +1,25 @@
...
+            {message}
+          </li>
+        )
+      )}
+    </ul>
+  );
+};
+
+Alert.propTypes = {
+  messages: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
```
jrhenderson1988: Would be better to use this

    PropTypes.arrayOf(PropTypes.string)

Also, aren't messages and type both required?

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -4,12 +4,12 @@ import "./SecuritiesTable.css";
...
 import { getSecuritiesData } from "../services/SecuritiesService";
 import AddSecurityButton from "./AddSecurityButton";
-import ErrorMessage from "./ErrorMessage";
+import Alert from "./Alert";
 
 const SecuritiesTable = () => {
   const [securitiesData, setSecuritiesData] = useState([]);
   const [isLoading, setIsLoading] = useState(true);
-  const [errorMessage, setErrorMessage] = useState(null);
+  const [errorMessages, setErrorMessages] = useState([null]);
```
jrhenderson1988: Why is the default state an array of null? Shouldn't it just be an empty array?

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -19,11 +19,11 @@ const SecuritiesTable = () => {
       .then(securities => {
         setIsLoading(false);
         setSecuritiesData(securities);
-        setErrorMessage(null);
+        setErrorMessages(null);
```
jrhenderson1988: Set to an empty array, not null

stockflux-securities-master/src/components/Alert.js
MaxOvenden: Okay, yeah will use that then. When I make type required it logs a warning on render because I'm setting formState to be null at the start. It might be worth leaving off the required for this filed then?

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: Okay done

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: Okay done

stockflux-securities-master/src/components/Alert.js
jrhenderson1988: Aren't you checking that the form state is one of the required states before you try to render it? If you're only allowing form state to be error or success then maybe you should only ever render the alert if the form state matches one of those instead?

stockflux-securities-master/src/components/Alert.js
jrhenderson1988: To clarify, my last review said that you don't need state since the argument (or prop) is passed.

stockflux-securities-master/src/components/Alert.js
MaxOvenden: Yeah I wasn't checking that. I've added the check in and made the message proptype required now. 

There are no proptype warnings in the console any more

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -125,14 +149,20 @@ const InputForm = ({ match }) => {
...
+                "input-submit-button"
               }
             >
               <button>{match.params.securityId ? "Edit" : "Create"}</button>
             </div>
           </div>
         </form>
       )}
+      {!!messages &&
+        formState !== stateEnum.sending &&
```
jrhenderson1988: An alert should *only* be shown if the state is error, or success right?

    {messages.length > 0 && [stateEnum.success, stateEnum.error].indexOf(formState) > -1 && (
      <Alert type={formState} messages={messages} />
    )}

MaxOvenden: good point, removed this line now

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -159,7 +159,6 @@ const InputForm = ({ match }) => {
         </form>
       )}
       {!!messages &&
-        formState !== stateEnum.sending &&
```
jrhenderson1988: Sure, that works too.

PR #1237, opened 2019-07-25
Fixed bug: security table body not displaying

MaxOvenden:
In my last PR I was prompted to make this change which I didn't realise broke the security table making it not display the data in the table body. The table body only renders if error is _falsy_.and `[]` is not considered _falsy_ by javascript but `null` is.


MaxOvenden:
> I don't think this is the right fix for that. A set of error messages should be an array and you should check to see if it contains anything or not when using it. It would be better to check to ensure that `errorMessages` contains items when it comes to rendering it, e.g.
> 
> ```
> errorMessages.length > 0 ? 'Errors...' : 'No errors'
> ```
> 
> Try to use the methods based on the actual types of the variables you're dealing with instead of relying on JavaScript coercing values and using the "falsiness" or "truthiness" wherever possible.

Okay I see why you made the comment now. Yeah alright, I will make a note of this and bear it in mind for future work.

I've made the suggested change.

PR #1239, opened 2019-07-31
Make update service to send patch request from edit form

MaxOvenden:
The user can now edit a security using a put request from the edit form

### DIFF
--- a/stockflux-securities-master/src/components/InputForm.css
+++ b/stockflux-securities-master/src/components/InputForm.css
@@ -9,6 +9,7 @@
 .input-form-body {
   display: flex;
   flex-direction: column;
+  height: 300px;
 }
 
 .input-form-title {
@@ -16,6 +17,11 @@
   font-weight: var(--font-weight-bolder);
   margin-left: var(--margin-small);
   font-size: var(--font-xxxlarge);
+}
+
+.input-form-spinner-container {
+  display: flex;
+  height: 300px;
 }
 
 .input-checkbox-container,
--- a/stockflux-securities-master/src/components/InputForm.css
+++ b/stockflux-securities-master/src/components/InputForm.css
@@ -9,7 +9,7 @@
 .input-form-body {
   display: flex;
   flex-direction: column;
-  height: 300px;
+  height: var(--input-form-height);
 }
 
 .input-form-title {
@@ -21,7 +21,7 @@
 
 .input-form-spinner-container {
   display: flex;
-  height: 300px;
+  height: var(--input-form-height);
 }
 
 .input-checkbox-container,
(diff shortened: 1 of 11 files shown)


stockflux-securities-master/src/services/SecuritiesService.js
```diff
@@ -1,3 +1,10 @@
+export class ValidationError extends Error {
```
AugusteBa: Maybe it would be worth moving this into it's own file?

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -63,26 +64,42 @@ const InputForm = ({ match }) => {
...
-        setSecurityState({
-          exchange: "",
-          symbol: "",
-          name: "",
-          visible: false,
-          enabled: false
+    if (match.params.securityId) {
+      updateSecurity(match.params.securityId, securityObject)
+        .then(() => {
+          setMessages(["Security was successfully updated"]);
```
AugusteBa: image

I think a single bullet point with a "Security was successfully updated" message looks a bit odd. Could we perhaps check the length of messages and if the array just contains a single element, maybe we can just display it as a paragraph then? Then if there is more than 1 element, we can display it as a list. This is just a suggestion, feel free to push back if you disagree. Also, I think we could reduce the font-size (currently 20px) to something a bot smaller, e.g. 15px or 16 px. In my personal, subjective and biased opinion, the message text size should probably be smaller than the button or input labels. Also might be worth changing the cursor type on interactive elements (e.g. regular buttons, toggle buttons, edit and delete icons) to pointer: `cursor: pointer;`.

stockflux-securities-master/src/services/SecuritiesService.js
MaxOvenden: Yep, done now

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: I've made all these changes including adding the pointer cursor to the securities table.

By the way, if you want the alert messages to be coloured green or red for success or error you need to `run npm build` as I added css variables for that. It should look like this:

image

stockflux-securities-master/src/components/InputForm.css
```diff
@@ -9,6 +9,7 @@
 .input-form-body {
   display: flex;
   flex-direction: column;
+  height: 300px;
```
jrhenderson1988: Is there a reason for hard-coding the height of the form to 300px? Is there not a variable somewhere that defines this height?

MaxOvenden: There is not a variable that defines the height although I could add one

MaxOvenden: I've added a css variable for this height now

PR #1242, opened 2019-08-02
Make delete service to remove a security and add buttons on table and edit form

MaxOvenden:
Added service to delete a security. The bin icons on the table delete a security and a delete button on the edit form links to the table with the security deleted


stockflux-securities-master/src/App.js
```diff
@@ -8,11 +8,11 @@ import { Redirect, Route, Switch } from "react-router-dom";
 import "./App.css";
 
 const App = () => {
-
   return (
     <div>
       <Components.Titlebar />
       <Switch>
+        <Route exact path="/table/:securityId" component={SecuritiesTable} />
```
jrhenderson1988: Why is there a route for the table with a `securityId`?

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -114,7 +114,7 @@ const InputForm = ({ match }) => {
           <Components.LargeSpinner />
         </div>
       ) : (
-        <form className="input-form-body" onSubmit={submitForm}>
+        <div className="input-form-body">
```
jrhenderson1988: Why has the form submit handler been removed in favour of a click handler on the button? This stops us submitting the form by hitting return etc.

MaxOvenden: Oh okay, I didn't realise that. I removed it because I was adding a second button which I didn't want to submit a form but I can add the form back and just add `type="button"` to the second button

stockflux-securities-master/src/App.js
MaxOvenden: I have removed this now and used the history prop to route from the form to the table if the deletion is successful. If it's unsuccessful then it will stay on the form and display an error. 

As we discussed, there is no need for a message after redirecting so no prop is being passed down anymore

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -94,16 +99,21 @@ const InputForm = ({ match }) => {
...
+    deleteSecurity(match.params.securityId)
+      .then(() => {
+        history.push("/");
+      })
+      .catch(err => {
+        errorHandler(err);
+      });
+  };
+
   return (
```
AugusteBa: Again, would suggest taking some stuff out of this into their own jsx definitions, because this return is getting very long.

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -43,51 +79,65 @@ const SecuritiesTable = () => {
...
+            {securitiesData.length === 0 && tableState !== stateEnum.error ? (
+              <div className="no-securities-container">
+                <div className="no-securities-message">
+                  You have no securities to show
+                </div>
+                <AddSecurityButton size="large" />
+              </div>
+            ) : (
+              <Components.ScrollWrapperY>
+                {securitiesData.map((item, index) => (
```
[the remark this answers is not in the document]
MaxOvenden: I've moved the table body into it's own component now

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -2,39 +2,75 @@ import React, { useEffect, useState } from "react";
...
+        errorHandler(err);
       });
   }, []);
 
+  const onClickDelete = securityId => {
+    setTableState(stateEnum.deleting);
+    setMessages([]);
+    deleteSecurity(securityId)
+      .then(response => {
+        getSecuritiesData()
```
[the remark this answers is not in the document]
MaxOvenden: Done

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: We agreed yesterday that this will be addressed in a later refactoring PR. This PR has already included a lot of refactoring.

stockflux-securities-master/src/components/AddSecurityButton.js
```diff
@@ -1,13 +1,15 @@
 import React from "react";
 import { Link } from "react-router-dom";
-import "./AddSecurityButton.css";
+import Button from "./Button";
 
 const AddSecuritiesButton = ({ size }) => {
```
AugusteBa: Filename and component names do not match. I would suggest naming them both to `ButtonAddSecurity`.

AugusteBa: Also maybe it would be worth having folder named "buttons" with all 3 types of buttons you have.

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -2,39 +2,66 @@ import React, { useEffect, useState } from "react";
 import Components from "stockflux-components";
 import "./SecuritiesTable.css";
 import { Link } from "react-router-dom";
```
AugusteBa: Is this import used anywhere?

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -2,39 +2,66 @@ import React, { useEffect, useState } from "react";
...
+      .catch(err => {
+        setSecuritiesData([]);
+        errorHandler(err);
       });
+  };
+
+  useEffect(() => {
+    setTableState(tableEnum.loading);
+    getSecuritiesHandler(!!match.params.message ? [match.params.message] : []);
   }, []);
```
AugusteBa: ```React Hook useEffect has missing dependencies: 'getSecuritiesHandler' and 'match.params.message'. Either include them or remove the dependency array.```

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -2,39 +2,66 @@ import React, { useEffect, useState } from "react";
...
+      })
+      .catch(err => {
+        errorHandler(err);
+      });
+  };
+
   return (
     <div className="securities-container">
       <div className="securities-title-container">
         <div className="securities-title">Securities</div>
```
AugusteBa: Shouldn't this be a heading or a paragraph?

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -2,39 +2,66 @@ import React, { useEffect, useState } from "react";
...
+        getSecuritiesHandler([response.message]);
+      })
+      .catch(err => {
+        errorHandler(err);
+      });
+  };
+
   return (
     <div className="securities-container">
       <div className="securities-title-container">
```
AugusteBa: Securities title container not only contains the title but also the button. Is this intended?

stockflux-securities-master/src/components/Button.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import "./Button.css";
+import PropTypes from "prop-types";
+
+const Button = ({ text, size, className, ...props }) => {
+  return (
+    <div className={`button-container ${size} ${className}`}>
```
jrhenderson1988: Why are you putting a `div` around the button here? Why not just add the `className` prop onto the `button` JSX element directly?

stockflux-securities-master/src/components/TableBody.js
```diff
@@ -0,0 +1,59 @@
+import React from "react";
```
AugusteBa: Hmm... I guess I meant you would move it out into a jsx (not a separate component) like this:

```
const tableBody  () => [ return jsx here]
```

Let's see what @jrhenderson1988 thinks as it works both ways.

stockflux-securities-master/src/components/Button.css
```diff
@@ -0,0 +1,42 @@
+.button-container.extraSmall > button,
```
jrhenderson1988: You're mixing `camelCase` with `kebab-case` here. Can we stick to a single naming convention?

stockflux-securities-master/src/components/Button.css
```diff
@@ -0,0 +1,42 @@
...
+.button-container.small > button,
+.button-container.large > button {
```
jrhenderson1988: You've got a lot of repeated classes here in your CSS. I would recommend having a single base class with a base style and then adding style for modifiers to the class.

```css
.button {
    /*base style here*/
}

.button.small {
  /*All overrides for small here*/
}

.button.small > span {
  /*Any overrides for direct span children of small buttons here*/
}
```

stockflux-securities-master/src/components/Button.js
```diff
@@ -0,0 +1,18 @@
...
+      <button {...props}>{text}</button>
+    </div>
+  );
+};
+
+Button.propTypes = {
+  size: PropTypes.oneOf(["extraSmall", "small", "large"]).isRequired,
```
jrhenderson1988: See above about keeping naming conventions consistent. If you want to accept `extraSmall` as a prop here, then you should at least pass `extra-small` to the `className` prop for consistency.

stockflux-securities-master/src/components/ConfirmationButton.js
```diff
@@ -0,0 +1,52 @@
...
+          size={size}
+        />
+      ) : (
+        <>
+          <div className="confirmation-message-container">
+            {confirmationText}
+          </div>
+          <div className="yes-no-button-container">
+            <Button
+            {...props}
```
jrhenderson1988: This needs to be indented once more

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: No, was used in the table body. Has been removed now

stockflux-securities-master/src/components/AddSecurityButton.js
MaxOvenden: I've used your naming convention and it has the same name as the file now

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: You mean instead of a div?

stockflux-securities-master/src/enums.js
```diff
@@ -0,0 +1,13 @@
+export const tableEnum = {
```
jrhenderson1988: I think these two could be named a bit better. Maybe something like `TableState` and `InputFormState`. Putting type type `enum` as part of the name is generally not a good idea. Descriptive names generally read much better.

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: Is it good practice to wrap text in <h> or <p> tags normally then?

stockflux-securities-master/src/services/SecuritiesService.js
```diff
@@ -64,3 +68,19 @@ export async function updateSecurity(securityId, security) {
...
+  const fetchOptions = {
+    method: "DELETE"
+  };
+  const options = await getWindowOptions();
+  const response = await fetch(
+    `${options.customData.apiBaseUrl}/securities-v2/${securityId}`,
+    fetchOptions
+  );
+  if (response.ok) {
+    return { message: "Security Successfully Deleted" };
```
jrhenderson1988: I don't think you should return a message here yourself, especially considering that you don't for any of the other service messages. Maybe just return nothing, which would indicate success still but it wouldn't resolve with a value (I think).

stockflux-securities-master/src/components/ConfirmationButton.js
```diff
@@ -0,0 +1,52 @@
+import React, { useState } from "react";
+import "./ConfirmationButton.css";
+import Button from "./Button";
+
+const ConfirmationButton = ({
```
jrhenderson1988: I think the approach you're making here is valid but it's not what I was looking for. Right now, you're effectively allowing the developer to use the `ConfirmationButton` component in place of the `Button`. There are some issues in your approach however, particularly in the way that you've passed the `props` down into both the button component as well as the yes button, but not the no button. You've also duplicated styles quite a bit. If the yes and no buttons both have the same style, then you can just make them use the same class.

My idea however, was to make it so that you could wrap a `Button` in a `Confirmation` component which would effectively hi-jack the click handler of that button and add the confirmation functionality. Then clicking the "yes" button would execute the button's original functionality.

```javascript
<Confirmation>
  <Button onClick={() => doSomething()} text="Button text" />
</Confirmation>
```

It might be worth looking into this approach since it will give you a deeper knowledge of how you can use React and how it works.

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: Yeah that's so the button appears in the same flex-box as the title.

stockflux-securities-master/src/components/TableBody.js
MaxOvenden: Alright, done now

stockflux-securities-master/src/services/SecuritiesService.js
MaxOvenden: Okay sure, if I want to display a message to the user should I set it in the components where it is used? Luke recently changed this service to not return a message so if I want to display one it will need to be hard coded either in the service or the components.

stockflux-securities-master/src/enums.js
MaxOvenden: I see what you mean. However my reason for the name was that `TableState ` is too similar to `tableState` which I'm currently using for the components state. Would tableStateConstants be a better name? Is there a convention for capitalisation?

stockflux-securities-master/src/components/ConfirmationButton.js
MaxOvenden: K, I'll have a go at this

stockflux-securities-master/src/App.js
```diff
@@ -8,12 +8,12 @@ import { Redirect, Route, Switch } from "react-router-dom";
 import "./App.css";
 
 const App = () => {
-
   return (
     <div>
       <Components.Titlebar />
       <Switch>
         <Route exact path="/" component={SecuritiesTable} />
+        <Route exact path="/securities/:message" component={SecuritiesTable} />
```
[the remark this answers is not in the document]
MaxOvenden: Yep changed to this way now, I'm just waiting on your reply to my question about where to set the message that's being passed through as you don't want me setting it in the service

stockflux-securities-master/src/components/Button.js
MaxOvenden: Yep changed the name here now

stockflux-securities-master/src/components/Button.css
MaxOvenden: yep good point, changed now

stockflux-securities-master/src/components/Button.js
MaxOvenden: I was copying what I had done before but yeah, I've changed it to this way now

stockflux-securities-master/src/services/SecuritiesService.js
jrhenderson1988: It should only be the purpose of the service to call the API and return/resolve back what the API responded with. If the API has an empty response then the service should return/resolve with nothing. Displaying a message to the user based on that success or response should be the responsibility of the UI so I would say the components. There are occasions in non-restful APIs where messages can be returned though but in this case I think it should be up to the component.

MaxOvenden: Okay great, that's what I thought so it is now done this way. Thanks for clarifying

stockflux-securities-master/src/enums.js
jrhenderson1988: The intention with this "enum" (which isn't really an enum since they're not supported in JS) is to avoid mistakes and limit the set of values which are considered to be correct. In this instance we're almost using your enum as if it was a type (enums are an actual type in Java and I think TypeScript too which compiles to JS). With this in mind, I would treat the enum as if it was a type, or a class and in which case the convention would usually be to `StudlyCase`. Variables in JavaScript tend to be `camelCase`, so it gives quite a good way of distinguishing what is what.

I'd definitely go for `TableState` and `InputFormState` or similar. You should avoid trying to use the name of the type as part of the name of something. For example, if you were to create a class, you would never name it `SomethingClass` or if you created a variable called `state` that had a boolean type you wouldn't call `stateBoolean`. Names of things should be descriptive as possible but also quite concise.

In your case I would be inclined to rename your `tableState` variable (the variable, not the enum) to something like `state`, if that can be done. If not, then maybe something else. Naming things is renowned for being one of a software developers greatest challenges and it might take some time to get right but it's good to try to get into the mindset of why things are named how they are and pick up good habits early.

> There are only two hard things in Computer Science: cache invalidation and naming things.
> -- Phil Karlton

stockflux-securities-master/src/components/Button.css
MaxOvenden: Yep, gotten rid of this unnessessary stuff now

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: I've renamed this div to securities header container as we discussed

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: I've made the headings for the table and the form into `<h1>` tags and modified the styling

AugusteBa: > Is it good practice to wrap text in `<h>` or `<p>` tags normally then?

yes

stockflux-securities-master/src/enums.js
MaxOvenden: Done this now

stockflux-securities-master/src/components/ConfirmationButton.js
MaxOvenden: Alright this has now been done with the help of Jack.

stockflux-securities-master/src/components/ButtonAddSecurity.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import { Link } from "react-router-dom";
+import Button from "./Button";
+
+const ButtonAddSecurity = ({ size }) => {
```
jrhenderson1988: I think this would be better named `AddSecurityButton`. I'm not sure it's worth having this as a separate component either since it's just a simple button and used only once. There are other places where you've got a `<Link>` wrapping a `<Button>` as well.

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -94,22 +96,43 @@ const InputForm = ({ match }) => {
...
+        setState(InputFormState.success);
+        setMessages(["Security Successfully Deleted"]);
+        setRedirect(true);
+      })
+      .catch(err => {
+        errorHandler(err);
+      });
+  };
+
+  const renderRedirect = () => {
```
jrhenderson1988: Personally, I would remove this function altogether in favour of actually conditionally rendering the redirect. This function isn't *pure* since you're relying on a value (redirect) that exists in a scope outside of it. If this value changes at some point, then your function will return a different value. A better signature would be to do something like this:

    const renderRedirect = (redirect) => {

The redirect is then passed to the function and this makes it pure because it has no side effects and it will always act the same given the same input. I'd still remove it though in favour of conditionally rendering the redirect component where you need to.

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -167,21 +190,34 @@ const InputForm = ({ match }) => {
...
+                <Button
+                  type="button"
+                  className="input-delete-button"
+                  text="Delete"
+                  size="large"
+                  onClick={onClickDelete}
+                />
+              </Confirmation>
+            )}
+            {renderRedirect()}
```
jrhenderson1988: As mentioned above, I think you should conditionally render the Redirect component directly here. You can use the `redirect && <Redirect ... />` syntax to do it quite easily. To make it even more clear, perhaps add a conditional return quite near the top of your component function, before even bothering to return JSX:

    if (redirect) {
        return <Redirect ... />
    }

    // ...

stockflux-securities-master/src/enums.js
```diff
@@ -0,0 +1,13 @@
+export const TableState = {
```
jrhenderson1988: One more suggestion here, but this is quite minor. It tends to be a convention that when you have unchanging, constant variables, particularly with enums that you define the variables in capital letters. So you would have something like:

    export const TableState = {
        LOADING: "loading",
        DELETING: "deleting",
        ERROR: "error",
        SUCCESS: "success"
    }

I think this would look and read better. Same goes for the `InputFormState` below too.

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: I've chosen the && method and kept it near the button as I think it's nice to keep it close to the button that renders it. They  did this in a tutorial I read as well

stockflux-securities-master/src/enums.js
MaxOvenden: DONE

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: Yep, I've removed the function in favour of conditionally rendering

stockflux-securities-master/src/components/ButtonAddSecurity.js
MaxOvenden: removed this now

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -1,93 +1,149 @@
...
+  getSecuritiesData,
+  deleteSecurity
+} from "../services/SecuritiesService";
+import ValidationError from "../services/ValidationError";
 import Alert from "./Alert";
+import { TableState } from "../enums";
+import Button from "./Button";
 
-const SecuritiesTable = () => {
+const tableBody = (onClickDelete, securitiesData, state) => {
```
AugusteBa: I think maybe you should have this inside `SecuritiesTable` as it very tightly related to Securitites table. It would alos save you passing in these arguments `tableBody(onClickDelete, securitiesData, state)` and you could just do `{tableBody}` rather than `{tableBody(onClickDelete, securitiesData, state)}`.

Another reason for this would be readability. Normally the first thing you would expect to see is the definition of the object that filename refers to, in this case "SecuritiesTable", rather than `tableBody`.

MaxOvenden: I see, good point. I have now done this

PR #1243, opened 2019-08-09
Make patch service to toggle visibility and enabled and add buttons in grid

MaxOvenden:
The service for the patch request has been added and the visibility and enabled/disabled are now toggle-able from within the table.


stockflux-securities-master/src/components/ToolTip.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import "./ToolTip.css";
+import PropTypes from "prop-types";
+
+const ToolTip = ({ children, message, className }) => {
+  return (
+    <div className={`tool-tip-container ${className}`}>
```
AugusteBa: This adds `undefined` class to the class list as shown below:
image

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -78,17 +91,53 @@ const SecuritiesTable = ({ location }) => {
...
+                      className={`securities-table-button ${
+                        item.enabled ? "" : "greyed-out"
+                      }`}
+                      onClick={() =>
+                        patchSecurityHandler(item.securityId, {
+                          enabled: !item.enabled
+                        })
+                      }
+                    >
+                      <span className="material-icons">
```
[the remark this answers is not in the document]
jrhenderson1988: Yeah, I agree, we should be consistent across all apps

stockflux-securities-master/src/services/SecuritiesService.js
```diff
@@ -85,3 +85,24 @@ export async function deleteSecurity(securityId) {
...
+    },
+    body: JSON.stringify(updates)
+  };
+  const options = await getWindowOptions();
+  const response = await fetch(
+    `${options.customData.apiBaseUrl}/securities-v2/${securityId}`,
+    fetchOptions
+  );
+  if (response.ok) {
+    return;
```
jrhenderson1988: The PATCH request returns the body of the security in its response with a 200 status so you should return it here. It's only when there's a 204 No Content, like the DELETE endpoint returns that there won't be any data to return/resolve.

stockflux-securities-master/src/components/SecuritiesTable.js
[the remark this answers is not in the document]
MaxOvenden: Jack recommended using material icons for the back button a while back. I can look into react-icons if it's what the other apps use though

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -78,17 +91,53 @@ const SecuritiesTable = ({ location }) => {
...
+                  </ToolTip>
+                  <ToolTip message="Delete">
+                    <button
+                      className="securities-table-button"
+                      onClick={() => deleteSecurity(item.securityId)}
+                    >
+                      <span className="material-icons">delete</span>
+                    </button>
+                  </ToolTip>
+                  <ToolTip message="Visibility">
```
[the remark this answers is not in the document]
MaxOvenden: Yep done this now

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -78,17 +91,53 @@ const SecuritiesTable = ({ location }) => {
...
+                          visible: !item.visible
+                        })
+                      }
+                    >
+                      <span className="material-icons">
+                        {item.visible ? "visibility" : "visibility_off"}
+                      </span>
+                    </button>
+                  </ToolTip>
+                  <ToolTip message="Enabled?">
```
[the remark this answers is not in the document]
MaxOvenden: Made the change now.

stockflux-securities-master/src/components/SecuritiesTable.js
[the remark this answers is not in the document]
MaxOvenden: Added react icons to the table now

stockflux-securities-master/src/services/SecuritiesService.js
MaxOvenden: Okay cool, made the change now. Is there a reason why it needs to return the security when it doesn't get used in the component?

stockflux-securities-master/src/components/ToolTip.js
MaxOvenden: I've changed this to `<div className={`tool-tip-container ${className ? className : ""}`}>` which fixes this. I've also done this to the `Alert`, `Button` and `Confirmation` components which also have has optional `className` props.

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -137,7 +186,7 @@ const SecuritiesTable = ({ location }) => {
                 <Alert messages={messages} type={state} />
               </div>
             )}
-          {state === TableState.DELETING && (
+          {(state === TableState.DELETING || state === TableState.UPDATING) && (
```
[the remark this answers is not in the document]
MaxOvenden: Okay I've added reducers to the table and the form to hold the state now. Let me know what you think.

stockflux-securities-master/src/reducers/securitiesTableReducer.js
```diff
@@ -0,0 +1,32 @@
...
+
+export const initialTableState = {
+  fetchStatus: TableState.LOADING,
+  hasErrors: false,
+  messages: []
+};
+
+export function securitiesTableReducer(state, action) {
+  switch (action.type) {
+    case "loading":
```
AugusteBa: I think you can do this:
`case TableState.LOADING:`

stockflux-securities-master/src/reducers/securitiesTableReducer.js
```diff
@@ -0,0 +1,32 @@
...
+      return { ...state, fetchStatus: TableState.LOADING, messages: [] };
```
AugusteBa: I think maybe you should reset `hasErrors` here to false as we are loading some new data in, so old error messages are no longer relevant.

stockflux-securities-master/src/reducers/securitiesTableReducer.js
```diff
@@ -0,0 +1,32 @@
...
+    case "deleting":
```
AugusteBa: Same might apply here

stockflux-securities-master/src/reducers/securitiesTableReducer.js
MaxOvenden: Yep I've added this in now

stockflux-securities-master/src/reducers/securitiesTableReducer.js
MaxOvenden: yep good point, changed now

stockflux-securities-master/src/reducers/securitiesTableReducer.js
MaxOvenden: Yep done now

stockflux-securities-master/src/components/Alert.js
```diff
@@ -4,7 +4,7 @@ import PropTypes from "prop-types";
 
 const Alert = ({ messages, type, className }) => {
   return (
-    <div className={`securities-alert-container ${className}`}>
+    <div className={`securities-alert-container ${className ? className : ""}`}>
```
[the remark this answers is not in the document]
AugusteBa: Also, I think in most places where you're using `PropTypes.oneOf([...])`, this should refer to enums. E.g. We should have enum called `ALERT_TYPE` with `SUCCESS` and `ERROR` in it.

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -10,19 +10,25 @@ import ToggleSwitch from "./ToggleSwitch";
...
   const [exchange, setExchange] = useState("");
   const [symbol, setSymbol] = useState("");
   const [visible, setVisible] = useState(false);
   const [enabled, setEnabled] = useState(false);
-  const [state, setState] = useState(
-    match.params.securityId ? InputFormState.LOADING : null
-  );
-  const [messages, setMessages] = useState([]);
+
   const [redirect, setRedirect] = useState(false);
```
AugusteBa: I think this would be worth changing to a single state object in your big refactoring PR.

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -1,24 +1,28 @@
...
 const SecuritiesTable = ({ location }) => {
+  
   const [securitiesData, setSecuritiesData] = useState([]);
-  const [messages, setMessages] = useState([]);
-  const [state, setState] = useState(TableState.LOADING);
+  const [state, dispatch] = useReducer(securitiesTableReducer, initialTableState);
 
   const handleError = err => {
-    setState(TableState.ERROR);
+    dispatch({type: "error"});
```
AugusteBa: Is it necessary t have the first dispatch `dispatch({type: "error"});` when error will be dispatched again from if/else below? Maybe this is redundant?

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -78,17 +92,57 @@ const SecuritiesTable = ({ location }) => {
                 <div className="securities-table-cell">{item.symbol}</div>
                 <div className="securities-table-cell">{item.name}</div>
                 <div className="securities-table-cell">
-                  <Link to={`/inputform/${item.securityId}`}>
-                    <button className="securities-table-button">
-                      <span className="material-icons">edit</span>
+                  <ToolTip message="Edit">
```
AugusteBa: Again, this might be a thing to add to your big refactoring PR, splitting out this huge return into smaller JSXs.

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -120,24 +174,24 @@ const SecuritiesTable = ({ location }) => {
...
                 className={`securities-message-container ${
                   securitiesData.length === 0 ? "no-securities" : ""
                 }`}
               >
-                <Alert messages={messages} type={state} />
+                <Alert messages={state.messages} type={state.hasErrors ? "error" : "success"} />
               </div>
             )}
-          {state === TableState.DELETING && (
+          {(state.fetchStatus === TableState.DELETING || state.fetchStatus === TableState.UPDATING) && (
```
AugusteBa: Also this file is missing PropTypes.

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -209,11 +222,12 @@ const InputForm = ({ match }) => {
...
-        )}
+      {!!state.messages && !state.fetchStatus && (
+        <Alert
+          type={state.hasErrors ? "error" : "success"}
+          messages={state.messages}
+        />
+      )}
       <Link to="/">
         <div className="back-button">
           <button>
```
AugusteBa: This file is missing PropTypes as well if we want to make it consistent.

stockflux-securities-master/src/components/ToolTip.css
```diff
@@ -0,0 +1,35 @@
...
+  top: 0px;
+  right: 105%;
+}
+
+.tool-tip-container .tool-tip-message:after {
+  content: " ";
+  position: absolute;
+  top: 50%;
+  left: 100%;
+  margin-top: -5px;
```
AugusteBa: Is this negative margin intended?

stockflux-securities-master/src/components/ToolTip.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import "./ToolTip.css";
+import PropTypes from "prop-types";
+
+const ToolTip = ({ children, message, className }) => {
```
AugusteBa: As far as I can see the only props that Tooltip can take are `children` and `message`, aka no `className`, so I think it can be removed.

stockflux-securities-master/src/components/ToolTip.js
```diff
@@ -0,0 +1,19 @@
...
+  return (
+    <div className={`tool-tip-container ${className ? className : ""}`}>
```
AugusteBa: Redundant `className` check.

stockflux-securities-master/src/components/ToolTip.js
```diff
@@ -0,0 +1,19 @@
...
+      {children}
+      <span className="tool-tip-message">{message}</span>
+    </div>
+  );
+};
+
+ToolTip.propTypes = {
+  message: PropTypes.string,
+  className: PropTypes.string
```
AugusteBa: `className` should also be removed from here if removed above.

stockflux-securities-master/src/enums.js
```diff
@@ -1,6 +1,7 @@
 export const TableState = {
   LOADING: "loading",
   DELETING: "deleting",
+  UPDATING: "updating",
```
AugusteBa: How's updating different from loading? If they are doing the same thing, maybe we should just keep one, remove redundant one and rename it so it would accommodate both usage scenarios?

stockflux-securities-master/src/reducers/securitiesTableReducer.js
```diff
@@ -0,0 +1,32 @@
...
+    case TableState.LOADING:
+      return { ...state, fetchStatus: TableState.LOADING, messages: [] };
+    case TableState.DELETING:
+      return { ...state, fetchStatus: TableState.DELETING, messages: [] };
+    case TableState.UPDATING:
+      return { ...state, fetchStatus: TableState.UPDATING, messages: [] };
+    case TableState.ERROR:
+      return {
+        ...state,
+        fetchStatus: null,
```
AugusteBa: I don't think we should set `fetchStatus` to null, especially if we are using enum elsewhere. Maybe enum should be changed to incorporate a case for this (not not null).

stockflux-securities-master/src/reducers/inputFormReducer.js
```diff
@@ -0,0 +1,24 @@
...
+export function inputFormReducer(state, action) {
+  switch (action.type) {
+    case InputFormState.LOADING:
+      return { ...state, fetchStatus: InputFormState.LOADING, messages: [] };
+    case InputFormState.LOADING:
+      return { ...state, fetchStatus: InputFormState.SENDING, messages: [] };
+    case InputFormState.ERROR:
+      return {
+        ...state,
+        fetchStatus: null,
```
AugusteBa: See comment below regarding setting `fetchStatus` to null. Note, that it is set to null in the case below as well.

stockflux-securities-master/src/reducers/inputFormReducer.js
```diff
@@ -0,0 +1,24 @@
...
+        hasErrors: true,
+        messages: action.messages
+      };
+    case InputFormState.SUCCESS:
+      return {
+        ...state,
+        fetchStatus: null,
+        hasErrors: false,
+        messages: action.messages
+      };
```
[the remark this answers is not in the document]
MaxOvenden: Added a default case just before you left this review!

stockflux-securities-master/src/reducers/securitiesTableReducer.js
```diff
@@ -0,0 +1,32 @@
...
+        hasErrors: true,
+        messages: action.messages
+      };
+    case TableState.SUCCESS:
```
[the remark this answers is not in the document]
MaxOvenden: Same as above

stockflux-securities-master/src/reducers/inputFormReducer.js
```diff
@@ -0,0 +1,24 @@
+import { InputFormState } from "../enums";
+
+export function inputFormReducer(state, action) {
+  switch (action.type) {
+    case InputFormState.LOADING:
+      return { ...state, fetchStatus: InputFormState.LOADING, messages: [] };
+    case InputFormState.LOADING:
```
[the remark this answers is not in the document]
MaxOvenden: Yep fixed this as well

stockflux-securities-master/src/enums.js
MaxOvenden: Updating and Deleting do similar thing but display different messages.

stockflux-securities-master/src/components/ToolTip.js
MaxOvenden: Yep done

stockflux-securities-master/src/components/ToolTip.js
MaxOvenden: As className isn't being used in this component I have removed it as a prop after our discussion today

stockflux-securities-master/src/components/Button.js
```diff
@@ -4,7 +4,10 @@ import PropTypes from "prop-types";
 
 const Button = ({ text, size, className, ...props }) => {
   return (
-    <button className={`button-container ${size} ${className}`} {...props}>
+    <button
+      className={`button-container ${size} ${className ? className : ""}`}
```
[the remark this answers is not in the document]
MaxOvenden: Yeah you're right, my reasoning was to make it more reusable but from our discussion today I get that it isn't needed here so have removed it.

stockflux-securities-master/src/components/Alert.js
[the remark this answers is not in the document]
MaxOvenden: I've changed this to work the way Jonathon found with the classes on the button component to allow `PropTypes.instanceOf(class)`.

stockflux-securities-master/src/reducers/securitiesTableReducer.js
MaxOvenden: My reasoning was that when the error or success has occured, the table has stopped loading or deleting. If after loading the fetch status doesn't get set to null it remains loading and the loading spinner remains shown which is not what you want

MaxOvenden: Ah I see, so instead of null you think I should call it `"completed"` or `"Inactive"` ?

stockflux-securities-master/src/reducers/inputFormReducer.js
MaxOvenden: I have addressed this in both reducers by setting it to an enum `TableState.COMPLETED` and `InputFormState.COMPLETED`

stockflux-securities-master/src/reducers/securitiesTableReducer.js
MaxOvenden: I've added this enum now

stockflux-securities-master/src/enums.js
MaxOvenden: Loading produces a large spinner in place of the securities, deleting and updating produce a spinner underneath the securities table in place of the alerts. It's a bit hard to find a name that incorporates both updating and deleting though because they are pretty different concepts. Maybe modifying?

stockflux-securities-master/src/components/Button.js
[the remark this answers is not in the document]
MaxOvenden: Yep removed the check

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -188,11 +202,12 @@ const InputForm = ({ match }) => {
           </div>
           <div className="input-buttons-container">
             <Button
-              size="large"
+              size={ButtonSize.LARGE}
               text={match.params.securityId ? "Save" : "Create"}
               className={
-                (state === InputFormState.SENDING ? "in-progress " : "") +
-                "input-submit-button"
+                (state.fetchStatus === InputFormState.SENDING
```
AugusteBa: Maybe this would be neater: 
```
className={`input-submit-button${state.fetchStatus === InputFormState.SENDING && ' in-progress'}`}
```
?

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -1,24 +1,33 @@
...
-  const [messages, setMessages] = useState([]);
-  const [state, setState] = useState(TableState.LOADING);
+  const [state, dispatch] = useReducer(
+    securitiesTableReducer,
+    initialTableState
+  );
 
   const handleError = err => {
-    setState(TableState.ERROR);
+    dispatch({ type: "error" });
```
AugusteBa: I think I've commented on this already but this is redundant. You are already going to dispatch error in either if or else below.

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -78,17 +97,57 @@ const SecuritiesTable = ({ location }) => {
...
+                      className="securities-table-button"
+                      onClick={() => deleteSecurity(item.securityId)}
+                    >
+                      <FaTrashAlt size={20} />
+                    </button>
+                  </ToolTip>
+                  <ToolTip message={item.visible ? "Hide" : "Show"}>
+                    <button
+                      className={`securities-table-button ${
+                        item.visible ? "" : "greyed-out"
```
AugusteBa: Again, try and avoid ternary operations, when one the options is an empty string: "". You can just use && instead.

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -78,17 +97,57 @@ const SecuritiesTable = ({ location }) => {
...
-                    className="securities-table-button"
-                    onClick={() => deleteSecurity(item.securityId)}
-                  >
-                    <span className="material-icons">delete</span>
-                  </button>
+                  </ToolTip>
+                  <ToolTip message={item.enabled ? "Disable" : "Enable"}>
+                    <button
+                      className={`securities-table-button ${
+                        item.enabled ? "" : "greyed-out"
```
AugusteBa: Same here

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -120,24 +179,26 @@ const SecuritiesTable = ({ location }) => {
...
-                }`}
-              >
-                <Alert messages={messages} type={state} />
-              </div>
-            )}
-          {state === TableState.DELETING && (
+          {state.messages.length > 0 && (
+            <div
+              className={`securities-message-container ${
+                securitiesData.length === 0 ? "no-securities" : ""
```
AugusteBa: Same here again

stockflux-securities-master/src/components/ToolTip.css
AugusteBa: Since you requested my review again, I am asking again :D Is this negative margin intended? We should avoid negative margins/paddings/positions etc. if possible.

stockflux-securities-master/src/enums.js
AugusteBa: Yeah, or maybe just keep it as `UPDATING`, as deleting is also updating in a way?

AugusteBa:
InputForm.js, SecuritiesTable.js is still missing proptypes.

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: k, changed this

stockflux-securities-master/src/components/ToolTip.css
MaxOvenden: Woops, I missed that comment but yes this negative margin is needed to keep the tooltip triangle arrow thing centred relative to the tool tip message.  I've tried to do it using positive bottom instead of negative top but this is the only way that seems to give the desired appearance. I know it is a bit hacky

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: Good spot, I don't think you mentioned that earlier but have now removed it

> InputForm.js, SecuritiesTable.js are still missing proptypes.

I've added propTypes to these components now

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: okay changed now

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: yep changed

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: yep changed now

stockflux-securities-master/src/components/Alert.js
```diff
@@ -2,31 +2,39 @@ import React from "react";
...
 
+AlertType.ERROR = new AlertType("error");
+AlertType.SUCCESS = new AlertType("success");
+
+const Alert = ({ messages, type }) => (
+  <div className="securities-alert-container">
+    {Array.isArray(messages) && messages.length > 1 ? (
+      <ul>
+        {messages.map((message, index) => (
+          <li className={`securities-alert ${type.valueOf()}`} key={index}>
```
jrhenderson1988: I think the base element of the alert should be the one to house the base CSS classes. In this case, the `ul` should be the element that gets this `className` and your CSS should be changed to reflect. If you want to style elements that fall within the base component, then you can use CSS to target elements within them either by class, or by element type, e.g.

    .securities-alert {
      font-size: var(--font-xlarge);
    }
    .securities-alert li {
      margin: var(--margin-small);
    }
    .securities-alert.error {
      color: var(--error-red);
    }

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -188,32 +203,31 @@ const InputForm = ({ match }) => {
...
             )}
           </div>
         </form>
       )}
-      {!!messages &&
-        (state === InputFormState.ERROR ||
-          state === InputFormState.SUCCESS) && (
-          <Alert type={state} messages={messages} />
-        )}
+      {!!state.messages && (
```
jrhenderson1988: Assuming `state.messages` is an array here, `!!state.messages` will be `true` when the array is empty here, resulting in rendering the alert even when there aren't any messages to show. Remember our discussion about trying to be more explicit and careful with types and not relying on JavaScript coercing them? I think you should change this to:

    {state.messages.length > 0 && (

stockflux-securities-master/src/components/SecuritiesTable.css
```diff
@@ -51,7 +51,7 @@
 
 .table-deleting-spinner-container,
 .securities-message-container {
-  margin-top: var(--margin-large);
+  margin: var(--margin-small) 0px;
```
jrhenderson1988: Just a tip: You can omit the units (`px` in this case) if you want to set the size of something to 0

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -1,24 +1,33 @@
...
 import "./SecuritiesTable.css";
 import * as service from "../services/SecuritiesService";
 import ValidationError from "../services/ValidationError";
-import Alert from "./Alert";
+import Alert, { AlertType } from "./Alert";
 import { TableState } from "../enums";
-import Button from "./Button";
+import ToolTip from "./ToolTip";
+import Button, { ButtonSize } from "./Button";
+import { FaPen, FaTrashAlt, FaEye, FaEyeSlash } from "react-icons/fa";
```
jrhenderson1988: We should stick to a single icon theme, otherwise we risk things looking noticeably different. We use Font Awesome in the other apps so we should stick to those. Font Awesome calls the cross you're looking for "times" and the tick/check is called "check".

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -1,24 +1,33 @@
...
+  const [state, dispatch] = useReducer(
+    securitiesTableReducer,
+    initialTableState
+  );
 
   const handleError = err => {
-    setState(TableState.ERROR);
     if (err instanceof ValidationError) {
-      setMessages(err.messages);
+      dispatch({ type: "error", messages: err.messages });
```
jrhenderson1988: A bit of a nitpick but I would be inclined to change this block to use a ternary, which could remove a good deal of duplication:

    const handleError = err =>
      dispatch({
        type: "error",
        messages: err instanceof ValidationError ? err.messages : [err.message]
      });

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -1,24 +1,33 @@
...
     } else {
-      setMessages([err.message]);
+      dispatch({ type: "error", messages: [err.message] });
```
jrhenderson1988: Also, type here being a string doesn't seem right. An enum style constant would be better. You're using `TableState.LOADING` etc. further down so I would imagine that this just slipped through the cracks.

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -54,18 +61,30 @@ const SecuritiesTable = ({ location }) => {
       });
   };
 
+  const patchSecurityHandler = (securityId, updates) => {
```
jrhenderson1988: I think the name of this function should be changed. Functions/methods that do things should be named almost like commands, "do this", "run that", "update this" rather than being something like "something handler". In this case I would  name the function something like `updateSecurity` or `patchSecurity` (which doesn't conflict with your service method since you call it on an object) to follow your existing naming convention.

stockflux-securities-master/src/components/TextField.js
```diff
@@ -2,22 +2,22 @@ import React from "react";
...
+      className={`input-form-input ${readOnly && "read-only"}`}
     />
   );
 };
 
 TextField.propTypes = {
-  onChange: PropTypes.func,
   readOnly: PropTypes.bool,
-  className: PropTypes.string
+  value: PropTypes.string.isRequired,
```
jrhenderson1988: I agree that value should be a string, but should it be required? What happens in the case of an add form where there isn't a value? What about in the event that you re-use this component in another place in the future where there's no default value, like a search box or something? I think this should be optional.

stockflux-securities-master/src/enums.js
```diff
@@ -1,13 +1,15 @@
 export const TableState = {
   LOADING: "loading",
-  DELETING: "deleting",
+  UPDATING: "updating",
```
jrhenderson1988: I thought these enums had changed to be functional classes to allow you to use the `instanceOf` prop type and operator? Right now, the `instanceOf` prop type won't work.

stockflux-securities-master/src/components/Alert.js
```diff
@@ -2,31 +2,39 @@ import React from "react";
...
+
+AlertType.prototype.valueOf = function() {
+  return this._value;
 };
 
+AlertType.ERROR = new AlertType("error");
+AlertType.SUCCESS = new AlertType("success");
+
+const Alert = ({ messages, type }) => (
+  <div className="securities-alert-container">
```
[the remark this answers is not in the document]
MaxOvenden: Okay that makes sense. There's already a securities-message-container which does the same thing anyway so it's definitely not needed

stockflux-securities-master/src/components/Alert.js
MaxOvenden: Okay, if you think it'a better that way. Because the styles are being applied to the list items or the single div I'll need to add a separate className for each :
`.securities-alert-list > li,
.securities-alert {
  font-size: var(--font-xlarge);
  margin: var(--margin-small);
}

.securities-alert-list.error > li,
.securities-alert.error {
  color: var(--error-red);
}

.securities-alert-list.success > li,
.securities-alert.success {
  color: var(--success-green);
}
`
`{Array.isArray(messages) && messages.length > 1 ? (
      <ul className={`securities-alert-list ${type.valueOf()}`}>
        {messages.map((message, index) => (
          <li key={index}>{message}</li>
        ))}
      </ul>
    ) : (
      <div className={`securities-alert ${type.valueOf()}`}>{messages}</div>
    )}`

stockflux-securities-master/src/components/InputForm.js
MaxOvenden: Ah I guess you're right. So it hasn't been a problem so far because it's just been rendering an empty string but still best to avoid rendering when it's not needed

stockflux-securities-master/src/components/SecuritiesTable.css
MaxOvenden: okay

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: Yep changed this

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: Yeah this just slipped through the cracks. Is changed now

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: Okay I've renamed it to patch security

stockflux-securities-master/src/components/TextField.js
MaxOvenden: Okay, I've made this optional now

stockflux-securities-master/src/enums.js
MaxOvenden: The functional classes were only for the button sizes ("extra-small", "small", "large") and for alert type ("error", "success"). The instanceOf propType still works in these components and these enums are being used in the same way as before (they are not used as props, just as enums for the state)

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: changed now

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -78,17 +102,55 @@ const SecuritiesTable = ({ location }) => {
...
-                    className="securities-table-button"
-                    onClick={() => deleteSecurity(item.securityId)}
-                  >
-                    <span className="material-icons">delete</span>
-                  </button>
+                  </ToolTip>
+                  <ToolTip message={item.visible ? "Hide" : "Show"}>
+                    <button
+                      className={`securities-table-button ${!item.visible &&
+                        "greyed-out"}`}
```
joelwilliamsmith: `greyed-out` is an odd class name - can we think of an intention of the class name to come up with a better one! `disabled` maybe?

stockflux-securities-master/src/reducers/securitiesTableReducer.js
```diff
@@ -0,0 +1,42 @@
...
+
+export const initialTableState = {
+  fetchStatus: TableState.LOADING,
+  hasErrors: false,
+  messages: []
+};
+
+export function securitiesTableReducer(state, action) {
+  switch (action.type) {
+    case TableState.LOADING:
```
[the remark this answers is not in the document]
MaxOvenden: Okay I've added an actions folder to hold actions for these with for specific names

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -78,17 +102,55 @@ const SecuritiesTable = ({ location }) => {
...
+                      {item.visible ? (
+                        <FaEye size={20} />
+                      ) : (
+                        <FaEyeSlash size={20} />
+                      )}
+                    </button>
+                  </ToolTip>
+                  <ToolTip message={item.enabled ? "Disable" : "Enable"}>
+                    <button
+                      className={`securities-table-button ${!item.enabled &&
```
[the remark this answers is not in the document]
MaxOvenden: Okay, I've now added the classNames function to each place where instances of this occur in the code.

stockflux-securities-master/src/components/SecuritiesTable.js
```diff
@@ -54,18 +66,30 @@ const SecuritiesTable = ({ location }) => {
...
   };
 
+  const patchSecurity = (securityId, updates) => {
+    dispatch({ type: TableState.UPDATING });
+    service
+      .patchSecurity(securityId, updates)
+      .then(() => {
+        loadSecurities(["Security Updated"]);
+      })
+      .catch(err => {
```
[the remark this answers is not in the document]
MaxOvenden: Okay that makes sense. I've made this change now.

stockflux-securities-master/src/reducers/securitiesTableReducer.js
```diff
@@ -0,0 +1,42 @@
...
+      return {
+        ...state,
```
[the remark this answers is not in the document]
MaxOvenden: To be consistent with the input form where I'm just using the reducer as state, I've migrated the securitiesData into this state making the `...state` necessary

stockflux-securities-master/src/components/InputForm.js
```diff
@@ -1,28 +1,37 @@
...
 import { InputFormState } from "../enums";
+import { inputFormReducer } from "../reducers/inputFormReducer";
+import PropTypes from "prop-types";
 
 const InputForm = ({ match }) => {
   const [name, setName] = useState("");
   const [exchange, setExchange] = useState("");
   const [symbol, setSymbol] = useState("");
   const [visible, setVisible] = useState(false);
   const [enabled, setEnabled] = useState(false);
```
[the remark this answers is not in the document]
MaxOvenden: I have moved all the form state into the reducer now for the form and to keep it consistent I've done this for the form as well

stockflux-securities-master/src/components/Alert.js
```diff
@@ -2,31 +2,37 @@ import React from "react";
...
+);
+
 Alert.propTypes = {
   messages: PropTypes.oneOfType([
     PropTypes.string,
     PropTypes.arrayOf(PropTypes.string)
   ]).isRequired,
-  type: PropTypes.oneOf(["error", "success"]).isRequired,
-  className: PropTypes.string
+  type: PropTypes.instanceOf(AlertType).isRequired
```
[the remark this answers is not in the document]
MaxOvenden: Done this now

stockflux-securities-master/src/components/Button.js
```diff
@@ -2,16 +2,30 @@ import React from "react";
...
+  >
+    {text}
+  </button>
+);
+
 Button.propTypes = {
-  size: PropTypes.oneOf(["extra-small", "small", "large"]).isRequired,
+  size: PropTypes.instanceOf(ButtonSize).isRequired,
   className: PropTypes.string
 };
```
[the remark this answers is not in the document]
MaxOvenden: Good point. I was wondering about the prop `text` as trying to make it as reusable as possible I didn't want to restrict it to just a string.

stockflux-securities-master/src/components/SecuritiesTable.js
MaxOvenden: Well even though the icon represents the security being disabled I wanted a className that could be used for both `disabled` and `visible`. Plus `disabled` is often used to describe buttons are inactive or cannot be clicked whereas these buttons do work. 

What do you think?

MaxOvenden: I've changed the className from `greyed-out` to `feature-off`

stockflux-securities-master/src/components/Button.js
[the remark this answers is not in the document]
MaxOvenden: I've made the button a wrapper component so it can contain anything as it's child

PR #1244, opened 2019-08-22
Refactored Securities Table and Form UIs

MaxOvenden:
The separate components have been put into different folders to keep the file structure easier to follow. 

The components `TableBody`, `TableRow`, `FormBody` and `FormRow` have been created to make `SecuritiesTable` and `InputForm` more manageable sizes. 

All in-line `onClick` callback functions have been moved into `clickHandler` functions.


stockflux-securities-master/src/components/input-form/FormBody.css
```diff
@@ -0,0 +1,22 @@
+.input-form-body {
+  display: flex;
+  flex-direction: column;
+}
+
+.input-buttons-container {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  margin: var(--margin-medium) 0 var(--margin-large) 0;
```
AugusteBa: I think you can simplify this to:
`margin: var(--margin-medium) 0;`

stockflux-securities-master/src/components/securities-table/SecuritiesTable.js
```diff
@@ -0,0 +1,133 @@
+import React, { useEffect, useCallback, useReducer } from "react";
+import Components from "stockflux-components";
+import { Link } from "react-router-dom";
+import "./SecuritiesTable.css";
+import * as service from "../../services/SecuritiesService";
+import ValidationError from "../../services/ValidationError";
+import Alert, { AlertType } from "../alert/Alert";
+import { TableState } from "../../enums";
```
AugusteBa: I think can combine TableState and InputFormState into one, because they are almost identical.

stockflux-securities-master/src/components/securities-table/SecuritiesTable.js
```diff
@@ -0,0 +1,133 @@
...
+              </Button>
+            </Link>
+          )}
+        </div>
+      </div>
+      <div className="heading-container">
+        <h2 className="securities-table-heading">Exchange</h2>
+        <h2 className="securities-table-heading">Symbol</h2>
+        <h2 className="securities-table-heading">Name</h2>
+        <h2 className="securities-table-heading">Edit / Delete</h2>
```
AugusteBa: You could also use map here, e.g. something similar to this:
```
['Exchange', 'Symbol', 'Name', 'Edit / Delete'].map(heading => (<h2 className="securities-table-heading">{heading}</h2>)
```

It might save a couple of lines and reduce repetition. It's up to you though :)

stockflux-securities-master/src/components/securities-table/TableRow.js
```diff
@@ -0,0 +1,82 @@
...
+      enabled: !item.enabled
+    });
+  };
+
+  return (
+    <div className="securities-table-row">
+      <div className="securities-table-cell">{item.exchange}</div>
+      <div className="securities-table-cell">{item.symbol}</div>
+      <div className="securities-table-cell">{item.name}</div>
+      <div className="securities-table-cell">
```
AugusteBa: You could also do similar mapping here using `Object.keys`. Again - just a suggestion.

stockflux-securities-master/src/components/securities-table/TableRow.js
```diff
@@ -0,0 +1,82 @@
...
+    patchSecurity(item.securityId, {
+      visible: !item.visible
+    });
+  };
+
+  const handleToggleEnabled = () => {
+    patchSecurity(item.securityId, {
+      enabled: !item.enabled
+    });
+  };
```
AugusteBa: `handleToggleVisible` and `handleToggleVisible` are slightly similar. Maybe something like this would help reduce repetition:
```
  const handleToggle = (feature) => {
    patchSecurity(item.securityId, {
      [feature]: !item[feature]
    });
  };
```
Not sure if this would work though, needs testing. And again, being very picky here.

stockflux-securities-master/src/components/confirmation/Confirmation.js
```diff
@@ -1,10 +1,13 @@
...
 import PropTypes from "prop-types";
 
 const Confirmation = ({ children, confirmationText }) => {
   const [clickedStatus, setClickedStatus] = useState(false);
+  const handleClickNoButton = () => {
+    setClickedStatus(false);
+  };
 
   return (
     <div className="confirmation-button-container">
```
AugusteBa: Hmm, I tend to avoid naming classes containers, unless that class wraps some component. What I mean is that usually a class with a word "container" in it is used in the parent component to, e.g. position the child element. Aka I think it should not be inside the component itself, but again not sure. @JoelMagee what do you think?

JoelMagee: Was this mention meant for @joelwilliamsmith ;)

AugusteBa: Yes, indeed! Sorry for the mixup, not sure how this happened!

stockflux-securities-master/src/components/securities-table/TableRow.js
MaxOvenden: I don't know if this will work as well as for the headings as `Object.keys` gives all the keys of the securities item but in the object the fields are not in the correct order to display and there are certain fields which do are not displayed such as `securityId`, `updated`, `visible` and `enabled`.

How about `[item.exhange, item.symbol, item.name].map(cellData => <div className="securities-table-cell">{item.exchange}</div>)` ?

stockflux-securities-master/src/components/securities-table/TableRow.js
MaxOvenden: This can work but it requires the button's `onClick` to be called inline like `onClick={() => handleToggle("enabled")}`

I think @joelwilliamsmith advised against using a button's `onClick` in this way.

stockflux-securities-master/src/components/securities-table/SecuritiesTable.js
MaxOvenden: Yep done now

stockflux-securities-master/src/components/securities-table/SecuritiesTable.js
MaxOvenden: Good idea, I've done this now and changed `Edit / Delete` to `Options` as there are now more than just the two options.

stockflux-securities-master/src/components/confirmation/Confirmation.js
MaxOvenden: It probably shouldn't be named "confirmation-button-container" either because it isn't a button component anymore. Maybe just "confirmation". 

 In the same component I have "confirmation-message-container" and "confirmation-option-button-container" as well so I'll drop the word container on each of these?

MaxOvenden: I've made this name change now but will wait for Joel to comment.

MaxOvenden: I've now made the change to include shortening the classNames in many of the other components in which this kind of thing is currently done.

PR #1261, opened 2019-09-11
Admin View FE changes

AugusteBa:
N.B. Need Mark to approval UI changes.

##### SECURITIES
image
image
image

##### NO SECURITIES
image
image
image

##### EDIT SECURITY
image
image
image

##### DELETE SECURITY CONFIRMATION PROMPT (temporary as it should be replaced with OpenFin modal pop-up)
image
image
image

##### CREATE SECURITY
image
image
image

##### CREATE SECURITY ERROR
(Keep in mind that eventually, validation should be refactored in the BE to return input specific error messages and they could be displayed just under the input field itself, highlighting the field red)
image
image


markjose:
I'm not sure we should restrict Symbols to 4 characters max, I'd maybe make it 8 or 10 for now.
I'm a little confused byt the edit screen as it has the delete Yes/No question on the bottom. I'd imagined this would pop up in it's own modal window if you clicked the 'Bin' button.
Love the new styling

stockflux-securities-master/public/app.dev.json
```diff
@@ -14,12 +14,13 @@
         "minWidth": 850,
         "maxWidth": 850,
```
markjose: Do we want to set Max Width and Height, I'd imagined 'Maximize' should be allowed for this as moving forward we may have a number of properties on each security.

stockflux-securities-master/src/components/form-view/form/Form.js
```diff
@@ -0,0 +1,125 @@
...
+
+  const submitEdit = securityId => {
+    service
+      .updateSecurity(securityId, getSecurityDTO())
+      .then(async response => await processResponse(response, 200));
+  };
+
+  const submitNew = () => {
+    service
+      .postSecurity(getSecurityDTO())
```
markjose: The presence of this in this component further reinforces that this should be component state and not in the store. you can post this with useEffect here if it was in component state.

stockflux-securities-master/src/components/form-view/form-controls/text-field/TextField.js
```diff
@@ -0,0 +1,29 @@
...
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import './TextField.css';
+
+const TextField = ({ label, id, value, disabled, onChange }) => {
+  return (
+    <span className="input">
+      <label htmlFor={id}>{label}</label>
+      <input
+        className={classNames({ disabled: disabled })}
```
jrhenderson1988: You can change this to just `{ disabled }` and it should have the same effect, I think

AugusteBa: It would only work if I changed the value of `disabled` prop to a string `disabled`. For example:
```
        <TextField
          label="EXCHANGE"
          id="name"
          disabled={securityId && 'disabled'}
          value={state.exchange}
          onChange={event => dispatch(action.setExchange(event.target.value))}
        />
```

Changed that now.

stockflux-securities-master/src/components/form-view/form-controls/text-field/TextField.js
```diff
@@ -0,0 +1,29 @@
...
+      />
+    </span>
+  );
+};
+
+TextField.propTypes = {
+  label: PropTypes.string.isRequired,
+  id: PropTypes.string.isRequired,
+  value: PropTypes.string.isRequired,
+  disabled: PropTypes.string,
```
jrhenderson1988: Did you intend for this to be a string? Maybe a boolean would be better here then you could do `disabled={disabled ? 'disabled' : ''}` or something similar

stockflux-securities-master/src/components/form-view/form/Form.js
```diff
@@ -0,0 +1,116 @@
...
+  const [state, dispatch] = useReducer(securityReducer, initialState);
+
+  useEffect(() => {
+    if (securityId) {
+      service
+        .getSecurity(securityId)
+        .then(security => {
+          setSecurity(security);
+        })
+        .catch(() => {});
```
jrhenderson1988: What happens if the service fails to fetch a security? Are errors handled somewhere else maybe?

stockflux-securities-master/src/components/securities/Securities.js
```diff
@@ -0,0 +1,70 @@
...
+            />
+            <HorizontalRule />
+            {state.fetchStatus === FetchState.FETCHING ? (
+                <div className="spinner-container">
+                    <Components.Spinner />
+                </div>
+            ) : state.securities && state.securities.length === 0 ? (
+                <NoSecurities />
+            ) : (
+                        <>
```
jrhenderson1988: The indentation of this file seems to have changed. Wasn't it 2 spaces before?

stockflux-securities-master/src/components/form-view/form-controls/text-field/TextField.js
markjose: On Jonathons comment he is correct, a common abbreviation for classNames is cx. The cx function takes params as strings or objects, for objects it evaluates the value to true and uses the object prop name ast the string name of the class.

`<input className={cx({disabled})} />` will apply the class name disabled is the value of disabled is true.

I'd imagine with the new change you are theming based on `input[disabled]` or `input::disabled`

stockflux-securities-master/src/components/form-view/form/Form.js
markjose: if this is not handling anything then don't do .catch at all and let the exception bubble up. Please dont'consume and do nothing.

stockflux-securities-master/src/services/SecuritiesService.js
```diff
@@ -1,37 +1,37 @@
...
+  else {
+    throw new Error(
+      `Something went wrong while fetching security ${securityId}.`
+    );
   }
-  throw new ValidationError(json.messages);
 }
 
-export async function postSecurity(security) {
+export async function postSecurity(security, errorCallback, successCallback) {
```
jrhenderson1988: I don't think this approach is right. An `async` function effectively returns a promise. When you return something from inside an async function, that's the equivalent of resolving the promise with that value. When you throw an error inside of an async function, that's the same as rejecting the promise with the error you've thrown. Passing success/error callbacks into functions is an older way of doing things asynchronously and is one of the main things that promises were introduced to improve. They should really be mutually exclusive, in that you would use one approach or the other - generally promises (possibly using async/await) is preferred.

In this case, the function shouldn't accept any callback functions as parameters and you should just resolve (in this case, `return`, since it's an async function) what you'd want to pass to the success handler. You should reject (in this case `throw`, since it's an async function) what you'd want to pass to an error handler.

stockflux-securities-master/src/services/SecuritiesService.js
```diff
@@ -41,19 +41,29 @@ export async function postSecurity(security) {
     `${options.customData.apiBaseUrl}/admin/securities`,
     fetchOptions
   );
-  const json = await response.json();
-
-  if (response.ok) {
-    return json;
+  if (response.status === 201) {
```
jrhenderson1988: I think it might be better to use [`response.ok`](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok) here instead of explicitly checking the status code. This will be `true` if the status code falls within 200 - 299 and `false` otherwise. In this scenario nothing would change, but if we changed the success response code in the API from `201` to `200` or if it was originally `200` and we wanted a more descriptive `201` then your service would work regardless, whereas here it would fail

stockflux-securities-master/src/services/SecuritiesService.js
AugusteBa: Ok, I've added the ValidationError back in.

stockflux-securities-master/src/components/form-view/form/Form.js
AugusteBa: Totally agree, thanks for spotting this!

stockflux-securities-master/src/components/form-view/form-controls/text-field/TextField.js
AugusteBa: I might have misunderstood Jonathan's initial comment - I thought he suggested not using `classnames` altogether and just use `className={disabled}`. Fixed now though to use `cx`.

stockflux-securities-master/src/components/form-view/form-controls/text-field/TextField.js
AugusteBa: Just added Boolean casting to TextField:
```
<TextField
          label="SYMBOL"
          id="name"
          disabled={Boolean(securityId)}
          value={state.symbol}
          onChange={event => dispatch(action.setSymbol(event.target.value))}
        />
```

stockflux-securities-master/src/components/form-view/form/Form.js
jrhenderson1988: It might be worth adding some error handling instead of removing the catch here. I think it would be beneficial to display a message to the user if the service fails to fetch the security

AugusteBa: Yeah, actually, you're right. Just added it now.

PR #1277, opened 2019-09-16
Update launcher icons

Johnston97:
Added a round border to launcher icons. 
Before:
image

After:
image


packages/stockflux-launcher/public/searchResultsWindow.css
```diff
@@ -65,7 +65,12 @@ body {
   background-color: transparent;
   cursor: pointer;
   outline: none;
-  color: #95989C;
+  color: #37d8bc;
+  border: solid 1px #37d8bc;
+  border-radius: 50%;
+  width: 40px;
```
jrhenderson1988: Can we make the icons/buttons smaller?

packages/stockflux-launcher/public/searchResultsWindow.css
```diff
@@ -65,7 +65,12 @@ body {
   background-color: transparent;
   cursor: pointer;
   outline: none;
-  color: #95989C;
+  color: #37d8bc;
```
jrhenderson1988: Is there a hover state for the buttons? Would you be able to add a screenshot if so?

AugusteBa: Maybe we should create a variable for the color if it does not exist yet?

Johnston97:
On hover:
image

packages/stockflux-launcher/public/app.dev.json
```diff
@@ -2,7 +2,7 @@
   "devtools_port": 9091,
   "runtime": {
     "arguments": "--enable-aggressive-domstorage-flushing",
-    "version": "13.76.43.31"
+    "version": "12.69.43.22"
```
jrhenderson1988: I don't think we should commit this change until we've agreed to roll everything back to 12. I think it's advisable to use version 12 while developing because of the issues we're seeing with 13, for now though

PR #1278, opened 2019-09-16
Eslint Config for FE

AugusteBa:


packages/stockflux-bitflux/src/assets/js/chart/legend.js
```diff
@@ -2,47 +2,79 @@ import d3 from 'd3';
...
-    return legend;
+  var priceFormat;
+  var volumeFormat;
+  var timeFormat;
+  var textYOffset = '0.71em';
+
+  var tooltip = fc.chart
+    .tooltip()
+    .items([
+      [
```
jrhenderson1988: I wonder if this could be condensed to something like the following, which may result in the styling rules keeping things more concise:

```
.items([
  [ 'T', (d) => timeFormat(d.date)],
  [ 'O', (d) => priceFormat(d.open)],
  [ 'H', (d) => priceFormat(d.high)],
  [ 'L', (d) => priceFormat(d.low)],
  [ 'C', (d) => priceFormat(d.close)],
  [ 'V', (d) => volumeFormat(d.volume)]
])
```

AugusteBa: Very good point, will reduce it now.

packages/stockflux-launcher/src/app-shortcuts/AppShortcuts.js
```diff
@@ -25,26 +25,11 @@ export default () => {
       {apps
         .filter(
           app =>
-            app.customConfig !== undefined &&
-            app.customConfig.showInLauncher &&
-            app.appId.indexOf("stockflux-") === 0
```
jrhenderson1988: Something has gone wrong here. Your PR which refactors the app shortcuts seems to have disappeared. I noticed because the check I added here to ensure that the `appId` begins with `stockflux-` which I added into your PR has been deleted.

AugusteBa: Yeah, I accidentally chose wrong changes when resolving conflicts. Should be fixed now, thanks for noticing.

packages/stockflux-watchlist/public/notificationHandler.js
```diff
@@ -1,4 +1,4 @@
-// eslint-disable-next-line no-unused-vars
+/*eslint-disable-next-line no-unused-vars*/
```
jrhenderson1988: I don't see any unused vars here. Might be worth just removing this comment. I wonder why eslint insisted on changing this to a multiline style comment `/**/` instead of single line style `//`

packages/stockflux-watchlist/src/components/watchlist-card/WatchlistCard.js
```diff
@@ -81,42 +88,45 @@ const WatchlistCard = props => {
...
         <div className="card darkens default-background" draggable="false">
           <div className="card-top">
             <div className="details-container">
               <div className="name">{stockData.name}</div>
-              <div className="symbol">{props.symbol}</div>
+              <div className="symbol">{symbol}</div>
             </div>
             <div className="icons">
-              <Components.NewsShortcut
+              <div
```
jrhenderson1988: I think this might have regressed as well. It looks like the changes here are a move back to how it was before you introduced the shortcut components

AugusteBa: Yeah, thanks for spotting this, should be fixed now.

packages/stockflux-watchlist/public/notificationHandler.js
AugusteBa: It's needed for the function itself as eslint thinks it's unused
image

Also changed it to single line comment now.

.circleci/config.yml
```diff
@@ -25,12 +33,129 @@ jobs:
       - run:
           name: Root package install
           command: npm ci
+      - save_cache:
```
jrhenderson1988: The `save_cache` command here is running after we run `npm ci`, which installs a fresh bunch of `node_modules`. So in this case we're not really benefiting from caching between CircleCI builds where the `package.json` file has not changed. I've also read some articles (e.g. [this one](https://hackernoon.com/circleci-performance-difference-between-cache-and-workspace-5567679c3601)) which suggest that persisting to workspace actually ends up being faster than cache restores and cache saves. When we're going for the caching approach, we also have the overhead of having to checkout the repository on every step too.

.circleci/config.yml
```diff
@@ -25,12 +33,129 @@ jobs:
...
+    docker:
+      - image: circleci/node:10.15.1
+    steps:
+      - checkout:
+          path: ~/stockflux
+      - restore_cache:
+          key: deps-{{ checksum "package-lock.json" }}
+      - run:
+          name: Restore Bitflux Cache
+          command: cd ~/stockflux/packages/stockflux-bitflux
```
jrhenderson1988: I'm pretty sure that the `run` step runs a new shell each time and is rooted at the `working_directory` by default, each time it is executed. This means that this command which changes to the `~/stockflux/packages/stockflux-bitflux` directory effectively does nothing. The following step of restoring the cache with the key `deps-bitflux-{{ checksum "package-lock.json" }} is actually using the `~/stockflux/package-lock.json` as the source file for the checksum, which isn't what we want.

.circleci/config.yml
```diff
@@ -25,12 +33,129 @@ jobs:
...
           command: npm run bootstrap
       - run:
-          name: Build packages
+          name: Build Packages
           command: npm run build
+      - run:
+          name: Save Bitflux Cache
+          command: cd ~/stockflux/packages/stockflux-bitflux
+      - save_cache:
+          key: deps-bitflux-{{ checksum "package-lock.json" }}
```
jrhenderson1988: A similar situation here, the `deps-bitflux-{{ checksum "package-lock.json" }}` key is using the root `~/stockflux/package-lock.json` file as the checksum in its cache key and we're actually caching the root `~/stockflux/node_modules` directory, which isn't what we want. I think the way this should be done would be to specify the full paths to the `~/stockflux/packages/stockflux-bitflux/package-lock.json` file for the checksum and `~/stockflux/packages/stockflux-bitflux/node_modules` directory for the `paths`. I think the build probably actually works because we're running `npm run bootstrap` above which installs all of the packages using Lerna.

AugusteBa:
> All in all, I think this PR is ready to be merged. I do have some concerns about the CircleCI configuration changes though. With your changes here I think the build will still succeed because we're still running `npm ci` in the root and we're running `npm run boostrap` which installs all of the `node_modules` for each package. I don't think the caching is doing what we intended and I think we're effectively doing a lot of extra work that we don't need (such as repeatedly caching the root `node_modules` with different keys that use the same checksum), which could be avoided by simply persisting to workspace.
> 
> Personally, I think we should revert these changes for now and perhaps investigate improving the CircleCI build as a separate task. I think this one should be focused strictly on adding ESLint and Prettier configuration.

Thanks for reviewing. I've reverted these changes now.

jrhenderson1988:
Approved. Great work getting this done!

Issue #1286, opened 2019-09-24
Create a more generic child window/popup modal.


AugusteBa:
Also includes fixing the icons in the SearchResults dropdown.

PR #1290, opened 2019-09-30
Generic Child Window and useChildWindow Hook

AugusteBa:
This PR refers to #1286 and introduces `useChildWindowHook`, which should eventually appear in the `openfin-react-hooks` as well.

image

image

image

image

image

image


packages/stockflux-launcher/src/free-text-search/FreeTextSearch.js
```diff
@@ -5,117 +5,106 @@ import React, {
...
-    event => {
-      setQuery(event.target.value);
-      if (event.target.value !== null && event.target.value === '') {
-        closeResultsWindow();
-      } else if (!resultsWindow && !isCreatingResultWindow) {
-        isCreatingResultWindow = true;
-        launchSearchResultsWindow(
+  const { window, launch, populateDOM, close } = childWindow;
+
+  const parentUuid = OpenfinApiHelpers.getCurrentWindowSync()
```
jrhenderson1988: I think there's a `getCurrentWindowOptions` helper function in openfinApiHelpers.js which could simplify this a little bit.

packages/stockflux-launcher/src/search-results/Launcher.js
```diff
@@ -33,16 +32,19 @@ export default async (
...
   };
 
   const win = await OpenfinApiHelpers.createWindow(childWindow);
 
-  if (dockedTo === ScreenEdge.LEFT || dockedTo === ScreenEdge.RIGHT) {
-    win
-      .getWebWindow()
-      .document.getElementById('searchbar-container').hidden = false;
-  }
+  // if (dockedTo === ScreenEdge.LEFT || dockedTo === ScreenEdge.RIGHT) {
```
kangelSL: Does this commented out code need to be deleted?

packages/stockflux-launcher/src/reducers/child-window/Actions.js
```diff
@@ -0,0 +1,8 @@
+export default {
```
jrhenderson1988: I'd be inclined to make these all uppercase with underscore separators since this effectively functions like an enum. Just personal preference though - if there's an existing convention then I'd stick to that. Not something I think should stop this PR though,

packages/stockflux-launcher/src/search-results/helpers/Launcher.js
```diff
@@ -0,0 +1,50 @@
...
+    waitForPageLoad: true,
+    alwaysOnTop: false,
+    maxWidth: 400,
+    minWidth: 400,
+    minHeight: 300
+  };
+
+  const win = await OpenfinApiHelpers.createWindow(childWindow);
+
+  // if (dockedTo === ScreenEdge.LEFT || dockedTo === ScreenEdge.RIGHT) {
```
kangelSL: Same here too

packages/stockflux-launcher/src/reducers/child-window/ChildWindow.js
```diff
@@ -0,0 +1,26 @@
...
+
+export const initialChildWindowState = {
+  window: null,
+  state: childWindowState.initial
+};
+
+export default (state, action) => {
+  switch (action.type) {
+    case actionType.changeState:
+      if (!Object.keys(childWindowState).indexOf(action.payload))
```
jrhenderson1988: I think there's a potential for a bug here. `indexOf` returns a number which refers to the position of the given parameter, which could be `0`. `-1` will be returned if the given parameter does not exist. `0` is considered a falsy value whereas `-1` is considered truthy. I assume that the intention here is to make sure that the state is one of the expected values, so it would probably be better to check explicitly for `-1` when using `indexOf` or perhaps use `.includes(...)` which reads a bit better.

packages/stockflux-launcher/src/search-results/Launcher.js
jrhenderson1988: You can probably remove this comment now

packages/stockflux-launcher/src/search-results/helpers/Launcher.js
jrhenderson1988: This can be removed too

jrhenderson1988:
@kangelSL I believe I should be able to clarify the second question you had:

> Is the search case sensitive? E.g. if a company has an all-caps name would it appear first in the list if searched for using all-caps?

This is handled by the Securities API Lambda in the backend/cloud. The frontend simply passes through what it receives in the input and the backend handles the search scoring. It's been a while since I looked at it but I believe that the backend normalises the query and the symbols/company names it compares to lowercase - therefore case is entirely ignored when scoring results. It does however, score

packages/stockflux-launcher/src/search-results/Launcher.js
AugusteBa: @kangelSL, you asked about the bit where you click away and the search remains open and if the window should close and the search bar clear. This PR does not add additional functionality, but rather introduces useChildWindow hook and ensured the launch buttons in the child window work. 

Regarding the second question, I think @jrhenderson1988 answered this perfectly and I have nothing to add! :)

> A couple of small things, but I was also wondering:
> 
> * When clicking away, the search remains open. Should the window close and the search bar clear if they click away?
> * Is the search case sensitive? E.g. if a company has an all-caps name would it appear first in the list if searched for using all-caps?

AugusteBa:
@jrhenderson1988, @kangelSL,

I addressed your comments now, thanks for all the observations!

packages/stockflux-launcher/src/free-text-search/FreeTextSearch.js
```diff
@@ -5,117 +5,106 @@ import React, {
...
-      if (event.target.value !== null && event.target.value === '') {
-        closeResultsWindow();
-      } else if (!resultsWindow && !isCreatingResultWindow) {
-        isCreatingResultWindow = true;
-        launchSearchResultsWindow(
+  const { window, launch, populateDOM, close } = childWindow;
+
+  const parentUuid = OpenfinApiHelpers.getCurrentWindowOptions().then(
+    options => options.uuid
+  );
```
jrhenderson1988: `parentUuid` will actually be a promise here, rather than a string/uuid as expected. My guess is that you want the UUID of the current window to pass to the `useInterApplicationBusSubscribe` hook below. I'd personally do something like this to achieve it:

```
const [parentUuid, setParentUuid[ = useState();

useEffect(async () => {
  const options = OpenFinApiHelpers.getCurrentWindowOptions();
  setParentUuid(options.uuid);
});

```

AugusteBa:
Yeah, we're aware of this, it's a much bigger issue and requires it's own PR. It will happen whenever you send a message and there's no app subscribed to that topic. We will eventually perhaps have a message hub in the launcher, which would deal with ensuring that apps are launched before they receive messages.

> One potential issue but LGTM otherwise
> 
> When I don't have watchlist open or anything subscribed to it and use the search, then click watchlist on one of the items, I get this error:
> 
> image
> 
> (It works fine if the watchlist is open already though)

packages/stockflux-launcher/src/reducers/child-window/ChildWindow.js
```diff
@@ -0,0 +1,26 @@
+import STATE_TYPE from './State';
+import ACTION_TYPE from './Action';
```
jrhenderson1988: We already discussed this but not sure if you've gotten around to it yet so I thought I'd leave a comment just for your reference. Can we change these to `StateType` and `ActionType`?

packages/stockflux-launcher/src/reducers/free-text-search/FreeTextSearch.js
```diff
@@ -1,37 +1,34 @@
-export const SEARCHING = 'searching';
-export const SUCCESS = 'success';
-export const ERROR = 'error';
-export const INITIALISE = 'initialise';
+import ACTION_TYPE from './Action';
```
jrhenderson1988: Same here, `ActionType`

packages/stockflux-launcher/src/reducers/search-result/SearchResult.js
```diff
@@ -1,34 +1,31 @@
-export const SEARCHING = 'searching';
-export const SUCCESS = 'success';
-export const ERROR = 'error';
-export const INITIALISE = 'initialise';
+import ACTION_TYPE from './Action';
```
jrhenderson1988: Same here, `ActionType`

packages/stockflux-launcher/src/search-results/ChildWindow.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import Components from 'stockflux-components';
+import MESSAGES from './messages';
```
jrhenderson1988: Same here, `Messages` instead of `MESSAGES`

packages/stockflux-launcher/src/search-results/helpers/useChildWindow.js
```diff
@@ -0,0 +1,139 @@
+import { useReducer, useEffect, useCallback } from 'react';
+import ReactDOM from 'react-dom';
+import reducer, {
+  initialChildWindowState
+} from '../../reducers/child-window/ChildWindow';
+import STATE_TYPE from '../../reducers/child-window/State';
+import ACTION_TYPE from '../../reducers/child-window/Action';
```
jrhenderson1988: And the same here (I think this is the last one) - `StateType` and `ActionType`

packages/stockflux-launcher/src/free-text-search/FreeTextSearch.js
```diff
@@ -5,117 +5,109 @@ import React, {
...
-  SEARCHING,
-  SUCCESS,
-  ERROR,
-  INITIALISE
-} from './FreeTextSearch.reducer';
+import { StockFlux } from 'stockflux-core';
+import reducer, {
+  initialSearchState
+} from '../reducers/free-text-search/FreeTextSearch';
+import SEARCH_ACTION from '../reducers/free-text-search/Action';
```
jrhenderson1988: One more here :]

PR #1291, opened 2019-10-07
Feat: news app links

kangelSL:
For #1140 

* Add button links to StockFlux news on header

Appearance:

image

Double line:

image

### DIFF
--- a/packages/stockflux-news/src/App.js
+++ b/packages/stockflux-news/src/App.js
@@ -91,7 +91,14 @@ function App() {
   return (
     <div className="stockflux-news">
       <Components.Titlebar />
-      <div className="header">{name ? name : symbol} News</div>
+      <div className="header">
+        <div>{name ? name : symbol} News</div>
+        <div>
+          {' '}
+          <Components.Shortcuts.Chart small={true} />
+          <Components.Shortcuts.Watchlist small={true} />
+        </div>
+      </div>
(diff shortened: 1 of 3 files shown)


packages/stockflux-news/src/App.js
```diff
@@ -91,7 +91,22 @@ function App() {
...
+          {' '}
+          <Components.Shortcuts.Chart
+            small={true}
+            symbol={symbol}
+            name={name}
+          />
+          <Components.Shortcuts.Watchlist
+            symbol={symbol}
+            name={name}
+            small={true}
```
SL-SWoods: for readability would it be worth putting the attributes in the same order?

packages/stockflux-news/src/App.js
```diff
@@ -91,7 +91,22 @@ function App() {
   return (
     <div className="stockflux-news">
       <Components.Titlebar />
-      <div className="header">{name ? name : symbol} News</div>
+      <div className="header">
+        {name ? name : symbol} News
+        <div className="icons">
+          {' '}
```
AugusteBa: Is this `{' '}` intentional? If this is to ensure the space between text and buttons, maybe we should add `margin-left` instead?

kangelSL: This is a mistake, think it was added by the linter and left in instead of being removed

PR #1298, opened 2019-10-11
Create preview windows for watch list charts. 

SL-SWoods:
Add semi-transparent preview displays to watch list. Pull request consists of - 
- When a hover is started on a watch list item a preview window is displayed, indicating you can drag the item outside to open it up larger. 
image
- Created a `PreviewWindow` component within `stockflux-components`.
- Moved `useChildWindow` into `stockflux-core` to enable use between different packages. 
- Added a position override to the Chart launcher, to allow it to open in a different position, recommend this is added to other launchers as well. 

**Preview Window Component** 
Component is available from within `stockflux-components` under `Components.PreviewWindow`.
e.g use
```javascript
<Components.PreviewWindow
  display={bool}
  htmlfile="previewImage.html"
  position={ { top: 0, left: 0 } }
  size={ { height: 0 , width: 0 } } >
{children}
</Components.PreviewWindow>
```
**Watch List Chart Preview**
Watch list preview component will show either on the left or right of the watch list, dependent on if there is enough space for the resulting popup window to view.


packages/stockflux-components/src/components/preview-window/PreviewWindow.js
```diff
@@ -0,0 +1,104 @@
...
+  }, [display, window]);
+
+  useEffect(() => {
+    window &&
+      window.moveTo(position.left, position.top, { moveindependently: true });
+  }, [position, window]);
+
+  useEffect(() => {
+    window &&
+      window.resizeTo(size.width, size.height, 'top-left', {
```
kangelSL: The window has setting `resizeable: false` above, is a resize effect needed?

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
```diff
@@ -114,18 +122,81 @@ const Watchlist = () => {
...
+        0
+          ? window.outerWidth + screenLeft + windowOffset
+          : screenLeft - windowOffset - previewOptions.defaultWidth;
+      setPreviewPosition({
+        left:
+          leftPosition > window.screen.availLeft
+            ? window.screenLeft > 0
+              ? leftPosition
+              : window.screen.availLeft + leftPosition
+            : window.outerWidth + screenLeft + windowOffset,
```
kangelSL: Is there any way of using a fixed position to reduce the amount of ternary statements needed?

packages/stockflux-components/src/components/preview-window/PreviewWindow.js
SL-SWoods: The resizeable:false only disabled the user from being able to resize on the screen, for the preview we never want to the user to interact with it so i made it unresizeable. 

The resize effect can still be called within code and it allows the preview window to be resized to the correct size on display, and can change dependent on whats being previewed. This is used to change the windows size from its initial to the same size as the charts window after its initial creation.

packages/stockflux-components/src/components/preview-window/PreviewWindow.js
```diff
@@ -0,0 +1,104 @@
...
+}) => {
+  const TRANSPARENT_WINDOW_NAME = 'transparent_window';
+  const TRANSPARENT_CSS_PATCH = 'transparentWindow.css';
+
+  const windowStatus = {
+    show: 'SHOW',
+    showing: 'SHOWING',
+    hiding: 'HIDING',
+    hide: 'HIDE',
+    hidden: 'HIDDEN'
```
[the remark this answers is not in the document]
SL-SWoods: The transitioning states were added to deal with the animation of the window fading in and out, as there were issues if you kept making the preview appear and disappear within its animation times.

For the example we are using on the watch list there is low chance of causing this now after it was changed to display on hover start, and not when you leave the window. But as this should be reusable in other areas, this will be required for where the transparent windows and display/hide rapidly and in succession.
The 'Show' status was completely unneeded though so its been removed.

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
SL-SWoods: I have tried to avoid using fixed positions, as the watch list window can be moved around, and i would prefer the popup located to be around that. The position is only calculated when the box is to be shown, as this handled is the parent window has been moved.  
The reason there is so many checks on positions as it should handle multiple monitors correctly as well with showing the box on the side which has the most space to fit it within without overlapping the watch list box itself.

packages/stockflux-core/src/custom-hooks/useChildWindow/reducers/ChildWindow.js
```diff
@@ -0,0 +1,26 @@
...
+
+export const initialChildWindowState = {
+  window: null,
+  state: childWindowState.initial
+};
+
+export default (state, action) => {
+  switch (action.type) {
+    case actionType.changeState:
+      if (!Object.keys(childWindowState).indexOf(action.payload))
```
kangelSL: Is this likely to occur or is it more a sanity check?

packages/stockflux-components/src/components/preview-window/PreviewWindow.js
```diff
@@ -0,0 +1,102 @@
...
+        autoShow: false,
+        showTaskbarIcon: false,
+        defaultHeight: size.height,
+        defaultWidth: size.width,
+        url: htmlfile,
+        frame: false,
+        resizable: false,
+        alwaysOnTop: true
+      }),
+    [launch]
```
AugusteBa: I know`react-hooks/exhaustive-deps` is a strict rule, but sometimes it does help to pinpoint dependencies. Shouldn't these be included in the dependency array: 'htmlfile', 'size.height', and 'size.width'? It's ok if they should not, just double checking.

packages/stockflux-components/src/components/preview-window/PreviewWindow.js
```diff
@@ -0,0 +1,102 @@
...
+        .then(() => setPopupStatus(windowStatus.hide));
+    }
+  };
+
+  useEffect(() => {
+    if (popupStatus === windowStatus.hide && display === false) {
+      window.hide();
+      setPopupStatus(windowStatus.hidden);
+    }
+  }, [popupStatus, window]);
```
AugusteBa: Same here: React Hook useEffect has missing dependencies: 'display', 'windowStatus.hidden', and 'windowStatus.hide'. Should we ignore these?

packages/stockflux-components/src/components/preview-window/PreviewWindow.js
```diff
@@ -0,0 +1,102 @@
...
+
+  useEffect(() => {
+    if (display) {
+      !window && launchChildWindow();
+      showWindow();
+    } else {
+      hideWindow();
+    }
+  }, [display, window]);
```
AugusteBa: And here: React Hook useEffect has missing dependencies: 'hideWindow', 'launchChildWindow', and 'showWindow'.

packages/stockflux-core/src/custom-hooks/useChildWindow/reducers/ChildWindow.js
AugusteBa: This is checking if a valid state has been passed in, aka if `action.payload` state is one of `childWindowState`. It is unlikely to occur, but I think it's better to handle a case where we pass in an invalid state that to assume that it always comes in valid.

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
```diff
@@ -24,6 +23,10 @@ const getDistinctElementArray = array => [...new Set(array)];
 const Watchlist = () => {
   const [dragOverIndex, setDragOverIndex] = useState(null);
   const [unwatchedSymbol, setUnwatchedSymbol] = useState(null);
+  const [displayPreview, setDisplayPreview] = useState(false);
+  const [previewOptions, setPreviewOptions] = useState();
+  const [previewPosition, setPreviewPosition] = useState({ top: 0, left: 0 });
+  const [previewSize, setPreviewSize] = useState({ height: 300, width: 300 });
```
AugusteBa: Just a suggestion - maybe it would better to hold all preview state in a single `useState` object? 
```
const [preview, setPreview] = useState(
  {
    options: null,
    position: { top: 0, left: 0 },
    size: { height: 300, width: 300 },
  }
);
```

And the use it with object spread like this:

`setPreview({...preview, size: { height: 500, width: 500 }})`

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
```diff
@@ -86,6 +89,11 @@ const Watchlist = () => {
 
   const getSymbolIndex = symbol => watchlist.indexOf(symbol);
 
+  const onDropOutside = (symbol, stockName) => {
+    // the 3rd param is for intents enabled.
+    Launchers.launchChart(symbol, stockName, null, previewPosition);
```
AugusteBa: Again, just a suggestion, but since it's very unlikely that we will be using intents soon, I would suggest reordering arguments so that the intents flag would be the very last argument.

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
```diff
@@ -114,18 +122,81 @@ const Watchlist = () => {
...
+  };
+
+  useEffect(() => {
+    getWindowOptions().then(value => {
+      setPreviewOptions(value);
+    });
+  }, []);
+
+  useEffect(() => {
+    const windowOffset = 5;
```
AugusteBa: What is 5? Also maybe we should declare this const using `UPPER_CASE_SNAKE_CASE`?

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
AugusteBa: Maybe we should split these into small helper functions, aka each ternary should be its own function? I think it might make it more readable.

AugusteBa:
Overall looks good to me, just a couple of small comments above.

Also, the dotted border seems to be a couple of pixels off (see 2 and 3)? And there seems to be additional dark border (see 1).
image

Lastly, I've noticed that a couple of times when I launched the watchlist and dragged the card out, for a brief second a white square-shaped window would appear and then reshape itself into the preview window. I've tried snapping a screenshot, but unfortunately, my attempts were unsuccessful and later the problem did not manifest itself anymore. Have you seen this before? Could it be the size `300/300` that we set initially?

Johnston97:
> Also, the dotted border seems to be a couple of pixels off (see 2 and 3)? And there seems to be additional dark border (see 1).
> image

I noticed the border being slightly off too. It looks like if you set the width/height to be the full 100vw/vh the windows scrollbar gets in the way. Since this is just a preview the scrollbars could be removed.

Another thing I noticed was when you drag a chart away from the watch list the mouse on-drag turns into a 'no symbol' which looks contradicting to the 'open a chart here' text

packages/stockflux-core/src/custom-hooks/useChildWindow/reducers/ChildWindow.js
SL-SWoods: I am not sure, the childWindow is a copy of the existing one, which I have moved from stockflux-launcher to stockflux-core. Eventually this will be replaced by the open-find-react-hooks library once the changes are merged in #1296.

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
SL-SWoods: A random small number to make sure there is a gap between the 2 windows, making them clearer to see.
Changed the constant to use the correct case.

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
SL-SWoods: combined the  `previewOptions`, `previewPosition` and `previewSize` together into `previewDetails`. Added `// eslint-disable-next-line react-hooks/exhaustive-deps` to the hooks, but they are correct as I only want them to first on specific values changing.

packages/stockflux-components/src/components/preview-window/PreviewWindow.js
SL-SWoods: I have added `// eslint-disable-next-line react-hooks/exhaustive-deps` to this and the other ones in this file. I have checked through dependencies and they are not needed for the hooks in question.

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
SL-SWoods: Reordered the inputs to `Launchers.launchChart` to have position before intents. Added parameters to the button shortcut for chart to allow you to pass in positions, this prevents having a `null` argument in `app-shortcutes/Chart.js`

packages/stockflux-components/src/components/preview-window/PreviewWindow.js
SL-SWoods: Fixed dependencies

packages/stockflux-components/src/components/preview-window/PreviewWindow.js
SL-SWoods: Fixed dependencies

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
SL-SWoods: Refactored the code to make the function more readable.

SL-SWoods:
> > Also, the dotted border seems to be a couple of pixels off (see 2 and 3)? And there seems to be additional dark border (see 1).
> > image
> 
> I noticed the border being slightly off too. It looks like if you set the width/height to be the full 100vw/vh the windows scrollbar gets in the way. Since this is just a preview the scrollbars could be removed.
> 
> Another thing I noticed was when you drag a chart away from the watch list the mouse on-drag turns into a 'no symbol' which looks contradicting to the 'open a chart here' text

I have noticed this as well, I have been trying to to find a way around this to make it show the correct icon, but nothing working as of yet, i would say we should raise it a separate issue.

AugusteBa:
I'm really being picky here, but I think the `border-bottom` is still 1px off (cut midway through):
image

SL-SWoods:
> I'm really being picky here, but I think the `border-bottom` is still 1px off (cut midway through):
> image

Tweaked with the CSS a bit, and it now should show the full border. For some reason with `height:100vh` it didn't want to show the last pixel.

packages/stockflux-components/src/components/preview-window/PreviewWindow.js
```diff
@@ -0,0 +1,105 @@
...
+    TRANSPARENT_CSS_PATCH
+  );
+
+  const { window, launch } = childWindow;
+
+  const launchChildWindow = useCallback(
+    () =>
+      launch({
+        name: TRANSPARENT_WINDOW_NAME,
+        waitForPageLoad: true,
```
kangelSL: The first time you drag the chart outside of the watchlist, there's a delay before it loads. The rest of the time it loads quickly, just wondering if this was the reason why and if it can load any quicker on the first load?

packages/stockflux-components/src/components/preview-window/PreviewWindow.js
```diff
@@ -0,0 +1,105 @@
...
+        url: htmlfile,
+        frame: false,
+        resizable: false,
+        alwaysOnTop: true
+      }),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [launch]
+  );
+
+  const [popupStatus, setPopupStatus] = useState('');
```
kangelSL: Should this be `hidden` as default?

SL-SWoods: Indeed it should be, changed to `windowStatus.hidden`

packages/stockflux-components/src/components/preview-window/PreviewWindow.js
SL-SWoods: This is because the first time the window has to appear it has to wait for the childwindow to be created and populated, which has a slight delay which I have tried to keep as small as possible by not having the popup have any dynamic or react components.
Every subsequent time it is just show/hiding the window so it loads faster.

packages/stockflux-components/src/components/buttons/round-buttons/app-shortcuts/Chart.js
```diff
@@ -3,10 +3,19 @@ import { Launchers } from 'stockflux-core';
...
+  small,
+  disabled,
+  intentsEnabled,
+  position
+}) => (
   <RoundButton
     className="shortcut chart"
-    onClick={() => Launchers.launchChart(symbol, name, intentsEnabled)}
+    onClick={() =>
+      Launchers.launchChart(symbol, name, position, intentsEnabled)
```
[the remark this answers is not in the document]
SL-SWoods: It was requested to change the order of the inputs on Launchers.launchChart to have intents last (https://github.com/ScottLogic/StockFlux/pull/1298#discussion_r335376553) as the order is important to that function, yet this one it doesn't matter due to the deconstruction on the export line.

SL-SWoods:
> Mostly little comments & nitpicks. Noticed that if you drop the chart outside the dotted drop zone the chart still appears but assume this is the correct behaviour?

Currently if you drop it anywhere outside the watch-list window it will open up the chart, but it will be placed in the previewed area.

Issue #1266, opened 2019-09-12
Launcher - Cloud mode


jrhenderson1988:
- Hide app launcher buttons
- Hide docking buttons and remove docking functionality
- Only grab handle, search bar and button should be visible
- Width should be sensibly adjusted
- Cloud mode should be the default when the app is registered with and launched from the OpenFin Cloud service

PR #1299, opened 2019-10-11
WIP: Cloud standalone

kangelSL:
PR for issue #1266 

* Will work in tandem with PR on StockFlux Cloud (https://github.com/ScottLogic/StockFlux-Cloud/pull/81)
* New custom data flags, `initialDocked` and `cloudMode` . `initialDocked` determines if app is started in standalone mode or in docked mode. `cloudMode` determines if the undock option should show at all. 
* Changed styling to keep borders separate from item alignment, to work with the new undocked mode

** Files UndockedWindowHook and UndockedWindow transition are mocks of unreleased code from open fin hooks. The code will need to be refactored to use the hook when it is released on the hooks repo **

To dos:
* Discussed on load bug on StockFlux\packages\stockflux-launcher\node_modules\openfin-react-hooks\src\useDockWindow.ts with @jrhenderson1988 and @AugusteBa which will need a fix on a separate PR 

Standalone mode: 
image
Docked mode:
image


kangelSL:
Made cleaner version without mock files at #1303

Issue #1294, opened 2019-10-09
Fix scrollbar in results-window


AugusteBa:

SL-SWoods:
[Pull request ](https://github.com/ScottLogic/StockFlux/pull/1300) opened.

PR #1300, opened 2019-10-11
#1294 Search Results Scroll Behaviour

SL-SWoods:
Change the styling of the scroll results sections to use the same components for handling scrolling sections. 
The scroll  bar section should fade in when there is results to scroll and you are over the window.


SL-SWoods:
In regards to the scroll bar, the same effect happens on the Watch List and News as well. We would require some insight from UX in if we want the scroll bars visible at all times. 

When we have a design to stick by changing the ScrollWrapperX component should fix all instances as they all use the same methods now.

Issue #1297, opened 2019-10-09
Undocked search window load issue


kangelSL:
Expected behaviour: search dropdown results stays beneath/adjacent to the search bar. 

Actual behaviour: strange behaviour occurs when the launcher is undocked.

The search dropdown appears next to the launch bar. 

![image](https://user-images.githubusercontent.com/51117162/66496297-9f590380-eab2-11e9-9a86-1d27f7365fc7.png)

PR #1301, opened 2019-10-14
1297 undocked launcher search box

SL-SWoods:
From Issue #1297 When the launcher bar is undocked the search results box appears in the 0,0 position. 

image
image

Fixed to work out where it should be showing from the window, works in both formats of the launcher (Horizontal / Vertical) and will handle if the launcher is dragged onto non-primary monitors. Should work on all sizes of launchers.

### DIFF
--- a/packages/stockflux-launcher/src/search-results/helpers/Positioner.js
+++ b/packages/stockflux-launcher/src/search-results/helpers/Positioner.js
@@ -17,7 +17,7 @@ export default (searchButtonRef, inputRef, dockedTo, windowBounds) => {
     case ScreenEdge.LEFT:
       return {
         defaultTop: parseInt(searchButtonRect.top),
-        defaultLeft: DEFAULT_LAUNCHER_SIZE,
+        defaultLeft: windowBounds.right,
         defaultWidth: DEFAULT_SEARCH_RESULTS_SIZE,
         defaultHeight: DEFAULT_SEARCH_RESULTS_SIZE
       };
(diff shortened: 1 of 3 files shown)


AugusteBa:
> On dragging the search box doesn't disappear or stick to the search bar. Should this be the case? If so, should we allow the search box to allow the user to drag it too, since we can decouple one from the other?

I think it would be nice if we could make it drag together, yet I think it might be not very easy to do. Is this something you have looked at already @SL-SWoods? If not, then maybe we should create a new ticket (or you can just edit this one to address this issue)? If it turns out to be hardly possible or hacky, we might need to resort to just closing the search window when the user clicks on the drag icon to prevent these two from appearing out of sync.

packages/stockflux-launcher/src/search-results/helpers/Positioner.js
```diff
@@ -29,11 +29,31 @@ export default (searchButtonRef, inputRef, dockedTo, windowBounds) => {
...
+          defaultWidth: DEFAULT_SEARCH_RESULTS_SIZE,
+          defaultHeight: DEFAULT_SEARCH_RESULTS_SIZE
+        };
+      } else {
+        const rightPosition =
+          windowBounds.left + windowBounds.width + DEFAULT_SEARCH_RESULTS_SIZE;
+        let leftPosition = windowBounds.left + windowBounds.width;
+        if (rightPosition > window.screen.width) {
+          leftPosition = windowBounds.left - DEFAULT_SEARCH_RESULTS_SIZE;
+        } else {
```
AugusteBa: could do `else if {` instead of `else { if {`

SL-SWoods: Good spot, changed to make it more readable.

SL-SWoods:
> On monitor setup 1 | 2, with monitor docked on 2 at left, opening search box opens far left on monitor 1:
> image

Changed the reference position for dealing with left alignments to work from window position not the primary display position.

Issue #1154, opened 2019-06-27
Can't open a new chart/news app with the same symbol as an existing chart


JackG188:
Due to how child windows are managed we don't allow multiple chart with the same symbol as it is used to create the window name. We should either allow this, e.g. add an number onto the end of the window name or show a dialogue explaining the decision to not create the new chart

markjose:
New UX means this is not a feature - won't do

PR #1302, opened 2019-10-15
Flag to allow multiple instances of window

SL-SWoods:
For issue #1154, added a flag to the Chart and News launcher which allow multiple instances of them to be open at one time. 
image
Each instance of a window after the first has a number appended to its name, numbers are reused again if the window is closed, and a new one spawned. 
Within the launcher if you pass `options.allowMultiple = true` to LaunchAsChildWindow this will allow multiple instances, if the flag is missing or false it reverts to previous behaviour and only allows a single instance of the screen to show. 

This only works with lauchAsChildWindow, and does not work with intent launchers.

### DIFF
--- a/packages/stockflux-core/src/services/ChildWindowLauncher.js
+++ b/packages/stockflux-core/src/services/ChildWindowLauncher.js
@@ -1,8 +1,8 @@
 import * as OpenfinApiHelpers from '../openfin-api-utils/openfinApiHelpers';
 
 const windowAlreadyExist = async (windowName, childWindows) => {
-  for (let i = 0; i < childWindows.length; i++) {
-    const childWindowOptions = await childWindows[i].getOptions();
+  for (const childWindow of childWindows) {
+    const childWindowOptions = await childWindow.getOptions();
     if (childWindowOptions.name === windowName) {
       return true;
     }
--- a/packages/stockflux-core/src/services/ChildWindowLauncher.js
+++ b/packages/stockflux-core/src/services/ChildWindowLauncher.js
@@ -1,18 +1,44 @@
 import * as OpenfinApiHelpers from '../openfin-api-utils/openfinApiHelpers';
 
+const windowAlreadyExist = async (windowName, childWindows) => {
+  for (let i = 0; i < childWindows.length; i++) {
+    const childWindowOptions = await childWindows[i].getOptions();
+    if (childWindowOptions.name === windowName) {
+      return true;
+    }
+  }
+  return false;
+};
+
 const createChildWindow = async options => {
   const childWindows = await OpenfinApiHelpers.getChildWindows();
   const currentOptions = await OpenfinApiHelpers.getCurrentWindowOptions();
 
   options.uuid = currentOptions.uuid;
-  for (let i = 0; i < childWindows.length; i++) {
-    const childWindowOptions = await childWindows[i].getOptions();
-    if (childWindowOptions.name === options.name) {
-      if (childWindows[i]) {
-        childWindows[i].bringToFront();
-        return true;
+
+  if (options.allowMultiple) {
+    if (await windowAlreadyExist(options.name, childWindows)) {
+      let windowNumber = 1;
+      while (
+        await windowAlreadyExist(
+          `${options.name} ${windowNumber}`,
+          childWindows
+        )
+      ) {
+        windowNumber++;
       }
-      break;
+      options.name = `${options.name} ${windowNumber}`;
+    }
+  } else {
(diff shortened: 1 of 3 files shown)


packages/stockflux-core/src/services/ChildWindowLauncher.js
```diff
@@ -1,18 +1,44 @@
...
-    const childWindowOptions = await childWindows[i].getOptions();
-    if (childWindowOptions.name === options.name) {
-      if (childWindows[i]) {
-        childWindows[i].bringToFront();
-        return true;
+
+  if (options.allowMultiple) {
+    if (await windowAlreadyExist(options.name, childWindows)) {
+      let windowNumber = 1;
+      while (
```
kangelSL: Maybe I'm missing something, but could you not filter a set of `childWindows` for ones of the same name and then just get the length and use that as the index of the next window?

packages/stockflux-core/src/services/ChildWindowLauncher.js
```diff
@@ -1,18 +1,44 @@
...
 
   options.uuid = currentOptions.uuid;
-  for (let i = 0; i < childWindows.length; i++) {
-    const childWindowOptions = await childWindows[i].getOptions();
-    if (childWindowOptions.name === options.name) {
-      if (childWindows[i]) {
-        childWindows[i].bringToFront();
-        return true;
+
+  if (options.allowMultiple) {
```
kangelSL: Assuming there is no need to clarify `allowMultiple` on the other files because this will evaluate `undefined` as `false`, so it saves lines to just clarify the ones that are `true`?

SL-SWoods: That was my intent, that way we do not need to change any buttons/links/window launchers we do not wish to have multiple of.

kangelSL: Sounds good, just wanted to clarify

packages/stockflux-core/src/services/ChildWindowLauncher.js
SL-SWoods: My theory was to re-use number as the become available. for example when you first open a window the list of windows would be 
`[name]`
opening a second would be
`[name,name1]`
removing the first window, it would end up 
`[name1]`
If I used the length of the list to work out the last index, in this instance it would try and make `name1` again. The way used above it would use `name` and then if a 3rd window was added `name2` would be used.

kangelSL: Fair point, I hadn't considered that windows could be removed!

packages/stockflux-core/src/services/ChildWindowLauncher.js
```diff
@@ -1,18 +1,44 @@
 import * as OpenfinApiHelpers from '../openfin-api-utils/openfinApiHelpers';
 
+const windowAlreadyExist = async (windowName, childWindows) => {
```
AugusteBa: Maybe this should be another openfinApiHelper? I think I might have seen this used elsewhere - most likely in useChildWindow. There's no need to update that one as it will be replaced eventually by `openfin-react-hooks` `useChildWindow`. Yet I think many apps might benefit from this functionality in the future. Feel free to push back if you don't agree.

packages/stockflux-core/src/services/ChildWindowLauncher.js
```diff
@@ -1,18 +1,44 @@
...
+  for (let i = 0; i < childWindows.length; i++) {
```
AugusteBa: Maybe we should change this to `for...of` loop?

SL-SWoods: Changed to `for...of `.

packages/stockflux-core/src/services/ChildWindowLauncher.js
SL-SWoods: I think this will be a future PR if it's needed elsewhere. As there is an issue in the backlog to change `useChildWindow` to use the `openfin-react-hooks` version.

kangelSL:
Will this PR be needed after the future UX designs are included? Is there still a need for multiples of the same window if tabs will be included in the long run?

AugusteBa:
> Will this PR be needed after the future UX designs are included? Is there still a need for multiples of the same window if tabs will be included in the long run?

@SL-SWoods , could you find out if this is still needed (I remember Mark mentioning that this might no longer be required)?

SL-SWoods:
No longer required as changes to Spec

PR #1303, opened 2019-10-15
Feat: new standalone mode using custom setting

### LINKED ISSUE #1266: Launcher - Cloud mode

kangelSL:
PR fixes #1266

- Will work in tandem with PR on StockFlux Cloud (ScottLogic/StockFlux-Cloud#81)
- New custom data flags, initialDocked and cloudMode . initialDocked determines if app is started in standalone mode or in docked mode. cloudMode determines if the undock option should show at all.
- Changed styling to keep borders separate from item alignment, to work with the new undocked mode
- Disables undock and change dock direction arrows when another window is attached to the launcher

Standalone mode:
image

New unlock on docked feature:
image

### DIFF
--- a/packages/stockflux-launcher/src/App.js
+++ b/packages/stockflux-launcher/src/App.js
@@ -15,13 +15,13 @@ import './App.css';
 
 export default () => {
   const [alignment, setAlignment] = useState('horizontal');
-  const [isCloudMode, setCloudMode] = useState(false);
+  const [isDockable, setIsDockable] = useState(false);
   const [isDocked, setDocked] = useState(true);
   const [options] = useOptions();
   const [undockWidth, setUndockWidth] = useState(1000);
   const [undockHeight, setUndockHeight] = useState(50);
   const [undockTop] = useState(50);
-  const [undockLeft, setUndockLeft] = useState(0);
+  const [undockLeft] = useState(0);
 
   const [edge, windowActions] = useDockWindow(
     ScreenEdge.TOP,
@@ -35,16 +35,6 @@ export default () => {
   );
 
   let currentDirection = edge !== ScreenEdge.NONE ? edge : ScreenEdge.TOP;
-  let left =
-    window.screenLeft >= window.screen.availWidth ? window.screenLeft : 0;
-  let right =
-    window.screenLeft >= window.screen.availWidth
-      ? window.screenLeft
-      : window.screen.availWidth - 50;
-  let midpoint =
-    left === 0
-      ? (left + right) / 4
-      : window.screen.availWidth + (left + right) / 8;
 
   // Handle initialDocked false resize to mini window
   useEffect(() => {
@@ -53,10 +43,8 @@ export default () => {
       options.customData &&
       options.customData.initialDocked === false
     ) {
-      setUndockLeft(midpoint);
-      windowActions.dockNone();
+      undock();
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [options]);
 
   useEffect(() => {
@@ -66,41 +54,24 @@ export default () => {
   }, [edge]);
 
   useEffect(() => {
-    setCloudMode(
-      options && options.customData && options.customData.cloudMode === true
+    setIsDockable(
+      options && options.customData && options.customData.isDockable === true
     );
-  }, [options, setCloudMode]);
+  }, [options, setIsDockable]);
 
   useEffect(() => {
-    setDocked(!isCloudMode && edge !== ScreenEdge.NONE);
-  }, [edge, isCloudMode]);
+    setDocked(isDockable && edge !== ScreenEdge.NONE);
+  }, [edge, isDockable]);
 
   const undock = async () => {
-    switch (currentDirection) {
-      case ScreenEdge.LEFT:
-        setUndockHeight(600);
-        setUndockWidth(50);
-        setUndockLeft(left);
-        break;
-      case ScreenEdge.RIGHT:
-        setUndockHeight(600);
-        setUndockWidth(50);
-        setUndockLeft(right);
-        break;
-      case ScreenEdge.TOP:
-        setUndockHeight(50);
-        setUndockWidth(1000);
-        setUndockLeft(midpoint);
-        break;
-      default:
-        break;
-    }
-
+    setUndockHeight(50);
+    setUndockWidth(1000);
+    //setUndockLeft(window.screenLeft);
     windowActions.dockNone();
   };
 
   return (
-    <div className={cx('app', alignment, edge)}>
+    <div className={cx('app', currentDirection)}>
       {alignment === 'vertical' && CloseButton}
       <FreeTextSearch dockedTo={edge} />
       <ToolBar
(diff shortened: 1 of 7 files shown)


packages/stockflux-launcher/src/App.js
```diff
@@ -1,66 +1,144 @@
...
+  let left =
+    window.screenLeft >= window.screen.availWidth ? window.screenLeft : 0;
+  let right =
+    window.screenLeft >= window.screen.availWidth
+      ? window.screenLeft
+      : window.screen.availWidth - 50;
+  let midpoint =
+    left === 0
+      ? (left + right) / 4
+      : window.screen.availWidth + (left + right) / 8;
```
SL-SWoods: Spoke with @kangelSL and thinking about pulling a lot of the screen positioning maths out into separate functions. So a additional issue will be raised later to add this into said functions.

packages/stockflux-launcher/src/App.js
```diff
@@ -1,66 +1,144 @@
...
+        break;
+      case ScreenEdge.RIGHT:
+        setUndockHeight(600);
+        setUndockWidth(50);
+        setUndockLeft(right);
+        break;
+      case ScreenEdge.TOP:
+        setUndockHeight(50);
+        setUndockWidth(1000);
+        setUndockLeft(midpoint);
```
SL-SWoods: When undocking the screen from a secondary monitor the position of the hover launcher resets to the main display. 
This only occurs on secondary monitors which are to the negative direction of the primary display.

packages/stockflux-launcher/src/App.js
```diff
@@ -1,66 +1,144 @@
...
+    if (
+      options &&
+      options.customData &&
+      options.customData.initialDocked === false
+    ) {
+      setUndockLeft(midpoint);
+      windowActions.dockNone();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [options]);
```
AugusteBa: Are we sure this eslint-disable-next-line is needed?

kangelSL: I added it here to avoid the effect also watching the window, since I only want the function to fire once - when the options has loaded and been checked (i.e. to undock on startup). I will check if it can be removed.

packages/stockflux-launcher/src/free-text-search/FreeTextSearch.css
```diff
@@ -19,19 +19,20 @@
...
   justify-content: flex-start;
 }
 
-.left .free-text-search input,
-.right .free-text-search input {
+.vertical .free-text-search input {
   display: none;
 }
 
 div#root > .free-text-search {
```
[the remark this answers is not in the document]
kangelSL: Changed the file back to as it was before as the undocked mode will be kept in the horizontal position, so `fixed` has been removed. Tested removing the style and it is in use, it adds padding when the FreeTextSearch is docked left or right and makes sure the search button is in the right spot.

packages/stockflux-launcher/src/App.js
kangelSL: The undock should now work for all monitor setups, including 1 | 2, 2 | 1 and vertical alignments 1 / 2 and 2 / 1. The undock will move to the top left of that monitor and allow dragging

packages/stockflux-launcher/src/App.js
kangelSL: Clarified position of undocked; undocked will always be in horizontal position which made simplifying code easier. Moved position code to its own function

kangelSL:
WIP: Adding check for whether window has an attached window before allowing undock

kangelSL:
Related: discovered an error on new react hooks release, reported here: https://github.com/ScottLogic/openfin-react-hooks/issues/22

packages/stockflux-launcher/src/App.js
```diff
@@ -1,66 +1,140 @@
...
+  const [options] = useOptions();
+  const [undockHeight] = useState(50);
+  const [undockWidth] = useState(1000);
+  const [undockTop, setUndockTop] = useState(50);
+  const [undockLeft, setUndockLeft] = useState(0);
+
   const [edge, windowActions] = useDockWindow(
     ScreenEdge.TOP,
-    OpenfinApiHelpers.getCurrentWindowSync(),
+    window.fin.Window.getCurrentSync(),
```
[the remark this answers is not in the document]
kangelSL: Good catch; I must have overwritten this when adding the new hook options in

packages/stockflux-launcher/src/App.js
```diff
@@ -1,66 +1,148 @@
...
 
-const isLauncherHorizontal = edge => edge === ScreenEdge.TOP;
-
 export default () => {
+  const [isDockable, setIsDockable] = useState(false);
+  const [isDocked, setDocked] = useState(true);
+  const [isHorizontal, setHorizontal] = useState(true);
+  const [options] = useOptions();
+  const [undockHeight] = useState(50);
+  const [undockWidth] = useState(1000);
```
SL-SWoods: Do we need these 2 variables stored on the state? As they are never updated, and only referenced once, could we not just keep them as constants.

packages/stockflux-launcher/src/App.js
AugusteBa: This will not fire once - it should fire every time options change (which should not happen ofc). If you want this to happen only once loaded then an empty array should suffice. I'm happy to approve this, just something to consider.

packages/stockflux-launcher/src/App.js
```diff
@@ -1,66 +1,146 @@
...
-import AppShortcuts from './app-shortcuts/AppShortcuts';
 import FreeTextSearch from './free-text-search/FreeTextSearch';
 import ToolBar from './toolbar/ToolBar';
 import CloseButton from './toolbar/CloseButton';
 import './App.css';
 
-const isLauncherHorizontal = edge => edge === ScreenEdge.TOP;
-
 export default () => {
+  const [isDockable, setIsDockable] = useState(false);
```
AugusteBa: I think we should read from the manifest whether it's dockable or not rather than store another state variable.

packages/stockflux-launcher/src/App.js
```diff
@@ -1,66 +1,146 @@
...
+  const [isDocked, setDocked] = useState(true);
```
AugusteBa: isDocked might be redundant because we already are storing what edge is the launcher docked to (line 30). If it;s not docked it's `ScreenEdge.None`.

packages/stockflux-launcher/src/App.js
```diff
@@ -1,66 +1,146 @@
...
+  const [isHorizontal, setHorizontal] = useState(true);
+  const [options] = useOptions();
+  const [undockTop, setUndockTop] = useState(0);
+  const [undockLeft, setUndockLeft] = useState(0);
```
AugusteBa: I'm not sure what we are achieving with these variables. Why we are only storing `undockLeft`, but not `undiockRight`? Could this not be covered by looking up which edge the launcher is docked to (line 30)? If this is truly needed then I would suggest having a single state variable:
```
const [undock, setUndock] = useState(
  {
    left: 0,
    top: 0
  }
)
```

kangelSL: `undockLeft` is stored but not `undockRight` as `undockLeft` is used to set the position of the window when undocked. But agree that the state can be simplified into an `undockPosition` object, which should hopefully make its use more obvious at a glance.

kangelSL:
Changes overlapped with PR #1318, issue #1266 will be fixed there

Issue #1254, opened 2019-09-09
OpenFin modal for "Are you sure?" prompt


AugusteBa:

SL-SWoods:
Using the `useChildWindow` hook, a generic confirmation popup box has been created.
![ConfirmationDialog](https://user-images.githubusercontent.com/49903895/66119204-22b8b780-e5d0-11e9-8a4a-844283fa4571.png)
Two new components have been created, a generic `PopupWindow` and a more structured `ConfirmationWindow` these can be wrapped around existing objects to create popup windows when they are clicked. 

For example for the `CloseButton` component we have wrapped this in the `ConfirmationWindow` passing in appropriate methods. 
```Javascript
<ConfirmationWindow
    message="Are you sure you wish to quit?"
    onConfirm={() => OpenfinApiHelpers.getCurrentWindowSync().close()}
  >
    <Components.Buttons.Close />
  </ConfirmationWindow>
```

`ConfirmationWindow` has 3 available inputs. 
- `message` - string, message to be displayed to the user
- `onConfirm` - func, function to be executed on confirmation of the window
- `onReject` - func, function to be executed on rejection of the window. 

`PopupWindow` has a more expandedable options
- `message` - string, message to be displayed to the user 
- `options` - array, of the following json for each option to be displayed
```
{
      name: 'string',
      icon:  'string/component',
      className: 'string',
      action: 'func'
}
```

PR #1306, opened 2019-10-18
Generic confirmation popup

SL-SWoods:
In relation to #1254, new component added to add a confirmation popup to the close buttons. This can also be reused for other options in which confirmation is required. 
image
Can be wrapped around existing buttons/options to include the functionality. Confirmation boxes take `message` `onConfirm` `onReject` to control behaviour. and the more Generic `PopupWindow` can be passed an array of options it is allowed to work on.


packages/stockflux-launcher/src/toolbar/CloseButton.js
```diff
@@ -1,9 +1,13 @@
...
 import Components from 'stockflux-components';
 import { OpenfinApiHelpers } from 'stockflux-core';
+import ConfirmationWindow from '../popups/ConfirmationWindow';
 
 export default (
-  <Components.Buttons.Close
-    onClick={() => OpenfinApiHelpers.getCurrentWindowSync().close()}
-  />
+  <ConfirmationWindow
+    message="Are you sure you wish to quit?"
```
kangelSL: Can this be made into a prop so that the message could be adapted in future e.g. the message for shutting down the launcher should probably be different than the message for other components, as the launcher closes everything. 

I'm not sure if we need to ask for the user to confirm close on every component also (unless some data is lost), but that's just my preference.

SL-SWoods: This instance of the `CloseButton` is only used on the launcher close option. If you wanted to use a custom version of confirmation you would use `ConfirmationWindow` which has the property to have a custom message alongside with the `onConfirm`/`onReject` functions.

But I think the `ConfirmationWindow` and `PopupWindow` should be moved into the `stockflux-components` section, so they can be used in other sections.

packages/stockflux-launcher/public/popupWindow.css
```diff
@@ -0,0 +1,31 @@
+body {
+    overflow: hidden;
+    padding: 4vh 4vw;
+}
+.popup{
+    color:var(--text-grey-lighter);
```
kangelSL: Can the popup have a border similar to the one used on the launcher?

packages/stockflux-launcher/public/popupWindow.css
```diff
@@ -0,0 +1,31 @@
+body {
+    overflow: hidden;
+    padding: 4vh 4vw;
+}
+.popup{
+    color:var(--text-grey-lighter);    
+}
+
+.popup-message{
```
kangelSL: Could we try the buttons on the same line as the message?

packages/stockflux-launcher/public/popupWindow.css
```diff
@@ -0,0 +1,31 @@
...
+    color:var(--text-teal-light);
+    border-color:var(--text-teal-light);
+}
+
+.modal-cancel {
+    color:var(--text-teal-light);
+    border-color:var(--text-teal-light);
+}
+
+.modal-confirm{
```
kangelSL: This looks the same as the one above, could use `button, .modal-cancel, .modal-confirm {}` rather than three separate styles

kangelSL: This looks the same as the one above, could use `button, .modal-cancel, .modal-confirm {}` rather than three separate styles

packages/stockflux-launcher/src/popups/ConfirmationWindow.js
```diff
@@ -0,0 +1,28 @@
...
+      name: 'Cancel',
+      icon: <FaTimes></FaTimes>,
+      className: 'modal-cancel',
+      action: onReject
+    },
+    {
+      name: 'Confirm',
+      action: onConfirm,
+      icon: <FaCheck></FaCheck>,
+      className: 'modal-confirm'
```
kangelSL: Nitpick, but can the variables be kept in the same order for readability purposes?

packages/stockflux-launcher/src/popups/PopupWindow.js
```diff
@@ -0,0 +1,82 @@
+import React, { useState, useEffect, useCallback } from 'react';
+import Components from 'stockflux-components';
+import useChildWindow from '../search-results/helpers/useChildWindow';
+
+const ModalPopup = ({ message, options, children }) => {
```
AugusteBa: Filename does not match component name.

packages/stockflux-launcher/src/popups/PopupWindow.js
```diff
@@ -0,0 +1,82 @@
...
+      }
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [setShowPopup]
+  );
+
+  useEffect(() => {
+    if (childWindow.window) {
+      const childWindowJsx = (
+        <>
```
AugusteBa: This React.Fragment might be redundant.

packages/stockflux-launcher/src/popups/PopupWindow.js
SL-SWoods: Changed component name to match filename.

packages/stockflux-launcher/public/popupWindow.css
SL-SWoods: Removed both `.modal-cancel` and `.modal-confirm` they are not needed as they have the button attribute already.

packages/stockflux-launcher/src/popups/ConfirmationWindow.js
SL-SWoods: The order in which the items appear in the list are the same as they appear in the window. I think we want the OK button to be the bottom left icon with the cancel to the right of it, it might be a question for UX on order of buttons.

packages/stockflux-launcher/src/popups/PopupWindow.js
SL-SWoods: Removed now.

packages/stockflux-launcher/public/popupWindow.css
SL-SWoods: Added in a border to the CSS for popup windows.

packages/stockflux-launcher/public/popupWindow.css
SL-SWoods: image
Example of the message appearing on the same line. This might cause issues with longer string messages?

kangelSL: There's a lot of space on the left that could be used, if the text was aligned further to the left side? There should probably also be more space between the buttons and the text.

I don't think it looks significantly better than the way it was before, though, so feel free to change it back if you prefer. Mainly I was thinking we could reduce the amount of empty space at the bottom if it was kept on one line and not two

packages/stockflux-launcher/src/toolbar/CloseButton.js
```diff
@@ -1,9 +1,13 @@
 import React from 'react';
 import Components from 'stockflux-components';
 import { OpenfinApiHelpers } from 'stockflux-core';
+//import ConfirmationWindow from '../popups/ConfirmationWindow';
```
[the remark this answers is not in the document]
SL-SWoods: Removed the old comment.

packages/stockflux-components/src/index.js
```diff
@@ -21,5 +23,6 @@ export default {
   LargeSpinner,
   Buttons: { Round, Close },
   Shortcuts: { Watchlist, Chart, News },
+  Popups: { PopupWindow, ConfirmationWindow },
```
kangelSL: Nitpick; could be alphabetised?

PR #1315, opened 2019-10-22
UX

asasuSL:


AugusteBa:
Could we perhaps rename folders to match the rest of the project, e.g. `assets`, `watchlist-icon` or something similar?

asasuSL:
By all means, I’m happy to use whatever naming convention you already established.

Alex

PR #1316, opened 2019-10-23
Change title bar to new UX style

SL-SWoods:
- Load new colour information into master css file. 
- Modify Title bar to use new Icons
- Add parameter to `TitleBar` component allowing window name to be added 
  - `<Components.Titlebar title="string" />` can now be used to include the window name in any title bar. 
- Added new svg icons for close and minimise  
image

### DIFF
--- a/packages/stockflux-components/src/index.css
+++ b/packages/stockflux-components/src/index.css
@@ -30,11 +30,11 @@
   --color-primary:#0D2437;
   --color-secondary:#091A29;
   /* Windows*/
-  --background-color:#0D2437;  
-  --header-bar:#071521;
-  --divider:#091A29;
+  --window-background-color:#0D2437;  
+  --window-header-bar:#071521;
+  --window-divider:#091A29;
   --window-icon:#FFFFFF;
-  --stock-hover:#091A29;
+  --window-stock-hover:#091A29;
   /* Charts */
   --chart-line: #29E9E3;
   --chart-gradient-top:#29E9E3; 
@@ -50,6 +50,7 @@
   --icon-active:#29E9E3;
   --icon-inactive: #8A96A6;
   --icon-hover: #1B3B56;
+  --icon-hover-light: rgba(256,256,256,0.15);
   /* Text */
   --text-color:#FFFFFF; 
   --text-increase:#41D442;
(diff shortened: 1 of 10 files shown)


SL-SWoods:
Merged scrollbar changes into general UX feedback. 
- Remove hover to show scrollbar functionality
- change colours to use from `index.css` variables. 
image
Currently does not have the up/down arrows, need more research into webkit-scrollbar to get them to show correctly.

packages/stockflux-components/src/index.css
```diff
@@ -22,13 +22,45 @@
 
   --chart-teal: #37d8bc;
   --chart-teal-lighter: #51f2d6;
```
kangelSL: Does these need to move to go with the other `chart` items?

packages/stockflux-components/src/index.css
```diff
@@ -22,13 +22,45 @@
...
+  --color-inactive: #8A96A6;
+  --color-hover:#1B3B56;
+  --color-primary:#0D2437;
+  --color-secondary:#091A29;
+  /* Windows*/
+  --background-color:#0D2437;  
+  --header-bar:#071521;
+  --divider:#091A29;
+  --window-icon:#FFFFFF;
+  --stock-hover:#091A29;
```
kangelSL: These could be prefixed with `window` and then the comments would be unnecessary, since the names will all begin with something describing what they are?

packages/stockflux-components/src/index.css
SL-SWoods: Those are the older colours, which will be removed once the charts are updated and then reference the new colours only.
I kept them separate from the new section to make sure we can tell which is which when removing them later, or if the value gets changed to the new one we can move them down into the `UX Colours` section.

kangelSL: Apologies, you did mention that. Maybe the older ones could be grouped under an outmoded styles comment and then a separate ticket raised for removing them?

packages/stockflux-components/src/index.css
SL-SWoods: Prefixed with `window-` on all of the variables. updated any references to them as needed.

packages/stockflux-components/src/index.css
SL-SWoods: Moved old colours to end of file, with comment above that they are not to be used. When full UX work is completed they can be removed.

AugusteBa:
I think maybe the scrollbar should be an overlay one. Currently, when you hover over a card it looks funny as the background only partially changes (leaving the scrollbar area's background as it was). Also the divider lines don't span the whole width.
image
image

SL-SWoods:
PR is now part of #1324

PR #1318, opened 2019-10-23
1310 UX launcher refactor (inc. 1266 undocked changes)

### LINKED ISSUE #1310: Update launcher layout to match UX refactor v1.0.
**Depends on #1307 and #1309** 

Should match the launcher in UX refactor v1.0. Mainly:

1. Have a title bar with the app name "StockFlux"
1. Have input field matching the UX design (possibly in StockFlux components)
1. Have the search icon near the input field
1. Have a watchlist shortcut on the LHS.

(Any icons needed should be taken from the assets folder)

kangelSL:
Fixes for both #1310 and #1266 

- UX overhaul to resolve #1310 
- Will work in tandem with PR on StockFlux Cloud (ScottLogic/StockFlux-Cloud#81)
- New custom data flags, initialDocked and isDockable. initialDocked determines if app is started in undocked mode or in docked mode. isDockable determines if the undock option should show at all.
- Changed styling to keep borders separate from item alignment, to work with the new undocked mode
- Uses its own title bar to ensure the correct dimensions from UX designs are used

Current launcher undocked:
image

Docked:
image

image


packages/stockflux-components/src/styles/button-icon.css
```diff
@@ -12,7 +12,7 @@
     background-image: url('../images/minimize.png');
```
SL-SWoods: I think these images can be replaced by the svgs we have for the title bar. 
#1324 has the examples of the ones for the other windows title bar.

kangelSL: Added svgs in and left component file as before

kangelSL:
> I think the hover/disabled is the wrong way around. Looking at the spec when its active/docked the icon itself should be the teal colour. And i think the hover function needs the same as the minimise & close buttons (white background 15% opacity)

Good catch; I missed part of the icon styling on the spec; I'll adjust the icons and CSS accordingly.

kangelSL:
> Overall, looks better than before for sure. Just a couple of minor things below.
> 
> There's some overlapping when the launcher is vertical (see watchlist button at the top). Also search button and dock buttons are not centered compared to the spec:
> ...
> Is this intended? If there's a newer spec, could you create a small PR to keep dev up to date?

The vertical alignment should look like this:
image

The undock height has been fixed but there are other issues as you've mentioned so I'll address those; and there are some discrepancies between the version of the spec I have and the version on /dev are different so I'll get some clarification on these.

kangelSL:
> Also, hovering on a currently docked position button does not change the background of the button like it does on other dock buttons:
> E.g. hovering on `Top`:
> image
> E.g. hovering on `Left`:
> image
> Is this intentional?

Fixed to use right colour and cover disabled states also

kangelSL:
@AugusteBa 

Good points!

2. The font-family is Roboto as listed in the text, but good catch on the colour, it needs fixing to white. 
4. The hover colour is correct to the spec but is very subliminal; I'll get some clarification on that and what active & inactive states are
5. It is intended, but will ask about how active and inactive should behave as mentioned above 

Docked to the side
2. Yes, this is intentional. Back when I asked for clarification on the undocked state it was decided to keep the undocked window as horizontal 

Agree with the rest and will address them!

SL-SWoods:
There is a strange stutter in the drag functionality, when you drag the title bar away from the docked area, it seams to follow for a short while then jump back to the docked position. The drag continues and it will position itself on the release to where the mouse pointer is.
The stutter doesn't always snap it back to the top though, as sometimes it just stutters around your mouse pointer. Appears to happen on the top docked bar more than other docked options.

kangelSL:
> There is a strange stutter in the drag functionality, when you drag the title bar away from the docked area, it seams to follow for a short while then jump back to the docked position. The drag continues and it will position itself on the release to where the mouse pointer is.
> The stutter doesn't always snap it back to the top though, as sometimes it just stutters around your mouse pointer. Appears to happen on the top docked bar more than other docked options.

Able to replicate this more easily when search function is open; will raise a separate ticket for this and any further UX changes that come down the line.

packages/stockflux-launcher/src/App.css
```diff
@@ -1,38 +1,64 @@
 html {
-  background-color: var(--middle-background);
+  background-color: var(--launcher-color);
```
AugusteBa: Is this color different from the other apps? If not, i would suggest just using `--window-background-color` or at leat specifying 
```
:root {
  --launcher-color: var(--window-background-color)
}
```
That way, if we ever decide to change `--window-background-color` we can do it all in one place. 

I've noticed there are some more hardcoded colors in `:root`, e.g.`--titlebar-color`, `--search-input-text-color: #ffffff`. I think that is used throughout the apps and again, should not be declared/redeclared in the launcher app specifically. It's more global than the launcher. See `packages\stockflux-components\src\index.css`.

packages/stockflux-launcher/src/App.js
```diff
@@ -1,69 +1,112 @@
...
-const isLauncherHorizontal = edge => edge === ScreenEdge.TOP;
 
 export default () => {
+  const [options] = useOptions();
+  const defaultHeight = options ? options.defaultHeight : 75;
+  const defaultWidth = options ? options.defaultWidth : 50;
+  const isDockable = options ? options.customData.isDockable : false;
+  const [leftPosition, setLeft] = useState(0);
+  const [topPosition, setTop] = useState(0);
+  const [isHorizontal, setHorizontal] = useState(true);
```
AugusteBa: As I said a couple of reviews back, I think this is unnecessary. We already can check is it's Horizontal and this `useState` variable is just repeating the same logic. That info is stored in ScreenEdge and by checking it we can determine if it's horizontal or not.

Also, the name `isHorizontal` is less specific than checking screen edge and I think we should strive for specificity.

Instead, we can have a helper function called `isHorizontal` to check if the current position is `ScreenEdge.Top` or `ScreenEdge.None`.

Again, feel free to push back if you disagree.

packages/stockflux-launcher/src/App.js
```diff
@@ -1,69 +1,112 @@
...
-import './App.css';
-
-const isLauncherHorizontal = edge => edge === ScreenEdge.TOP;
 
 export default () => {
+  const [options] = useOptions();
+  const defaultHeight = options ? options.defaultHeight : 75;
+  const defaultWidth = options ? options.defaultWidth : 50;
+  const isDockable = options ? options.customData.isDockable : false;
+  const [leftPosition, setLeft] = useState(0);
```
AugusteBa: Maybe this could be renamed to `undockPosition`?

packages/stockflux-launcher/src/App.js
```diff
@@ -1,69 +1,112 @@
...
+      dockedHeight: defaultHeight
+    },
+    {
+      undockPosition: { left: leftPosition, top: topPosition },
+      undockSize: { width: 724, height: 88 }
+    }
   );
 
-  const prevEdgeRef = useRef();
+  const iconStyle = isHorizontal ? 'icon horizontal' : 'icon vertical';
```
AugusteBa: Just a suggestion for less repetition: 
`const iconStyle = `icon ${isHorizontal ? 'horizontal' : 'vertical'}`

packages/stockflux-launcher/src/App.js
```diff
@@ -1,69 +1,112 @@
...
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [edge]);
 
-  const edgeToBeChecked = edge === ScreenEdge.NONE ? prevEdge : edge;
+  useEffect(() => {
+    setHorizontal(
+      edge === ScreenEdge.TOP || edge === ScreenEdge.NONE ? true : false
+    );
+  }, [edge]);
```
AugusteBa: Again, I think it's a bit overkill having this in `useEffect`, it could just be a helper function that gets called only when we need to know if the launcher is horizontal or not rather than rerendering everytime `edge` form dependency array changes.

packages/stockflux-launcher/src/App.js
```diff
@@ -1,69 +1,112 @@
...
+        windowActions.dockNone();
+      });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [edge]);
 
-  const edgeToBeChecked = edge === ScreenEdge.NONE ? prevEdge : edge;
+  useEffect(() => {
+    setHorizontal(
+      edge === ScreenEdge.TOP || edge === ScreenEdge.NONE ? true : false
```
AugusteBa: Could be shortened by `[ScreenEdge.TOP, ScreenEdge.NONE].includes(edge)`;

packages/stockflux-launcher/src/App.js
```diff
@@ -1,69 +1,112 @@
...
-            className: 'drag-handle'
-          }
-        ]}
-      />
-      {isLauncherHorizontal(edgeToBeChecked) && CloseButton}
+    <div className="launcher-container">
+      <div className="launcher-title">
+        <Titlebar dockedTo={edge} />
+      </div>
+      <div className={cx('app', edge)}>
```
AugusteBa: Is using `cx` here is necessary?

packages/stockflux-launcher/src/helpers/getUndockedPosition.js
```diff
@@ -0,0 +1,102 @@
...
+  }
+}
+
+function getMiddleScreenPosition(screen) {
+  let amount = screen.left + screen.right;
+  let midpoint = amount / 2;
+  return midpoint - 362;
+}
+
+function getLeftPosition(
```
AugusteBa: Maybe could be renamed to `getUndockPosition`?

packages/stockflux-launcher/src/index.css
```diff
@@ -1,7 +1,14 @@
...
+  --icon-inactive: #8A96A6;
+  --icon-hover-light: rgb(27,59,86, 0.15);
+  --launcher-color: #0D2437;
   --search-input-min-width: 250px;
-  --search-input-width: 400px;
+  --search-input-text-color: #FFFFFF;
+  --search-input-width: 396px;
+  --search-input-width-undocked: 360px;
   --search-input-height: 30px;
+  --titlebar-color: #071521;
```
AugusteBa: Some of these colors are already defined `packages\stockflux-components\src\index.css` and are repeated here. We should avoid this and strive for color declarations being defined only once so that if color changes, we would only need to change it in one place.

packages/stockflux-launcher/src/titlebar/Titlebar.js
```diff
@@ -0,0 +1,68 @@
...
+    const currentWindow = await OpenfinApiHelpers.getCurrentWindow();
+    currentWindow.minimize();
+  }, []);
+
+  const onCloseClick = useCallback(async () => {
+    const currentWindow = await OpenfinApiHelpers.getCurrentWindow();
+    currentWindow.close();
+  }, []);
+
+  const dockedHorizontal = dockedTo === 'top' || dockedTo === 'none';
```
AugusteBa: Should we not use `ScreenEdge` here? Also could be simplified by `[ScreenEdge.None, ScreenEdge.Top].includes(dockedTo)`;

packages/stockflux-launcher/src/titlebar/Titlebar.js
```diff
@@ -0,0 +1,68 @@
...
+
+  return (
+    <div className="titlebar">
+      {dockedHorizontal && (
+        <div
+          className={cx(
+            'column column-left',
+            dockedTo === 'top' || dockedTo === 'none' ? 'add-padding-left' : ''
```
AugusteBa: Should we not use `ScreenEdge` here? Also could be simplified by `[ScreenEdge.None, ScreenEdge.Top].includes(dockedTo)`;

AugusteBa: It would definitely be worth having this as a helper function.

packages/stockflux-launcher/src/titlebar/Titlebar.js
```diff
@@ -0,0 +1,68 @@
...
+            <span className="title-strong">Stock</span>
+            <span className="title-light">Flux</span>
+          </p>
+        </div>
+      )}
+
+      <div
+        className={cx(
+          'column column-right header-icon',
+          dockedTo === 'top' || dockedTo === 'none' ? 'add-padding-right' : ''
```
AugusteBa: Again, same logic repetition.

packages/stockflux-launcher/src/titlebar/Titlebar.js
```diff
@@ -0,0 +1,68 @@
...
+        onClick={onCloseClick}
+        title="Close"
+      >
+        <Components.Icons.Toolbar.CloseIcon />
+      </div>
+
+      <div
+        className={cx(
+          'column column-right header-icon',
+          dockedTo === 'top' || dockedTo === 'none' ? 'add-padding-right' : ''
```
AugusteBa: Same here

packages/stockflux-launcher/src/titlebar/Titlebar.js
```diff
@@ -0,0 +1,68 @@
...
+        )}
+        onClick={onCloseClick}
+        title="Close"
+      >
+        <Components.Icons.Toolbar.CloseIcon />
+      </div>
+
+      <div
```
AugusteBa: These divs seem similar as well, could be combined into it's own component or a function that returns a JSX snippet.

packages/stockflux-launcher/src/titlebar/Titlebar.css
```diff
@@ -0,0 +1,76 @@
...
+    user-select:none;
+}
+
+.title-light {
+    font-family: 'Roboto';
+}
+
+.title-strong {
+    font-family: 'Roboto';
+    font-weight: 500;
```
AugusteBa: It should not be hardcoded, we have font-weights defined in `packages\stockflux-components\src\index.css`

packages/stockflux-launcher/src/titlebar/Titlebar.css
```diff
@@ -0,0 +1,76 @@
...
+.icons-container {
+    -webkit-app-region: no-drag;
+}
+
+.titlebar-container > .icons-container > .button-icon {
+    height: 15px;
+}
+
+.header-icon {
+    color: #FFFFFF;
```
AugusteBa: It should not be hardcoded, see `packages\stockflux-components\src\index.css`

packages/stockflux-launcher/src/titlebar/Titlebar.css
```diff
@@ -0,0 +1,76 @@
...
+    display: inline-block;    
+    cursor: pointer;
+    padding:8px;
```
AugusteBa: Space missing

packages/stockflux-launcher/src/toolbar/ToolBar.js
```diff
@@ -2,19 +2,21 @@ import React from 'react';
...
-      ))}
+    <div className={style}>
+      {tools
+        .filter(tool => tool.visible)
+        .map((tool, index) => (
+          <Components.Buttons.Round
+            key={index}
+            onClick={() => tool.onClick()}
+            disabled={!!tool.disabled}
+            className={tool.className ? tool.className : ''}
```
AugusteBa: Could we use `cx({[tool.className]: tool.className})`?

packages/stockflux-launcher/src/titlebar/Titlebar.js
AugusteBa: Also, I think this should use titlebar from `stockflux-components` rather than define it's own. The one in `stockflux-components` can be amended to accept JSX `<p><strong>Stock</strong>Flux</p>` as `children`.

packages/stockflux-launcher/src/titlebar/Titlebar.js
AugusteBa: Also, I think this should use `titlebar` from stockflux-components rather than define it's own. The one in stockflux-components can be amended to accept JSX <p><strong>Stock</strong>Flux</p> as children.

packages/stockflux-launcher/src/App.js
kangelSL: Discussed and raised #1351 to cover this as the edge/windowActions hook and dockedWidth is currently too tightly coupled so a state variable is necessary atm. 

Resolving anything `isHorizontal` related on this PR

packages/stockflux-launcher/src/App.js
SL-SWoods: Could it go to just `(edge === ScreenEdge.TOP || edge === ScreenEdge.NONE)` as well?

packages/stockflux-launcher/src/titlebar/Titlebar.js
kangelSL: Raised https://github.com/ScottLogic/StockFlux/issues/1352

packages/stockflux-launcher/src/App.js
kangelSL: Needed to evaluate to `app top`, `app none`, which share styles, or `app left` or `app right`, which share styles

packages/stockflux-launcher/src/titlebar/Titlebar.js
```diff
@@ -0,0 +1,67 @@
...
+        className={cx(
+          'column column-right header-icon',
+          dockedHorizontal ? 'add-padding-right' : ''
+        )}
+        onClick={onClick}
+        title={title}
+      >
+        {closeButton && <Components.Icons.Toolbar.CloseIcon />}
+        {!closeButton && <Components.Icons.Toolbar.MinimizeIcon />}
```
SL-SWoods: Not sure this is loading the correct svg, as both close and minimise icons are the same. 
image

packages/stockflux-launcher/src/App.js
```diff
@@ -1,69 +1,108 @@
...
-  FaChevronRight,
-  FaRegHandRock
-} from 'react-icons/fa';
-import AppShortcuts from './app-shortcuts/AppShortcuts';
+import './App.css';
 import FreeTextSearch from './free-text-search/FreeTextSearch';
+import getUndockedPosition from './helpers/getUndockedPosition';
+import { ReactComponent as LeftIcon } from './icons/left.svg';
+import { ReactComponent as RightIcon } from './icons/right.svg';
+import { ReactComponent as TopIcon } from './icons/top.svg';
```
AugusteBa: I think we should all agree on how we are importing `svgs`. Does importing it as `ReactComponent` help in any way?

packages/stockflux-launcher/src/helpers/getUndockedPosition.js
```diff
@@ -0,0 +1,102 @@
...
+
+async function getMonitorInfo() {
+  let monitorInfo = await window.fin.System.getMonitorInfo();
+  return monitorInfo;
+}
+
+function boundsCheck(bounds, screen, primaryCheck) {
+  if (primaryCheck) {
+    const lInBounds = bounds.left >= screen.left && bounds.left < screen.right;
+    const tInBounds = bounds.top >= screen.top && bounds.top < screen.bottom;
```
AugusteBa: Do these mean LeftInBounds, TopInBounds, RightInBounds?

kangelSL: Updated variable names for clarity

packages/stockflux-launcher/src/titlebar/Titlebar.js
kangelSL: Good catch; this is a typo on the import in the `components` folder that I missed

packages/stockflux-launcher/src/App.js
kangelSL: The main reason it's imported this way is that I was getting a React error when I didn't import it as a ReactComponent, as mentioned in this thread as a common issue: https://github.com/facebook/create-react-app/issues/5293 
Agree it should be standardised though so this should be brought in line with the others

packages/stockflux-launcher/src/helpers/getUndockedPosition.js
```diff
@@ -0,0 +1,103 @@
...
+
+async function getBounds() {
+  let bounds = await OpenfinApiHelpers.getCurrentWindowSync().getBounds();
+  return bounds;
+}
+
+async function getMonitorInfo() {
+  let monitorInfo = await window.fin.System.getMonitorInfo();
+  return monitorInfo;
+}
```
SL-SWoods: Can we move this into `OpenfinApiHelpers` as that is the only other place within the system which references `window.fin`.

packages/stockflux-launcher/src/App.js
kangelSL: Due to complexity of this issue, I have moved the icons to the components folder and will raise a separate ticket if we are keeping all the icons separate in their packages in future. 

Importing the images any other way either requires using a messy import, e.g.
`// eslint-disable-next-line import/no-webpack-loader-syntax
import LeftIcon from '-!react-svg-loader!./icons/left.svg';`

Or adding in `react-svg-loader` and rebuilding the entire package using webpack, which I feel is out of scope of this PR solely for the inclusion of the launcher icons

PR #1324, opened 2019-10-28
UX Watchlist changes

SL-SWoods:
image
Apply UX changes to Watch list, includes changes for loading the svg images into the components. Includes CSS changes to give access to new colour information etc to all other components.


SL-SWoods:
This PR also includes the changes as part of #1316 - Title bar and scroll bar UX changes.

AugusteBa:
The changes look very nice!

I'm just confused about hover functionality, it does not seem to match the spec on dev branch. 
image

Shouldn't hover active state be blue outline without any background color change?
image

Is there a more up to date spec, which is not published on dev?

packages/stockflux-components/src/components/buttons/borderless-buttons/app-shortcuts/Chart.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Launchers } from 'stockflux-core';
+// import ChartIcon from '../../../glyphs/small/chart.svg';
```
AugusteBa: Looks like a leftover comment

AugusteBa:
Should glyphs not come from `assets` folder or is it causing problems not being in the `src` directory? I think it would be nice if we could keep it all in one place, if it's not causing too much headache. If it is, then maybe we should consider moving all assets in `stockflux-components/src`?

SL-SWoods:
> The changes look very nice!
> 
> I'm just confused about hover functionality, it does not seem to match the spec on dev branch.
> image
> 
> Shouldn't hover active state be blue outline without any background color change?
> image
> 
> Is there a more up to date spec, which is not published on dev?
> 
> Also dragging a card out when clicked on card-top does not delete the card, but shows preview. Is this intended?

On the hover states, none of the icons currently have an active on them, so they are acting like the top option with the white background. I don't know if we are going with the active flag shows if the window is already open. Unless I have read the spec wrong and they teal is for the hover?

Drag currently has kept with the default of wherever you drag it just opens the chart. I have a separate branch for investigating if / how the different drag events can be handled dependent on where you click.

packages/stockflux-components/src/components/buttons/borderless-buttons/app-shortcuts/Chart.js
SL-SWoods: Removed now.

packages/stockflux-components/src/components/glyphs/arrows/priceArrowDown.svg
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="9px" height="7px" viewBox="0 0 9 7" version="1.1" 
+    xmlns="http://www.w3.org/2000/svg" 
+    xmlns:xlink="http://www.w3.org/1999/xlink">        
+    <g id="Assets" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <path fill="currentColor" d="M4.5,0 L9,7 L0,7 L4.5,0 Z" transform="translate(4.500000, 3.500000) scale(1, -1) translate(-4.500000, -3.500000) "></path>
```
kangelSL: Nitpick; does there need to be a space at the end of the `transform` variable?

packages/stockflux-components/src/components/icons/minimize.svg
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="12px" height="2px" viewBox="0 0 12 2" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 59 (86127) - https://sketch.com -->
```
kangelSL: Nitpick; these comments could be removed from `minimize`, `link`, `close` and `watchlist` to keep them consistent with the other cut down files?

packages/stockflux-components/src/components/icons/minimize.svg
```diff
@@ -0,0 +1,14 @@
...
+    <title>Minimize</title>
+    <desc>Created with Sketch.</desc>
```
kangelSL: Same with these `<desc>` tags?

SL-SWoods:
> The divider line on the spec looks to be black but is white here; this might need clarification as to which colour it should be?

Just checked with Sarah, and have been given the hex code for the dividing line.

packages/stockflux-watchlist/src/components/watchlist-card/WatchlistCard.js
```diff
@@ -162,25 +171,36 @@ const WatchlistCard = ({
             />
             <div className="details">
               {<div className="price">{stockData.price || 'N/A'}</div>}
-              {<div className="delta">{stockData.delta || ''}</div>}
-
-              <div className="percentage">
+              <div
+                className={classNames({
+                  'percentage price_positive': stockData.percentage > 0,
+                  'percentage price_negative': stockData.percentage < 0
```
kangelSL: Could use a ternary instead? `stockData.percentage > 0 ? 'percentage price_positive' : 'percentage price_negative'
Or classNames cx 'percentage', then the ternary statement for adding price_positive or price_negative

packages/stockflux-components/src/components/glyphs/arrows/priceArrowDown.svg
SL-SWoods: removed and still works correctly.

packages/stockflux-components/src/components/icons/minimize.svg
SL-SWoods: Removed comments and bloat from all svgs so far.

packages/stockflux-components/src/components/icons/minimize.svg
SL-SWoods: Removed comments and bloat from all svgs so far.

packages/stockflux-watchlist/src/components/watchlist-card/WatchlistCard.js
```diff
@@ -162,25 +171,36 @@ const WatchlistCard = ({
...
+                  </span>
                 ) : (
                   ''
                 )}
               </div>
+              {
+                <div
+                  className={classNames({
+                    'delta price_positive': stockData.percentage > 0,
+                    'delta price_negative': stockData.percentage < 0
```
kangelSL: Same here as above

SL-SWoods:
> There seems to be overlap on watchlist items with longer names or bigger numbers:
> 
> image
> 
> The eye symbol is also covered by the scrollbar

Fixed the eye issue my making sure it cannot be resized too small, I had missed the min-width when i changed default/opening width. 
I can see the issue with the price overlaying the chart, seams to happen now when you get more than 7 digits in the price. Will investigate if its possible to make it fit for larger numbers. as i believe we should try and make it fit up to 999,999.99 as the highest stock price in the world is that many digits.

packages/stockflux-components/src/components/buttons/borderless-buttons/Close.js
```diff
@@ -0,0 +1,13 @@
...
+import { FaTimes } from 'react-icons/fa';
+import cx from 'classnames';
+import BorderlessButton from '../borderless-button/BorderlessButton';
+
+export default props => (
+  <BorderlessButton
+    {...props}
+    className={cx('close', { [props.className]: props.className })}
+  >
+    <FaTimes />
```
kangelSL: Which symbol is this using? If it closes a window should it be a cross symbol?

SL-SWoods: The times symbol is used for close as it gives the normal "x" icon. the cross icon from FA gives the christian cross.
image
When using FA to look for icons, the times icon is the one presented when looking for close as well. [https://fontawesome.com/icons?d=gallery&q=Close](https://fontawesome.com/icons?d=gallery&q=Close)

packages/stockflux-components/src/components/Titlebar/Titlebar.js
```diff
@@ -24,25 +27,24 @@ export default () => {
...
+      </div>
       <div className="icons-container">
-        <div
-          className="button-icon minimize"
-          onClick={onMinimizeClick}
-          title="Minimize"
-        >
-          &nbsp;
-        </div>
         {isDocked[0] && (
```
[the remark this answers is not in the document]
AugusteBa: Maybe the link icons should be a separate PR?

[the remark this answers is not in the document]
SL-SWoods: This appears to be happening on every branch. As the return from the hook is always returning false, believe this is a openfin issue as it was working before and even on branches with no changes this is no longer working.
Raised into the document of openfin issue to be reviewed.

packages/stockflux-watchlist/src/components/watchlist-card/WatchlistCard.js
SL-SWoods: removed the classNames function and used a normal ternary instead, makes code a lot more readable.

packages/stockflux-components/src/components/buttons/borderless-button/BorderlessButton.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import cx from 'classnames';
+import './BorderlessButton.css';
+
+export default ({ children, small, className, onClick, disabled }) => (
+  <button
+    className={cx('borderless', { small, [className]: className })}
```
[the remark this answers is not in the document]
SL-SWoods: Changed, the old roundButton still has this method in, but this will be removed eventually.

PR #1325, opened 2019-10-28
1319: useNotification

AugusteBa:
Refers to #1319 
We will need to make a tiny change once my PR gets merged into `openfin-react-hooks`, but for now - this works.


packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
```diff
@@ -36,17 +43,16 @@ const Watchlist = () => {
...
-        watchlistName: 'My Watchlist',
-        alreadyInWatchlist,
-        messageText: `${alreadyInWatchlist ? ' moved' : ' added'} to the top`
-      }
-    });
+    setNewSymbol(newSymbol);
+    if (notification) {
+      notification.launch({
+        url: 'notification.html',
+        timeout: 'never'
```
kangelSL: The notification is currently appearing and never vanishing; should it be made to disappear after a programmed interval? Currently the user needs to drag it away to make it disappear which is not immediately obvious; is there a way to add an 'X' to show the user how they can close it?

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
```diff
@@ -59,6 +65,33 @@ const Watchlist = () => {
...
+            />
+          </div>
+          <div id="notification-content">
+            <p id="title">WATCHLIST UPDATED</p>
+            <hr />
+            <p id="info">
+              <span id="symbol">{newSymbol}</span>
+              <span id="message">{`${
+                alreadyInWatchlist ? ' moved' : ' added'
+              } to the top`}</span>
```
kangelSL: If I add something new to the watchlist, I get the message 'moved to the top', even if the list is empty. Should it be 'added to the top'? Also when I add the same item twice the notification fires twice. Is this the correct behaviour?

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
AugusteBa: Thanks, Katriona, that's a very good point, removed it now altogether, which defaults to 3600 milliseconds. 
There is a way to add X button, but there's no such requirement in UX specs.

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
AugusteBa: Yeah, that is expected behavior (firing twice). Messages are going to change either way in the Notification UX refactor PR.

Issue #1313, opened 2019-10-18
Remove `useChildWindow` instance in`stockflux-components` and replace it with `openfin-react-hooks` one.


AugusteBa:
Ensure all StockFlux suite apps work with the new import.

PR #1329, opened 2019-10-28
1313: useChildWindow

AugusteBa:
This refers to #1313. 

This PR replaces all usages of inbuilt `useChildWindow` with `useChildWindow` from `openfin-react-hooks`.


packages/stockflux-components/package.json
```diff
@@ -1,32 +1,33 @@
...
+    "peerDependencies": {
+        "react": "^16.8.6",
+        "openfin-react-hooks": "^1.3.0"
+    },
+    "scripts": {
+        "build": "webpack"
+    },
+    "dependencies": {
+        "classnames": "^2.2.6",
+        "openfin-react-hooks": "^1.3.0",
```
kangelSL: Has the spacing changed on the file? It looks as though the `openfin-react-hooks` are the only lines that have changed?

AugusteBa: Yep, thanks for noticing this. Should be fixed now.

PR #1356, opened 2019-11-01
News: Change layout to new ux standard

SL-SWoods:
image

Update news to fit in with new UX standards.
- Colours updated to new standards
- Title bar matches watch list title bar.
- Placeholder for news 'tabs'.

### DIFF
--- a/packages/stockflux-components/src/index.css
+++ b/packages/stockflux-components/src/index.css
@@ -61,7 +61,7 @@
 
   /* padding */
   --padding-small: 5px;
-  --padding-medium: 10px;
+  --padding-medium: 12px;
   --padding-large: 15px;
   --padding-xlarge: 20px;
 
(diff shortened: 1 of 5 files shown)


packages/stockflux-components/src/index.css
```diff
@@ -61,7 +61,7 @@
 
   /* padding */
   --padding-small: 5px;
-  --padding-medium: 10px;
+  --padding-medium: 12px;
```
AugusteBa: Does this not break any other components?

SL-SWoods: I have checked through the other references where this is used, and it makes no difference to them.
- One is for when the mini-chart errors. so the change just aligns the text to better fit the section
- The other is in the padding of watch list items to the left, so now it matches the scrollbar width making it centred.

PR #1359, opened 2019-11-04
New UX changes

AugusteBa:


AugusteBa:
> Do we have a standard for the file names? It currently looks like lowercase with hyphens for specs and usual names with spaces for ref files? Currently `pop up Ref` and and `Popup_Specs` are the odd ones out given that the capitalisation/conventions are mixed.

@kangelSL, good point. I just had a call with @sarahlallemandsl and we discussed how we should tidy up assets folder in general. This included renaming some of the files to make it more consistent. There should be a commit soon that fixes it.

PR #1363, opened 2019-11-05
add confirmClose option to titlebar

SL-SWoods:
Add a new parameter to the title bar options which will allow for close confirmation to be applied. 
`<Components.Titlebarr confirmClose={bool}>` can be used to toggle a confirmation dialog on close of a window. If the parameter is omitted or false no dialog is shown and windows close automatically. 

In order for popups to work correctly, the `.html` and `.css` files for the popup need to be included in the project which this is being used.

### DIFF
--- a/packages/stockflux-components/src/components/Titlebar/Titlebar.js
+++ b/packages/stockflux-components/src/components/Titlebar/Titlebar.js
@@ -5,10 +5,11 @@ import { OpenfinApiHelpers } from 'stockflux-core';
 import CloseIcon from '../icons/close.svg';
 import MinimizeIcon from '../icons/minimize.svg';
 import LinkIcon from '../icons/link.svg';
+import ConfirmationWindow from '../popups/ConfirmationWindow';
 
 import './Titlebar.css';
 
-export default ({ title }) => {
+export default ({ title, confirmClose }) => {
   const isDocked = useDocked();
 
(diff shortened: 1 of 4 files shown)


SL-SWoods:
This change is needed for #1358 to work correctly.

packages/stockflux-components/src/components/Titlebar/Titlebar.js
```diff
@@ -43,9 +44,21 @@ export default ({ title }) => {
...
-        </div>
+        {!confirmClose && (
+          <div className="header-icon" onClick={onCloseClick} title="Close">
+            <CloseIcon />
+          </div>
+        )}
+        {confirmClose && (
+          <div className="header-icon">
+            <ConfirmationWindow
+              message="This will close all related windows. Do you wish to continue?"
```
kangelSL: Should this be `this will close the current window` instead? Or is it `all related` as in any child windows spawned specifically from this window?

SL-SWoods: `all related` is needed as it would also close any children spawned from the window in question. 
I suspect the wording will change when UX have looked at the confirmation boxes.

kangelSL: That's what I thought; thanks for the clarification.

PR #1364, opened 2019-11-05
Watchlist - Change drag functionality based on drag start location

SL-SWoods:
As per the UX design, change the watch list so that
- the top dark bar when dragged outside the window will remove the entry from watch list
- the chart lighter section when dragged outside will show the charts. 
- the relevant preview window is shown dependent on the drag being performed. 
DragTopBar
DragBottomBar

Basic preview windows are being used for each drag out option currently.


packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
```diff
@@ -24,6 +24,7 @@ const Watchlist = () => {
   const [dragOverIndex, setDragOverIndex] = useState(null);
   const [unwatchedSymbol, setUnwatchedSymbol] = useState(null);
   const [displayPreview, setDisplayPreview] = useState(false);
+  const [displayDelete, setDisplayDelete] = useState(false);
```
kangelSL: Would this be any simpler if it was collapsed into one variable? Maybe `displayWindow` and then `setDisplayMode`?
So there could be three states - false to display nothing, a constant for chart state and a constant for delete state?

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
```diff
@@ -103,12 +104,18 @@ const Watchlist = () => {
     }
   };
 
+  const setPreviewMode = mode => {
+    mode === 'CHART' && setDisplayPreview(true);
+    mode === 'DELETE' && setDisplayDelete(true);
```
kangelSL: From the above, this could be a ternary:
`mode === 'CHART'? setDisplayWindow(SOME_CONSTANT_CHART) : setDisplayWindow(SOME_CONSTANT_DELETE)`?

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
```diff
@@ -191,17 +198,18 @@ const Watchlist = () => {
...
         onDragOver(e, watchlist, dragOverIndex, setDragOverIndex);
       }}
       onDragEnd={() => {
         setDisplayPreview(false);
+        setDisplayDelete(false);
         resetDragState(setDragOverIndex);
       }}
       onDrop={e => {
         setDisplayPreview(false);
+        setDisplayDelete(false);
```
kangelSL: And here there would only need to be one setDisplayX as false instead of two

packages/stockflux-watchlist/src/components/watchlist-card/WatchlistCard.js
```diff
@@ -179,7 +194,11 @@ const WatchlistCard = ({
           </div>
           <div
             className="card-bottom"
-            onClick={() => bindings.onDropOutside(symbol, stockData.name)}
+            onClick={() => {
+              setDragOutcome('');
+              bindings.onDropOutside(symbol, stockData.name);
+            }}
+            onMouseDown={() => setDragOutcome('CHART')}
```
kangelSL: Asking for clarification, the `setDragOutcome` is done twice to set it back to the default value?

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
SL-SWoods: Changed due to refactor.

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
SL-SWoods: That's a good idea. refactored code to have it use a single variable now,

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
SL-SWoods: Changed to single variable method.

packages/stockflux-watchlist/src/components/watchlist-card/WatchlistCard.js
SL-SWoods: It should not be done twice, removed this instance of it as it will be dealt with `setDragOutside`.

PR #1367, opened 2019-11-06
Preview Window - Update to match UX designs

SL-SWoods:
Update the chart preview window to match UX designs.
image

### DIFF
--- a/packages/stockflux-watchlist/public/preview-chart.html
+++ b/packages/stockflux-watchlist/public/preview-chart.html
@@ -12,9 +12,9 @@
         box-sizing: border-box;
         width: 100vw;
         height: 100vh;
-        background-color:#1B3B56;
-        color:white;
-        font-weight: 20px;
+        background-color:var(--window-background-color);
+        color:var(--text-color);
+        font-weight: 20px;        
       }
       #root {        
         width:inherit;
@@ -27,7 +27,7 @@
         flex-direction: column;
         justify-content: center;
         align-items: center;
-        box-shadow: 0px 2px 4px 1px rgba(13, 36, 55, 1);
+        box-shadow: 0px 2px 9px 3px rgba(27, 59, 86, 1);        
       }
     </style>
   </head>
--- a/packages/stockflux-watchlist/public/preview-chart.html
+++ b/packages/stockflux-watchlist/public/preview-chart.html
@@ -12,12 +12,12 @@
         box-sizing: border-box;
         width: 100vw;
         height: 100vh;
-        background-color:var(--window-background-color);
-        color:var(--text-color);
+        background-color: var(--window-background-color);
+        color: var(--text-color);
         font-weight: 20px;        
       }
       #root {        
-        width:inherit;
+        width: inherit;
         height: inherit;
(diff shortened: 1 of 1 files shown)


packages/stockflux-watchlist/public/preview-chart.html
```diff
@@ -12,6 +12,9 @@
         box-sizing: border-box;
         width: 100vw;
         height: 100vh;
+        background-color:#1B3B56;
```
kangelSL: Assuming it's not possible to make this into a `var` since it's a public/.html file?

packages/stockflux-watchlist/public/preview-chart.html
```diff
@@ -12,6 +12,9 @@
...
+        color:white;
```
kangelSL: Nitpick; shouldn't there be a space between the colon and the colour?

Also, can this be a hex value instead of a name to keep things consistent?

packages/stockflux-watchlist/public/preview-chart.html
```diff
@@ -21,20 +24,22 @@
         display: flex;
         width: 100%;
         height: 100%;
-        box-sizing: border-box;
-        border: dotted 3px #fff;
-        overflow: hidden;
+        flex-direction: column;
         justify-content: center;
         align-items: center;
+        box-shadow: 0px 2px 4px 1px rgba(13, 36, 55, 1);
```
kangelSL: Is this correct to the spec? The shadow values there are given as blur = 9 and spread = 3, so should it be
`box-shadow: 0px 2px 9px 3px` ?

Also, the spec lists the colour as #1B3B56, which I think is 27, 59, 86 in rgb?

packages/stockflux-watchlist/public/preview-chart.html
SL-SWoods: That's what i thought initially, but upon changing them to use the var from the main index file it could reference them, so changed to use the var values.

packages/stockflux-watchlist/public/preview-chart.html
SL-SWoods: Changed the drop shadow values.

packages/stockflux-watchlist/public/preview-chart.html
SL-SWoods: no longer needed as using the var values from index.css

packages/stockflux-watchlist/public/preview-chart.html
```diff
@@ -12,6 +12,9 @@
         box-sizing: border-box;
         width: 100vw;
         height: 100vh;
+        background-color:var(--window-background-color);
+        color:var(--text-color);
```
kangelSL: Nitpicking again; can we put a space here for readability?

packages/stockflux-watchlist/public/preview-chart.html
AugusteBa: Is this not using `useChildWindow`? This would allow CSS vars to actually be rendered to appropriate values.

PR #1368, opened 2019-11-06
1362: Notification UX Refactor v1.0

AugusteBa:
notification17


packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
```diff
@@ -66,36 +68,22 @@ const Watchlist = () => {
...
-            </p>
-          </div>
-        </>
+        <div className="notification">
+          <div className="icon"></div>
+          <p>
+            {newSymbol.symbol}
+            {newSymbol.isNew
+              ? ' has been added to your Watchlist'
+              : ' has been moved to the top of your Watchlist'}
```
SL-SWoods: I wonder if its worth pulling out every instance where we have text like this into a single place? so that way if text ever needs changing it is only in a single place for every place?
Not part of this PR, but one for the backlog, as it could be useful for text like "no search results" etc.

AugusteBa: Yeah, a very good idea, that would be very nice indeed!

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
```diff
@@ -45,18 +45,20 @@ const Watchlist = () => {
   const [options] = useOptions();
   const notification = useNotification(NOTIFICATION_OPTIONS);
 
-  const displayNotification = newSymbol => {
-    setNewSymbol(newSymbol);
+  const displayNotification = async () => {
     if (notification) {
-      notification.launch({
+      await notification.launch({
```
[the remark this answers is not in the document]
AugusteBa: Oh yeah, it is indeed necessary. Removed now, thanks for spotting this!

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
```diff
@@ -0,0 +1,262 @@
+import React, { useState, useEffect } from 'react';
```
SL-SWoods: Should this file still be in the PR? it doesn't get referenced anymore, due to the refactor into app.js

packages/stockflux-watchlist/src/components/watchlist/Watchlist.css
```diff
@@ -0,0 +1,103 @@
+* {
```
SL-SWoods: Should this file still be in the PR? it doesn't get referenced anymore, due to the refactor into app.js

packages/stockflux-watchlist/src/components/watchlist/Watchlist.js
AugusteBa: Yeah, they can go! Thanks!

PR #1369, opened 2019-11-06
UX refactor dialog box

SL-SWoods:
Update confirmation dialog box to keep with new UX designs
image
- Remove border
- Change icons to words
- Larger buttons for selection options
- Centre dialog text accordingly. 

Unable to match specs on 
- Rounded border of the window. 
Need to confirm colours, as they do not appear to match the main colours being used within Stockflux.

### DIFF
--- a/packages/stockflux-components/src/components/buttons/dialog-button/DialogButton.css
+++ b/packages/stockflux-components/src/components/buttons/dialog-button/DialogButton.css
@@ -1,0 +1,43 @@
+.dialog {
+  display: inline-block;
+  color: var(--text-color);
+  border: none;
+  outline: none;
+  background-color: inherit;
+  cursor: pointer;
+}
+
+.dialog.small {
+  height: 30px;
+  width: 30px;
+  padding: 5px;
+  margin: 2px;
+}
+
+.dialog svg {
+  margin: auto;
+  display: block;
+  shape-rendering: auto;
+  fill: currentColor;
+}
+
+.dialog svg:hover {
+  color: inherit;
+}
+
+.dialog:hover {
+  background-color: var(--color-hover);
+}
+
+.dialog:active {
+  background-color: var(--color-highlight);
+}
+
+.dialog:disabled {
+  color: var(--color-inactive);
+  cursor: default;
+}
+
+.dialog:disabled:hover {
+  background-color: inherit;
+}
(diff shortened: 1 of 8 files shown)


packages/stockflux-components/src/components/popups/PopupWindow.js
```diff
@@ -26,11 +26,10 @@ const PopupWindow = ({ message, options, children }) => {
       saveWindowState: false,
       showTaskbarIcon: false,
       resizable: false,
-      backgroundColor: '#28313D',
       waitForPageLoad: true,
       alwaysOnTop: true,
-      defaultWidth: 300,
-      defaultHeight: 100
+      defaultWidth: 270,
```
kangelSL: Should this be 275? The spec width looks like it's 50px + 50px + 175px = 275px

packages/stockflux-components/src/components/buttons/dialog-button/DialogButton.css
```diff
@@ -0,0 +1,43 @@
...
+  height: 30px;
+  width: 30px;
+  padding: 5px;
+  margin: 2px;
+}
+
+.dialog svg {
+  margin: auto;
+  display: block;
+  shape-rendering: auto;
```
kangelSL: If the default value is `auto`, is this needed?

packages/stockflux-launcher/public/popupWindow.css
```diff
@@ -1,23 +1,40 @@
+html {
+    border-radius:20px;
```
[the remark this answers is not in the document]
SL-SWoods: should have been 5px, it was 20px when i was testing to see if radius was working correctly.

packages/stockflux-components/src/components/popups/PopupWindow.js
SL-SWoods: the specs say 270x145, but if you look at the maths for the individual sections it adds up to 275. unsure which is the correct ones will confirm with UX.

packages/stockflux-components/src/components/buttons/dialog-button/DialogButton.css
SL-SWoods: that entire section of css was not needed as we have no svgs within the dialogbuttons, so i have removed it.

packages/stockflux-components/src/components/popups/PopupWindow.js
SL-SWoods: Confirmed the width should be 270.

PR #1373, opened 2019-11-07
Watchlist : Refactor out of app.js

SL-SWoods:
Refactor watch list app.js to pull titlebar component into the Watchlist.


packages/stockflux-watchlist/src/App.js
```diff
@@ -1,15 +1,9 @@
...
 import './App.css';
 
 const App = () => {
-  return (
-    <>
-      <Components.Titlebar title="Watchlist" />
-      <Watchlist />
-    </>
-  );
+  return <Watchlist />;
```
AugusteBa: Maybe we should move logic from Watchlist.js into this file? For it to be similar to other apps.

SL-SWoods: Changed the direction of the refactor as discussed.

Issue #1376, opened 2019-11-08
Watchlist: open window highlight doesn't work if chart opened via drag. 


SL-SWoods:
The open window highlight on the chart icon does not show if the chart is opened from dragging it out of the window or when clicking the chart graphic. 
If the watch list is them refreshed at any point the icon shows as expected. 
Suspect useEffect is missing a dependency etc.

Issue #1378, opened 2019-11-08
Watchlist: Opening chart from button click is in incorrect position


SL-SWoods:
When opening the chart from the new chart button, the resulting window does not load into the same location as when dragged or clicking on the chart. 
- Change so all windows open in the same correct position.

gsmith-scottlogic:
Now behaving correctly

PR #1379, opened 2019-11-08
Watchlist: Chart opens in incorrect position when button is used. 

SL-SWoods:
Related to #1378 Fix now charts are loaded from watch list. If the chart was launched from the button it was appearing out of place.
This is now fixed so they all open in the same place but has caused issue #1376 to happen on every open of the chart. Suspect this is due to the window not open before the check for if its open is performed.


SL-SWoods:
Fixed issue #1376 as well so all methods of open chart from the watch list will cause the "isOpen" highlight to show.

SL-SWoods:
No longer required.

PR #1388, opened 2019-11-15
add initial unit test for watch list Reducer

SL-SWoods:
Add unit test for reducers within stockflux-watchlist.

New commands added to `package.json`
`npm run watchlist:test` - run tests for watchlist.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "chart:launch-appd": "cd packages/stockflux-chart && npm run launch-appd",
     "launcher:launch-appd": "cd packages/stockflux-launcher && npm run launch-appd",
     "news:launch-appd": "cd packages/stockflux-news && npm run launch-appd",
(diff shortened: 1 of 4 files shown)


packages/stockflux-launcher/src/reducers/free-text-search/FreeTextSearch.test.js
```diff
@@ -0,0 +1,58 @@
+import reducer, { initialSearchState } from './FreeTextSearch';
+import action from './Action';
+
+describe('Action Reducer', () => {
+  it('Should return inital state when given an unknown action', () => {
```
markjose: I'd imagine this should throw, unknown actions are a no no, especially if you're using constants for the action names.

SL-SWoods: - Changed reducers to throw errors on any unknown action. 
- Tests updates to check for thrown errors.

Issue #1397, opened 2019-11-21
News - UX V2 changes


sarahlallemandsl:
V2 changes for the news window are: 
- New background colours
- Updated text styles

sarahlallemandsl:
**Specs**: [https://github.com/ScottLogic/StockFlux/tree/UX/assets/specs/news/v2](url)
**Invision to reference:** [https://projects.invisionapp.com/d/main?origin=v7#/projects/prototypes/18813715?scrollOffset=1830](url)

sarahlallemandsl:
Based on Friday's meeting I have written some notes with corrections and things to check over.
![News_changes](https://user-images.githubusercontent.com/57100449/69627761-2baf8d80-1043-11ea-81c1-a7312842da9a.png)

Johnston97:
I've addressed these comments!

![image](https://user-images.githubusercontent.com/23056937/69825495-7da11080-1207-11ea-80cd-ec975238a676.png)

Issue #1392, opened 2019-11-19
Update the nav bars - new UX design


sarahlallemandsl:
Slight updates to the font and components on the nav bar.

sarahlallemandsl:
Specs: [https://github.com/ScottLogic/StockFlux/tree/UX/assets/specs/nav%20bar/v2](url)
Open/close icons: [https://github.com/ScottLogic/StockFlux/tree/UX/assets/icons/v2/component](url)

PR #1398, opened 2019-11-21
Minor ux changes

Johnston97:
Address issues:

- https://github.com/ScottLogic/StockFlux/issues/1392

- https://github.com/ScottLogic/StockFlux/issues/1397


packages/stockflux-news/src/components/news-item/NewsItem.css
```diff
@@ -33,7 +36,7 @@
     font-size: var(--font-xlarge);
     text-transform: capitalize;
     font-weight: var(--font-weight-bold);
-    color: var(--color-highlight);
+    color: white;
```
SL-SWoods: should we not use --color-text instead of white

PR #1426, opened 2019-11-28
allow use of extra apps in the launcher

SL-SWoods:
Modify the launcher to allow multiple apps to show on the launcher.
Includes ReadMe file within `developer-proxy` with instructions on how to add new options to the list.

### DIFF
--- a/developer-proxy/readMe.md
+++ b/developer-proxy/readMe.md
@@ -1,0 +1,43 @@
+# Developer Proxy
+
+Use of the developer proxy allows you to develop locally in reguards to each seperate app. This can be ran be using the command `npm start` with `./developer-proxy` directory. 
+
+# App Directory
+To modify the app directory to use the local proxy, modify the file `./src/static/apps.json` new stockflux or external apps can be added in for example this is the addition for a Giant Machines Watchlist
+```json
+{
+    "images": [],
+    "intents": [
+      {
+        "name": "WatchlistAdd"
+      },
+      {
+        "name": "WatchlistView"
+      }
+    ],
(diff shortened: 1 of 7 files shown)


developer-proxy/readMe.md
```diff
@@ -0,0 +1,43 @@
+# Developer Proxy
+
+Use of the developer proxy allows you to develop locally in reguards to each seperate app. This can be ran be using the command `npm start` with `./developer-proxy` directory.
```
Johnston97: The command for this has been changed to npm run proxy

packages/stockflux-components/src/components/buttons/borderless-buttons/app-shortcuts/External.js
```diff
@@ -0,0 +1,26 @@
...
+  if (small) {
+    return (
+      <RoundButton
+        className="shortcut external"
+        onClick={() => Launchers.launchChildWindow(manifest)}
+        small={small}
+      >
+        {manifest.description}
+      </RoundButton>
+    );
```
Johnston97: I don't think we need the small button here. External apps are only going to appear in the launcher.

developer-proxy/readMe.md
Johnston97: The command for this has been updated to npm run proxy

packages/stockflux-components/src/components/buttons/borderless-buttons/app-shortcuts/External.js
Johnston97: I don't think we need the small button here as external apps will only appear in the launcher and not the search results window.

PR #1436, opened 2019-12-03
V2 launcher design

SL-SWoods:
Refactor launcher to new V2 designs. 
image

- Popup selection of docked mode
- Larger search section
- Updated search results
- New icons
- Multiple program icons supported.


SL-SWoods:
Updated to include non-wrapping text in the search results as per the specifications - 
image

